PR #17, opened 2016-04-25
Adding unit test to verify grammar against known-good test data

mijamo911:
This is just a copy-paste of our internal logic.

Antlr insists that the grammar name be the same as the grammar file minus the .g4 extension, so I've renamed the file to remove the .txt extension. The negative side effect is that when you try to view the grammar on our website you'll get a download instead of just viewing it inline since it's no longer considered text. A follow-up change to augment the mime type database used by Github Pages will fix the issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+bin
+target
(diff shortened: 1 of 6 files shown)


tgregg:
Equivalent functionality ended up being merged in #61. Closing.

PR #27, opened 2016-07-06
Adding C cookbook entry for reading values from C strings

mijamo911:


tgregg:
Closing this in favor of #84, which will involve adding this sample and more to the cookbook in its current form.

PR #43, opened 2018-05-16
Adds ion-python 0.3.0 release to news/libs.

almann:
Tested with local Jekyll.


jonwilsdon (approved):
Fancy. Looks good to me. Only a year and five days between Python releases. ;)

PR #64, opened 2019-03-28
Adds in-browser demo of Ion parsing using ion-js

pbcornell:
Note that ion-bundle.js was simply copied from amzn/ion-js for the time being.


tgregg (approved):
Is there a reason why you're not using an ion-js text writer to write the output?

What's the ideal longer-term strategy for consuming ion-bundle.js? Submodule, or something else?

jonwilsdon:
Would you want to use `ion-bundle.min.js` instead of `ion-bundle.js`? It is approximately half the size.

pbcornell:
> Is there a reason why you're not using an ion-js text writer to write the output?

Why yes, several reasons:  ;)
* didn't realize there was one
* the stale ion-bundle.js I was working with didn't have the code
* I can't get clobs to work with it (a relatively minor issue at this time)
* most relevant:  using the pretty text writer limits future use cases for the widget (e.g., correlating text and binary representations via dynamic highlighting)

I propose sticking with the current approach.

> What's the ideal longer-term strategy for consuming ion-bundle.js? Submodule, or something else?

At some point, I'd suggest we update the ion-docs build to pull it from the JS public package manager.

pbcornell:
> Would you want to use `ion-bundle.min.js` instead of `ion-bundle.js`? It is approximately half the size.

Yes, good idea.  Figured out how to build ion-js, and have switched to the latest `ion-bundle.min.js`.

PR #68, opened 2019-04-16
Adds a FAQ entry regarding typed null.

### LINKED ISSUE #66: Add to the FAQ an explanation for why Ion has typed nulls.
The question can be:  "Why does Ion have typed nulls?"

tgregg:
Fixes #66


toddjonker (approved):

dlurton:
Does it make sense to note that typed nulls have been around longer than annotations and that it is also possible to accomplish the same thing today with `int::null` for instance?

tgregg:
@dlurton Is what's already there insufficient?

> As such, applications may choose to emit and accept typed `null` as desired, use annotations instead (e.g. `int::null`), or use untyped `null` for all null values.

zslayton:
While `int::null` is pretty elegant and might be more palatable to some, it does come with some hefty overhead in terms of data size. 

`null.int` => `2f`
`int::null` => `f3 01 9d 0f`

Note that 4 bytes is the smallest possible representation for `int::null` -- it may be larger depending on `int`'s position in the symbol table.

dlurton:
@tgregg Somehow I missed that.  That leaves indicating that typed nulls have been around longer than annotations, and possibly also including @zslayton's comment about typed nulls being smaller than using annotations to accomplish the same thing, but I'll leave that up to you as the main question of the FAQ is adequately answered as it is.

PR #69, opened 2019-04-30
ion-java 1.4.0 release note and groupId change notice

raganhan:

### DIFF
--- a/_posts/2019-04-30-ion-java-1_4_0-released.md
+++ b/_posts/2019-04-30-ion-java-1_4_0-released.md
@@ -1,0 +1,18 @@
+---
+layout: news_item
+title: "Ion Java 1.4.0 Released"
+date: 2019-04-24 14:30:00 -0800
+categories: news
+---
+
+From version 1.4.0 forward ion-java has moved to a new maven group id and java package name: `com.amazon.ion`. This move was necessary to keep the publishing of Amazon libraries in maven consistent.
+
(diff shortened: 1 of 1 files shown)


_posts/2019-04-30-ion-java-1_4_0-released.md
```diff
@@ -0,0 +1,18 @@
+---
+layout: news_item
+title: "Ion Java 1.4.0 Released"
+date: 2019-04-24 14:30:00 -0800
+categories: news
+---
+
+From version 1.4.0 forward ion-java has moved to a new maven group id and java package name: `com.amazon.ion`. This move was necessary to keep the publishing of Amazon libraries in maven consistent.
+
+We'll keep supporting `software.amazon.ion` for the foreseeable future by publishing mirrored releases but users are strongly encouraged to migrate. The migration is trivial as the only differences are the group id and package names.
```
therapon: Is it trivial? I thought we had differently named methods but I might be wrong here. 

Also I would add a date to the deprecation, does not have to be specific data, it could be a year or month year and we can send out a specific date later when we have one.

raganhan: It is trivial, since 1.4.0 the only difference is the java package name and group id. We even generate s.a.i releases from the c.a.i with a script

raganhan: For the deprecation date we talked as team and decided not to put one for now until as we don't have a good  way to track external users. 

We can discuss it again on the next team meeting though

therapon (approved):

PR #70, opened 2019-05-03
Adds a Getting Started section to index.md and reorganizes the page a bit

pbcornell:

### DIFF
--- a/index.md
+++ b/index.md
@@ -3,72 +3,134 @@ title: Amazon Ion
 description: "Amazon Ion is a richly-typed, self-describing, hierarchical data serialization format offering interchangeable binary and text representations. Ion was built to address rapid development, decoupling, and efficiency challenges faced every day while engineering large-scale, service-oriented architectures. Ion has been addressing these challenges within Amazon for nearly a decade, and we believe others will benefit as well."
 ---
 
-Amazon Ion is a [richly-typed][13], [self-describing][15], hierarchical data serialization
+<div style="float:right; width:200px" markdown="block">
+
+| [News][7] |
+|------|{% for post in site.posts limit:5 %}
+|<a href="{{site.baseurl}}{{post.url}}">{{ post.title }}</a>|{% endfor %}
+
+</div>
+
+**<font size="+1">Amazon Ion</font>** is a [richly-typed][13], [self-describing][15], hierarchical data serialization
 format offering [interchangeable binary and text][14] representations. The [text format][10]
 (a superset of [JSON][1]) is easy to read and author, supporting rapid
 prototyping. The [binary representation][11] is [efficient to store, transmit, and
 skip-scan parse][16].  The rich type system provides unambiguous semantics for
-long-term preservation of business data which can survive multiple generations
+long-term preservation of data which can survive multiple generations
 of software evolution.
 
-**Available Libraries:** [Ion Java][3] -- [Ion C][4] -- [Ion Python][5] -- [Ion JavaScript][6]
+Ion was built to address rapid development, decoupling, and efficiency
+challenges faced every day while engineering large-scale, service-oriented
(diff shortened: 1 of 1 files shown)


index.md
```diff
@@ -3,72 +3,134 @@ title: Amazon Ion
...
+-1.2e0
+6.02e23
+-inf
+```
+
+* **symbol:**  provides efficient encoding for frequently occurring strings
+```
+inches
+dollars
+'high-priority'    // symbols with special characters are enclosed in single-quotes
```
therapon: I would add in the comment that the special character here is `-`

index.md
```diff
@@ -3,72 +3,134 @@ title: Amazon Ion
...
+dollars::100.0
+height::inches::72
+lotto_numbers::[7, 9, 19, 40, 42, 44]
+```
+
+The [Specification][10] provides an overview of the full set of Ion types.
+
+### Binary Encoding
+
+Ion provides two encodings for the type system:  human-readable text (as shown above), and a space- and read-efficient binary encoding.  When binary-encoded, every Ion value is prefixed with the value's type and length.  The following illustrates a few of the efficiences provied by Ion's binary encoding:
```
therapon: I would remove the fragment "for the type system" in the first sentence. The encoding is of the values not of the types.

index.md
```diff
@@ -3,72 +3,134 @@ title: Amazon Ion
...
+
+* The following timestamp encoded as a JSON string requires 26 bytes:  "2017-07-26T16:30:04.076Z".  This timestamp requires just 11 bytes when encoded in Ion binary:
+```
+6a 80 0f e1 87 9a 90 9e 84 c3 4c
+```
+That first byte `6a` indicates the value is a timestamp (type 6) represented by the subsequent 10 bytes (that's what the `a` represents).  If this particular timestamp value is not of interest, a reader can simply skip 10 bytes.  While this may seem trivial, the efficiencies can be substantial if some or all of the bytes for large container values can be skipped!
```
therapon: Formatting nit: (type 6) => (type `6`)
"a reader can simply skip 10 bytes" => a reader can jump over this value by skipping 10 bytes ahead.

I would not use the fragment "While this may seem trivial",  perhaps
"The ability to skip an Ion value without the need to read an Ion value in its entirety provides faster navigation of Ion streams and Ion documents."

index.md
```diff
@@ -3,72 +3,134 @@ title: Amazon Ion
...
+
+* Binary encoding of a symbol replaces the text of a symbol with an integer that can be resolved to the original text via a symbol table.  This can result in substantial space savings for symbols that occur frequently!
+
+* While blob data is base-64 encoded in text (which produces 4 bytes for every 3 bytes of the original data), a blob encoded as Ion binary is simply encoded as is&mdash;no base-64 expansion required!
+
+Similar space efficiencies are found in other aspects of Ion's binary encoding, but rather than being exhaustive here, perhaps you'd like to . . .
```
therapon: The sentence here lets me to infer that I can use the JS area below to try more examples and see space savings. I would split this sentence into 2 

1. Similar space efficiencies are found in other aspects of Ion's binary encoding. 
2. Give Ion a try!

Make sentence 1, the last sentence of the preceding paragraph. 

Make sentence 2. a new paragraph (add empty line between the preceding paragraph) immediately followed by the JS area.

index.md
pbcornell: added comment re `-`

index.md
pbcornell: removed

index.md
pbcornell: changed (type 6) to (type `6`)

applied different wording to the last sentence

index.md
pbcornell: left 1. as-is (a separate sentence after the bulleted list); made 2. the header of the subsequent section

therapon (approved):

PR #88, opened 2019-11-19
Fixes timestamp type code to be 6 in table

jonwilsdon:
*Description of changes:*

Whoops! Incorrectly added Timestamp as type 5 instead of 6, fixed!


almann (approved):
:shipit:

PR #90, opened 2019-11-26
Relaxes the requirement that new Ion versions must extend the previous version's system symbol table.

tgregg:
*Description of changes:*

After this change, the spec will allow subsequent Ion versions to define a system symbol table that is not necessarily related to the previous version's system symbol table. It does not prevent such versions from logically extending the previous version's symbol table.

This change also stops referring to the system symbol table by name ("$ion"). This did not seem to add any value, as the spec already disallows explicitly importing "$ion".

### DIFF
--- a/docs/symbols.md
+++ b/docs/symbols.md
@@ -59,9 +59,8 @@ binary data, extracting the text of frequently used symbols (field
 names, enumerations, keywords, *etc.*) out of individual documents and
 into a common data structure.
 
-A *system symbol table* is a shared table with the name `"$ion"`. Every version
-of the Ion notation maps to a specific version of the system symbol
-table.
+A *system symbol table* is an implicit shared symbol table. Every version
+of the Ion notation maps to a specific system symbol table.
 
 A *local symbol table* is for the sole use of a well-defined scope of
 Ion data. Since it does not need to be referenced from other contexts,
@@ -74,9 +73,10 @@ Ion text into binary.
 At any point during processing, there is a *current symbol table*, which
 is either a local symbol table or a system symbol table. At the start of
 input, the current symbol table is initialized to be the system symbol
-table for Ion 1.0. The current symbol table is only changed in two
-circumstances: (1) encountering a system identifier at the top-level, or
-(2) encountering a local symbol table at the top-level.
+table for the Ion version denoted by the Ion version marker. The current
+symbol table is only changed in two circumstances: (1) encountering a
+system identifier at the top-level, or (2) encountering a local symbol table
+at the top-level.
 
 ### The Catalog
 
@@ -403,6 +403,8 @@ as the *local min\_id*.
 with the same name, perhaps even the same version. Such a situation provides
 redundant data and allocates unnecessary symbol IDs but is otherwise harmless.
 
+**Note:** This specification disallows (by ignoring) imports named `"$ion"`.
+
 ### Semantics
 
 When mapping from symbol ID to string, there is no ambiguity. However,
@@ -426,16 +428,9 @@ Put another way, string-to-SID mappings have the following precedence:
 System Symbols
 --------------
 
-The version included in the system identifier is independent of the
-version of the implied system symbol table (named `"$ion"`). Each version of
-the Ion specification defines the corresponding system symbol table version.
-Ion 1.0 uses the `"$ion"` symbol table, version 1, and future versions of Ion
-will use larger versions of the `"$ion"` symbol table. `$ion_1_1` will probably
-use version 2, while `$ion_2_0` might use version 5.
-
-Applications and users should never have to care about these symbol
-table versions, since they are never explicit in user data: this
-specification disallows (by ignoring) imports named `"$ion"`.
+Each version of the Ion specification defines a corresponding system symbol
+table. Applications and users should never have to care about these system
+symbol tables, since they are never explicit in user data.
 
 Here are the system symbols for Ion 1.0.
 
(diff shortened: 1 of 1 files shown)


docs/symbols.md
```diff
@@ -74,9 +73,10 @@ Ion text into binary.
 At any point during processing, there is a *current symbol table*, which
 is either a local symbol table or a system symbol table. At the start of
 input, the current symbol table is initialized to be the system symbol
-table for Ion 1.0. The current symbol table is only changed in two
-circumstances: (1) encountering a system identifier at the top-level, or
-(2) encountering a local symbol table at the top-level.
+table for the Ion version denoted by the Ion version marker. The current
+symbol table is only changed in two circumstances: (1) encountering a
+system identifier at the top-level, or (2) encountering a local symbol table
```
zslayton, line 78: I find the phrase "system identifier" (which was in the original text) to be rather ambiguous. I'd suggest swapping it out for "Ion version marker" if that's what it means.

zslayton (approved):

docs/symbols.md
toddjonker: The term also shows up at L334:

> When immediately following an explicit system ID, a top-level struct whose first annotation is `$ion_symbol_table` is interpreted as a *local symbol table*.

I think that when-clause is incorrect, and that such a top-level struct is always an LST, even if it's not immediately following an IVM.

tgregg: @toddjonker Agreed, that should be fixed.

docs/symbols.md
```diff
@@ -486,17 +481,6 @@ Here are the system symbols for Ion 1.0.
...
-
-    $ion_shared_symbol_table::
-    {
-      name: "$ion", version: 1,
-      symbols:
-      [ "$ion", "$ion_1_0", "$ion_symbol_table", "name", "version",
-        "imports", "symbols", "max_id", "$ion_shared_symbol_table"
-      ]
-    }
-
```
toddjonker, line 499: Shouldn't just remove this, it's accurate for Ion 1.0 and we can't change that.

docs/symbols.md
```diff
@@ -426,16 +428,9 @@ Put another way, string-to-SID mappings have the following precedence:
...
-Ion 1.0 uses the `"$ion"` symbol table, version 1, and future versions of Ion
-will use larger versions of the `"$ion"` symbol table. `$ion_1_1` will probably
-use version 2, while `$ion_2_0` might use version 5.
-
-Applications and users should never have to care about these symbol
-table versions, since they are never explicit in user data: this
-specification disallows (by ignoring) imports named `"$ion"`.
+Each version of the Ion specification defines a corresponding system symbol
+table. Applications and users should never have to care about these system
+symbol tables, since they are never explicit in user data.
```
toddjonker, line 434: This change is significant, and I'm overall not inclined to make it without good reason.  The original text was intended to ensure that system symbols are not removed nor renumbered nor reassigned, because doing so adds unnecessary complexity to the system-level processing.  It puts the system symbol table compatibility mechanisms on the same ground as shared symbol tables, and really just treats the system table as a built-in shared table that's automatically imported.

toddjonker (commented):
This change feels like its lacking rationale. Until there's consensus on a 1.1 or 2.0 format, it seems unnecessary and premature.

docs/symbols.md
tgregg: It was never allowed for a user to import a shared symbol table named "$ion". I'm not sure how removing this breaks anything in Ion 1.0.

docs/symbols.md
tgregg: After this change, system symbol tables can still be treated as built-in shared symbol tables that are automatically imported. The change just clarifies that system symbol tables do not need to relate to each other across versions.

I don't think I agree that this adds significant complexity. Implementations already will need to swap in logic upon encountering an IVM that declares a different Ion version. Part of this logic would be to use the correct system symbol table.

tgregg:
> This change feels like its lacking rationale. Until there's consensus on a 1.1 or 2.0 format, it seems unnecessary and premature.

Here's some rationale:
1. There are symbols in the Ion 1.0 system symbol table that are never useful. Having the flexibility to remove these in Ion 1.1 would free up valuable 1-byte SID space.
2. Declaring that the system symbol table is named "$ion" has no impact on the user, as users are explicitly disallowed from importing tables with this name. Additionally, implementations do not care about the name of the system symbol table, as there is only one system symbol table per version. Discussing the system symbol table's name adds noise to the spec and implies that all versions of Ion must extend the "$ion" table.

docs/symbols.md
almann: I don't think we should change the definition for Ion 1.0, it is what it is and I don't think we should back-track on it.  

@toddjonker - can you help elucidate why Ion 2.0 would be required to have a version bump on the `$ion` symbol table (which is already this weird thing that no one can reference).  I am a little more amenable to Ion 1.x requiring support of the of previous symbols to keep the implementations simpler.

Another way I can think of this proposed change, is that the system symbol table is uniquely identified by the Ion version and has no user defined name (i.e. you cannot import it).  I don't really find this onerous to implement personally, but I grant that it adds some marginal complexity.

docs/symbols.md
almann: @toddjonker - is your point that this is helpful for illustrative purposes, because it is actually not valid for any other actual purpose (since `$ion` is reserved)

toddjonker: @almann yes.  It's the proper definition of the Ion 1.0 system symtab, and an example of how shared symtabs are serialized.

docs/symbols.md
toddjonker: > elucidate why Ion 2.0 would be required to have a version bump on the $ion symbol table

I don't intend to say that (today).  I'm saying that, today, we have no idea what Ion 2.0 needs, and therefore this change is premature.  With 2.0 we could throw all of this out if we have better ideas and can provide a smooth migration path.

I also don't see what benefit we get by retroactively dropping the spec that system symbol tables have a well-defined name.  The intent is that system symbol tables _are_ shared symbol tables, and shared symtabs have names and versions.  This is already encoded in our APIs:

> `SymbolTable.isSystemTable()`
> Determines whether this symbol table is a system symbol table, and therefore shared, named, versioned, and read-only.

What benefit is achieved by allowing implementations to have different names for the same symtab?

PR #95, opened 2020-01-31
Adds C examples to the Cookbook

pbcornell:
Adds example ion-c usage; punting on the shared symbol table examples for now.

### DIFF
--- a/guides/cookbook.md
+++ b/guides/cookbook.md
@@ -85,7 +85,7 @@ implied) implementation.
 <div class="tabpane C" markdown="1">
 The following example shows how text Ion data can be read from a string:
 ```c
-#include <cstdlib>
+#include <stdlib.h>
 #include "ionc/ion.h"
 
 #define ION_OK(x) if (x) { printf("Error: %s\n", ion_error_to_str(x)); return x; }
@@ -122,20 +122,17 @@ int main(int argc, char **argv) {
 
     free(field_name);
     free(value);
+
+    return 0;
 }
 ```
 
 The following example shows how data can be written to a buffer as Ion text:
 ```c
-#include <cstdlib>
+#include <stdlib.h>
 #include "ionc/ion.h"
 
 #define ION_OK(x) if (x) { printf("Error: %s\n", ion_error_to_str(x)); return x; }
-
-void ion_string_from_cstr(const char *cstr, ION_STRING *out) {
-    out->value = (BYTE *)cstr;
-    out->length = (SIZE)strlen(cstr);
-}
 
 int main(int argc, char **argv) {
     const char *ion_text = (char *)malloc(200);
@@ -144,6 +141,8 @@ int main(int argc, char **argv) {
     ION_WRITER_OPTIONS options;
 
     memset(&options, 0, sizeof(ION_WRITER_OPTIONS));
+    options.output_as_binary = false;                                 // text output is the default behavior;
+                                                                      // set to true if binary output is desired
 
     ION_OK(ion_writer_open_buffer(&writer,
                                   (BYTE *)ion_text,
@@ -153,11 +152,11 @@ int main(int argc, char **argv) {
     ION_OK(ion_writer_start_container(writer, tid_STRUCT));           // step into a struct
 
     ION_STRING field_name_string;
-    ion_string_from_cstr("hello", &field_name_string);
+    ion_string_assign_cstr(&field_name_string, "hello", strlen("hello"));
     ION_OK(ion_writer_write_field_name(writer, &field_name_string));  // set the field name for the next value to be written
 
     ION_STRING value_string;
-    ion_string_from_cstr("world", &value_string);
+    ion_string_assign_cstr(&value_string, "world", strlen("world"));
     ION_OK(ion_writer_write_string(writer, &value_string));           // write the next value
 
     ION_OK(ion_writer_finish_container(writer));                      // step out of the struct
@@ -165,6 +164,8 @@ int main(int argc, char **argv) {
     ION_OK(ion_writer_close(writer));                                 // close the writer
 
     printf("%s\n", ion_text);                                         // prints {hello:"world"}
+
+    return 0;
 }
 ```
 </div>
@@ -374,7 +375,7 @@ Pretty-printing results in output similar to the following:
 Ion data can be pretty-printed by setting `ION_WRITER_OPTIONS.pretty_print` to `true` as follows:
 
 ```c
-#include <cstdlib>
+#include <stdlib.h>
 #include "ionc/ion.h"
 
 #define ION_OK(x) if (x) { printf("Error: %s\n", ion_error_to_str(x)); return x; }
@@ -413,6 +414,8 @@ int main(int argc, char **argv) {
     printf("%s\n", pretty_text);
 
     free(pretty_text);
+
+    return 0;
 }
 ```
 </div>
@@ -602,10 +605,12 @@ representations of a numeric Ion value.
 <script>writeTabs()</script>
 <div class="tabpane C" markdown="1">
 ```c
-#include <cstdlib>
+#include <stdlib.h>
 #include "ionc/ion.h"
 
 #define ION_OK(x) if (x) { printf("Error: %s\n", ion_error_to_str(x)); return x; }
+
+#define ION_ASSERT_TYPE(type, x) if ((x) != (type)) { ION_OK(IERR_INVALID_STATE); }
 
 int main(int argc, char **argv) {
     const char* ion_text = "1.23456 1.2345e6 123456 12345678901234567890";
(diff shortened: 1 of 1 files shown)


guides/cookbook.md
```diff
@@ -82,6 +82,94 @@ implied) implementation.
...
+```
+
+The following example shows how data can be written to a buffer as Ion text:
+```c
+#include <cstdlib>
+#include "ionc/ion.h"
+
+#define ION_OK(x) if (x) { printf("Error: %s\n", ion_error_to_str(x)); return x; }
+
+void ion_string_from_cstr(const char *cstr, ION_STRING *out) {
```
tgregg: You can also use the existing `ion_string_assign_cstr` function from `ion_string.h` if you want (here and elsewhere).

guides/cookbook.md
```diff
@@ -82,6 +82,94 @@ implied) implementation.
...
+    ION_STRING value_string;
+    ion_string_from_cstr("world", &value_string);
+    ION_OK(ion_writer_write_string(writer, &value_string));           // write the next value
+
+    ION_OK(ion_writer_finish_container(writer));                      // step out of the struct
+
+    ION_OK(ion_writer_close(writer));                                 // close the writer
+
+    printf("%s\n", ion_text);                                         // prints {hello:"world"}
+}
```
tgregg, line 168: `free(ion_text);`

guides/cookbook.md
```diff
@@ -82,6 +82,94 @@ implied) implementation.
...
+    out->length = (SIZE)strlen(cstr);
+}
+
+int main(int argc, char **argv) {
+    const char *ion_text = (char *)malloc(200);
+
+    hWRITER writer;
+    ION_WRITER_OPTIONS options;
+
+    memset(&options, 0, sizeof(ION_WRITER_OPTIONS));
```
tgregg, line 142: Consider also explicitly setting `options.output_as_binary = false;` to illustrate how the output format is specified.

guides/cookbook.md
```diff
@@ -457,6 +600,64 @@ Because Ion has richly defined numeric types, there are often multiple possible
...
+    memset(&options, 0, sizeof(ION_READER_OPTIONS));
+
+    ION_OK(ion_reader_open_buffer(&reader,
+                                  (BYTE *)ion_text,
+                                  (SIZE)strlen(ion_text),
+                                  &options));
+
+    ION_TYPE ion_type;
+
+    ION_OK(ion_reader_next(reader, &ion_type));
```
tgregg, line 629: Maybe add a macro like `#define ASSERT_TYPE(type, x) if ((x) != (type)) { ION_OK(IERR_INVALID_STATE); }`

Then call, e.g., `ASSERT_TYPE(type, tid_DECIMAL);` after each `ion_reader_next` to illustrate that the numeric values have different expected types.

guides/cookbook.md
```diff
@@ -457,6 +600,64 @@ Because Ion has richly defined numeric types, there are often multiple possible
...
+    ION_DECIMAL value_ion_decimal;
+    ION_OK(ion_reader_read_ion_decimal(reader, &value_ion_decimal));
+    char string[50];
+    ION_OK(ion_decimal_to_string(&value_ion_decimal, string));
+    printf("ion_decimal: %s\n", string);
+
```
tgregg, line 637: `ION_OK(ion_decimal_free(&value_ion_decimal));`

guides/cookbook.md
```diff
@@ -457,6 +600,64 @@ Because Ion has richly defined numeric types, there are often multiple possible
...
+    ION_OK(ion_reader_read_int(reader, &value_int));
+    printf("        int: %d\n", value_int);
+
+    ION_OK(ion_reader_next(reader, &ion_type));
+    ION_INT *value_ion_int;
+    ION_OK(ion_int_alloc(NULL, &value_ion_int));
+    ION_OK(ion_reader_read_ion_int(reader, value_ion_int));
+    ION_STRING istring;
+    ION_OK(ion_int_to_string(value_ion_int, NULL, &istring));
+    printf("    ion_int: %s\n", ion_string_strdup(&istring));
```
tgregg:
```c
ion_int_free(value_ion_int);
```
Also, set the result of `ion_string_strdup` to a variable and `free` it when done.

guides/cookbook.md
```diff
@@ -579,6 +780,78 @@ examining the type annotations of each top-level struct and comparing against
...
+
+#define ION_OK(x) if (x) { printf("Error: %s\n", ion_error_to_str(x)); return x; }
+
+void ion_string_from_cstr(const char *cstr, ION_STRING *out) {
+    out->value = (BYTE *)cstr;
+    out->length = (SIZE)strlen(cstr);
+}
+
+int main(int argc, char **argv) {
+    const char ion_binary[] = { ... };    // bytes representing the Ion binary encoding of the above data
```
tgregg: Make it a `BYTE *` since it doesn't represent characters.

guides/cookbook.md
```diff
@@ -579,6 +780,78 @@ examining the type annotations of each top-level struct and comparing against
...
+    ION_OK(ion_reader_next(reader, &ion_type));
+
+    while (ion_type != tid_EOF) {
+        if (ion_type == tid_STRUCT) {
+            BOOL annotation_found;
+            ION_OK(ion_reader_has_annotation(reader, &foo, &annotation_found));
+            if (annotation_found) {
+                ION_OK(ion_reader_step_in(reader));
+
+                ION_TYPE ion_type2;
```
tgregg: Any reason not to reuse `ion_type`?

guides/cookbook.md
```diff
@@ -579,6 +780,78 @@ examining the type annotations of each top-level struct and comparing against
...
+                ION_OK(ion_reader_next(reader, &ion_type2));
+
+                while (ion_type2 != tid_EOF) {
+                    ION_STRING ion_string;
+                    ION_OK(ion_reader_get_field_name(reader, &ion_string));
+                    char *field_name = ion_string_strdup(&ion_string);
+                    if (strcmp(field_name, "quantity") == 0) {
```
tgregg: I recommend creating an ION_STRING for "quantity" and use `ION_STRING_EQUALS` (from `ion_string.h`) instead of allocating a new `char *` and using `strcmp` for each comparison.

guides/cookbook.md
```diff
@@ -771,6 +1109,10 @@ This shared symbol table can be stored in a file (or in a database) to be
 resurrected into a symbol table at runtime.
 
 <script>writeTabs()</script>
+<div class="tabpane C" markdown="1">
+To be determined.
```
tgregg: Maybe better terminology would be "Example not yet implemented." That way we don't indicate that we don't know how to do this in ion-c; we just haven't included an example of it yet. Same below.

tgregg (commented):
Thanks for putting this together. Comments below.

High-level comment: after the examples that show how to instantiate a writer and reader, we could model subsequent examples as functions that accept an already-instantiated writer or reader as parameters (like we do in the Java examples). That avoids some repetition in each example.

guides/cookbook.md
```diff
@@ -82,6 +82,94 @@ implied) implementation.
 ## Reading and Writing Ion Data
 
 <script>writeTabs()</script>
+<div class="tabpane C" markdown="1">
+The following example shows how text Ion data can be read from a string:
+```c
+#include <cstdlib>
```
therapon: I thought cstdlib was a C++ library and not a C library. The point might be mood here but we might want to be explicit on what compiler and libs we are assuming for the examples.

guides/cookbook.md
```diff
@@ -82,6 +82,94 @@ implied) implementation.
...
+    char *value = ion_string_strdup(&ion_string);
+
+    ION_OK(ion_reader_step_out(reader));            // step out of the struct
+    ION_OK(ion_reader_close(reader));               // close the reader
+
+    printf("%s %s\n", field_name, value);           // prints:  hello world
+
+    free(field_name);
+    free(value);
+}
```
therapon, line 127: nit: add `return 0;` as the last statement to the `int main(..)` procedures

guides/cookbook.md
therapon: Are we ok with examples in the cookbook that cannot be executed with a copy-paste?

guides/cookbook.md
therapon: +1. Also `strcmp` has some issues and we should be using `strncmp` if we really have to 
https://www.geeksforgeeks.org/difference-strncmp-strcmp-c-cpp/

guides/cookbook.md
pbcornell: Good catch; I've updated to `#include <stdlib.h>` instead.  And I honestly can't fathom a more confusing name for a C++ library than "cstdlib"....

guides/cookbook.md
pbcornell: thank you, added

guides/cookbook.md
pbcornell: Yes thanks, updated all occurrences

guides/cookbook.md
pbcornell: added along with comment indicating text is default behavior

guides/cookbook.md
pbcornell: Seems to be needed for `char *` vars, but not applicable for `const char *`.  No change made

guides/cookbook.md
pbcornell: Agreed, this is a nice addition

guides/cookbook.md
pbcornell: added

guides/cookbook.md
pbcornell: thanks, done

guides/cookbook.md
pbcornell: done, thanks

pbcornell: I'd suggest we have a strong bias toward examples that _DO_ execute after copy-paste.  tgregg's `BYTE` comment enables this example to do so without lots of casting, so I've included the actual bytes

guides/cookbook.md
pbcornell: Seemed safer (and thus a better example) to define `ion_type2`, but I think you're right, this is overly cautious and not necessary; removed `ion_type2`

guides/cookbook.md
pbcornell: Updated per @tgregg's recommendation

guides/cookbook.md
pbcornell: Agreed; updated here and below

guides/cookbook.md
tgregg: We should not dynamically allocate the memory if we're not going to free it. Just make it a `char[200]`.

(You can force it to work as-is with, e.g., `free((void *)ion_text);` but that's much uglier than the alternative.)

pbcornell: Updated declaration to `char[200]`

tgregg (approved):

Issue #93, opened 2020-01-27
Clarify spec to allow overpadded zero integers.


tgregg:
Currently, the spec states

> If the value is zero then T must be 2, L is zero, and there are no length or magnitude subfields.

This can be interpreted that `0x20` is the only valid representation of integer `0`. However, overpadded representations (e.g. `0x21 0x00`) should also be allowed. I suggest wording like:

"If the value is zero then T must be 2 and L may be zero with no length or magnitude subfields."

PeytonT:
As the raiser of this specification issue, I can attest to its non-hypothetical nature. I have been working on a binary Ion parser in Rust from the specification and had implemented a [parsing error](https://github.com/PeytonT/serde_ion/commit/b5df8e8678c0ce9215b5ca18619e81e84cc39e7f#diff-9a269209bed069ffb36d5565a91c9840 ) for representations of zero other than `0x20` prior to learning that this was an error in the specification.

> "If the value is zero then T must be 2 and L may be zero with no length or magnitude subfields."

Recognizing that the implementation of the 4 Ion team supported parsers form something of a de facto standard superseding the specification, I think it's worth raising the point that the fix as proposed is technically a breaking change. A parser implemented against the Ion 1.0 spec as written will mis-parse `0x21 0x00` as invalid going forward.

To avoid making a breaking change I suggest instead to make the behavior optional at the level of the parser implementation, with whichever direction as the recommendation.

There are already at least two places in the spec with implementation-determined behavior.
[Optional reordering of fields in structs.](http://amzn.github.io/ion-docs/docs/spec.html#struct)
[Use of symbol ID zero for the name of fields in binary structs with NOP pad values.](http://amzn.github.io/ion-docs/docs/binary.html#13-struct)
There are likely a few more, but these two were easy to find.
There's also the matter of the non-uniform and de facto implementation defined maximum values and maximum sizes discussed in #79.

tgregg:
When we make changes to the spec, we always have to assess the risk. In this case, the proposal is to be more permissive of what binary Ion parsers accept. This is inherently low-risk to the user because all existing data will continue to be parsed successfully. This would only break the user if they were depending on overpadded zero-valued integers being *rejected* by the parser. This is highly unlikely for a few reasons:
1. It is hard to imagine a use case for expecting failure for a specific uncommon encoding of a single value for a single type.
[...]

PR #96, opened 2020-02-02
Allow non-compact binary encodings of int zero

PeytonT:
*Issue #, if available:* #93

*Description of changes:*

Updates the binary encoding of integers with value 0 to allow the current behavior of the core parsers, while permitting parsers that currently comply with the specification as written to maintain current behavior without violating the updated specification.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the Apache License 2.0.

### DIFF
--- a/docs/binary.md
+++ b/docs/binary.md
@@ -291,9 +291,8 @@ Int value |  2 or 3 |    L    |
 
 Zero is always stored as positive; negative zero is illegal.
 
-If the value is zero then _T_ must be 2, _L_ is zero, and there are no length or
-magnitude subfields. As a result, when _T_ is 3, both _L_ and the magnitude
-subfield must be non-zero.
+If _L_ is 0 the value is zero, and there are no length or magnitude subfields.
+As a result, when _T_ is 3, both _L_ and the magnitude subfield must be non-zero.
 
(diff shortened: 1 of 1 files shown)


docs/binary.md
```diff
@@ -291,9 +291,10 @@ Int value |  2 or 3 |    L    |
 
 Zero is always stored as positive; negative zero is illegal.
 
-If the value is zero then _T_ must be 2, _L_ is zero, and there are no length or
-magnitude subfields. As a result, when _T_ is 3, both _L_ and the magnitude
-subfield must be non-zero.
+If the value is zero then _T_ must be 2, _L_ should be zero, and if _L_ is zero 
+there must be no length or magnitude subfields. However, implementations should
```
tgregg: As discussed in #93, allowing for inconsistent handling of this case between implementations is probably not justified. Let's remove the second sentence altogether, resulting in:

> If the value is zero then _T_ must be 2, _L_ should be zero, and if _L_ is zero there must be no length or magnitude subfields. When _T_ is 3, both _L_ and the magnitude subfield must be non-zero.

tgregg (approved):
Thanks!

PR #97, opened 2020-02-06
Adds C# examples to the Cookbook

### LINKED ISSUE #102: Update C# cookbook examples to Dispose() of readers
IIonReader is now IDisposable, so the examples should dispose (probably via `using`) of readers.

pbcornell:
Additionally:
* updates the Java "readNumericTypes" example to align more closely with others
* positions note about local symbol tables below the CSV-to-Ion examples

I propose we merge these changes only after the initial release of ion-dotnet and include a link to its API documentation.

Resolves #102

### DIFF
--- a/guides/cookbook.md
+++ b/guides/cookbook.md
@@ -46,9 +46,10 @@ description: "This cookbook provides code samples for some simple Amazon Ion use
 <!--
 function writeTabs() {
   document.write('<div class="tabs">');
-  ['C', 'Java', 'JavaScript'].forEach(lang =>
-    document.write('<button class="tab ' + lang + '"' + ' onclick="openTab(\'' + lang + '\')">' + lang + '</button>')
-  );
+  ['C', 'C#', 'Java', 'JavaScript'].forEach(lang => {
+    var tabName = lang == 'C#' ? 'C-sharp' : lang;
+    document.write('<button class="tab ' + tabName + '"' + ' onclick="openTab(\'' + tabName + '\')">' + lang + '</button>')
+  });
   document.write('</div>');
 }
 // -->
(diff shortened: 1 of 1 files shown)


guides/cookbook.md
```diff
@@ -169,6 +170,59 @@ int main(int argc, char **argv) {
...
+        writer.SetFieldName("hello");      // set the field name for the next value to be written
+        writer.WriteString("world");       // write the next value
+        writer.StepOut();                  // step out of the struct
+        writer.Finish();
+        Console.WriteLine(tw.ToString());  // prints:  {hello:"world"}
+    }
+}
+```
+
+If the desired output is pretty text, pass an instance of `IonTextOptions` with `PrettyPrint` set to `true` to `IonTextWriterBuilder.Build()`.    If Ion binary encoding is desired, use `IonBinaryWriterBuilder` (instead of `IonTextWriterBuilder`).
```
tgregg: Superfluous spacing between sentences.

tgregg: Is the first sentence necessary, given that we have explanation and an example of pretty printing below?

guides/cookbook.md
```diff
@@ -169,6 +170,59 @@ int main(int argc, char **argv) {
...
+{
+    static void Main(string[] args)
+    {
+        TextWriter tw = new StringWriter();
+        IIonWriter writer = IonTextWriterBuilder.Build(tw);
+        writer.StepIn(IonType.Struct);     // step into a struct
+        writer.SetFieldName("hello");      // set the field name for the next value to be written
+        writer.WriteString("world");       // write the next value
+        writer.StepOut();                  // step out of the struct
+        writer.Finish();
```
tgregg, line 218: This comment is applicable to all examples that use `IIonWriter`. Since `IIonWriter` implements `IDisposable`, its `Dispose()` method should be invoked (either directly or via `try/finally`) before it goes out of scope.

As an aside, should we make `IIonReader` implement `IDisposable`too, in case future implementations would need to use it to free resources?

pbcornell: A [`using` declaration](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#using-declarations) (or statement) is C#'s `try/finally` equivalent.  I've added `using` to all the `TextWriter` and `IIonWriter` declarations in this PR.

And yes, I'd say `IIonReader` should extend `IDisposable` too.  I've created https://github.com/amzn/ion-dotnet/issues/70 to track.

guides/cookbook.md
pbcornell: Superfluous spacing removed.

pbcornell: Agreed, and I like the idea of keeping each example focused on a relatively narrow set of concepts.  I've removed information about pretty-printing from the corresponding JS example as well.

tgregg (approved):

tgregg (approved):

PR #104, opened 2020-03-06
RFC: Ion 1.1

zslayton:
*Description of changes:*

~~This RFC draft proposes a straw man version of a new feature for Ion 1.1: **Templates**. It is intended to act as a jumping-off point for conversations around a formal specification.~~

This PR has evolved into the staging area for an overarching Ion 1.1 proposal, which includes two new features: _inline symbols_ and _templates_.

### --> [Rendered view](https://github.com/amzn/ion-docs/blob/bf33a708d806e46bc24e6bad4a95c12fa359bac8/rfcs/ion_1_1/ion_1_1.md#rfc-ion-11) <-- 

### DIFF
--- a/rfcs/ion_1_1/feature-inline_symbols.md
+++ b/rfcs/ion_1_1/feature-inline_symbols.md
@@ -1,0 +1,755 @@
+# RFC: Inline Symbols
+
+- [RFC: Inline Symbols](#rfc-inline-symbols)
+  - [Summary](#summary)
+  - [Motivation](#motivation)
+    - [Short-lived streams](#short-lived-streams)
+    - [Long-lived streams](#long-lived-streams)
+    - [Serializing maps, dictionaries, and associative arrays](#serializing-maps-dictionaries-and-associative-arrays)
+  - [Inline symbols](#inline-symbols)
+    - [Text encoding](#text-encoding)
+    - [Binary encoding](#binary-encoding)
+      - [`0xF3`: Inline symbol values](#0xf3-inline-symbol-values)
+      - [`0xF4`: Structs with inlineable field names](#0xf4-structs-with-inlineable-field-names)
+      - [Inlineable annotations](#inlineable-annotations)
+        - [`0xE1`: Single inlineable annotation](#0xe1-single-inlineable-annotation)
+        - [`0xE2`: Multiple inlineable annotations](#0xe2-multiple-inlineable-annotations)
+	- [Inline symbols' relationship to Shared Symbol Tables](#inline-symbols-relationship-to-shared-symbol-tables)
+    - [Combining inline symbols with Ion templates](#combining-inline-symbols-with-ion-templates)
+  - [Alternatives considered](#alternatives-considered)
+    - [Add some new encodings, but not others](#add-some-new-encodings-but-not-others)
+    - [Encode maps, dicts, etc as something other than a
+      struct](#encode-maps-dicts-etc-as-something-other-than-a-struct)
+    - [Structs with ordered fields](#structs-with-ordered-fields)
+    - [Disallow inline symbols in symbol tables](#disallow-inline-symbols-in-symbol-tables)
+      - [The symbol table is treated as user-level data instead.](#invalid-symbol-table-encodings-are-treated-as-user-data)
+      - [The symbol table is ignored.](#the-symbol-table-is-ignored)
+      - [An error is raised by the reader.](#an-error-is-raised-by-the-reader)
+
+## Summary
+
+This RFC introduces syntax for defining *inline symbols* in binary Ion.
+
+Inline symbols make it possible to write new struct field names, annotations, and symbols to a
+binary Ion stream without first having to modify the active symbol table. This functionality is
+already supported in Ion text, which has the option of either indexing into the symbol table
+(e.g. `$10`) or defining the symbol inline (e.g. `foo` or `'foo'`).
+
+This capability streamlines the writing process, reduces the memory footprint of both readers and
+writers, and can shrink the overall size of the stream.
+
+The changes described in this document are part of the larger [Ion 1.1 RFC](ion_1_1.md#rfc-ion-11).
+
+-----
+
+## Motivation
+
+All Ion streams have a *symbol table*. A symbol table is a list of known strings which can be
+referenced by their offset into the list--their *symbol ID*--instead of writing out their complete
+text each time.
+
+Ion 1.0's [binary encoding spec](https://amzn.github.io/ion-docs/binary.html) requires that all
+symbol values, struct field names, and annotations be encoded as [symbol
+IDs](http://amzn.github.io/ion-docs/docs/symbols.html). This means that before a given text value
+can be used as a symbol, field name, or annotation, it *must* be added to the symbol table first.
+
+This constraint incurs a number of expenses:
+
+* Writers must examine each top-level value that they intend to emit (however deeply nested) to
+  determine which of its symbols, field names, and annotations are not yet in the symbol table.
+* If any such elements are found, a new Local Symbol Table (LST) or LST-append must be written
+  to the stream establishing a new symbol ID for each of those elements.
+* Writing the top-level value itself must be deferred until the new symbols are added to the table, 
+  necessitating additional buffering of encoded output.
+* Both readers and writers must hold the complete symbol table in memory. The more entries there
+  are in the symbol table, the more memory it will consume.
+* Symbol IDs are encoded as variable-length unsigned integers. The more symbol IDs there are, the
+  more bytes it will take to encode the newest symbol IDs.
+  
+The symbol ID encoding requirement is designed to amortize the cost of writing out the complete text
+of recurring text elements over the lifetime of the Ion stream; by paying an up-front cost to
+detect and define new symbols, all future references to the same text element can be made
+substantially cheaper to write. These overhead costs are worthwhile for symbols that are frequently
+referenced. However, there are a variety of use cases for which the costs are never recouped.
+
+### Short-lived streams
+
+Consider this short-lived Ion text stream representing data collected from a weather station:
+
+```js
+{
+  sensorId: 12345,
+  type: sensorData,
+  reading: {
+    temperature: celsius::12.5,
+    time: 2020-10-22T16:00:00Z
+  }
+}
+```
+
+When the same data is re-encoded in binary, all of the symbols, struct field names, and annotations
+must be added to a local symbol table definition that appears at the beginning of the stream.
+
+```js
+$ion_1_0
+$ion_symbol_table::{ // Add all of our symbols, field names, and annotations to the symbol table
+  symbols: [
+    "sensorId",   // $10
+    "type",       // $11
+    "sensorData", // $12
+    "reading",    // $13
+    "temperature",// $14
+    "celsius",    // $15
+    "time",       // $16
+  ]
+}
+{
+  $10: 12345, // Then encode our symbols, field names, and annotations as symbol IDs
+  $11: $12,
+  $13: {
+    $14: $15::12.5,
+	$16: 2020-10-22T16:00:00Z
+  }
+}
+```
+
+The symbol ID encoding requirement forces us to process our sensor data twice: once to detect any
+symbols, annotations, and field names that needed to be added to the symbol table and once to write
+the sensor data itself. However, because none of the values being added to the symbol table appear
+more than once, we do not benefit from the compactness of the symbol ID encodings. The overhead
+associated with adding each of them to the symbol table slows down the writing process and inflates
+the overall size of the data.
+
+| Format | Size in Bytes |
+|:--|:--:|
+| Text | 94 |
+| Binary | 100 |
+
+In some cases, this can be partially addressed by using a Shared Symbol Table at the cost of
+complexity. See the section [*Inline symbols' relationship to Shared Symbol
+Tables*](#inline-symbols-relationship-to-shared-symbol-tables) for more information.
+
+### Long-lived streams
+
+In long-lived Ion streams, the symbol ID encoding mandate can cause the symbol table to become quite
+large.  This causes two problems:
+
+1. Both Readers and Writers are required to hold the entire symbol table in memory. The larger it
+   becomes, the more memory it consumes. Left unchecked, this can lead to program crashes, making it
+   a potential denial-of-service vector.
+2. Symbol IDs are encoded as variable-length unsigned integers; higher ID numbers require more bytes
+   to encode. Adding infrequently referenced text to the symbol table means that more valuable
+   symbols will be costlier to write out.
+   
+The only way to remove entries from the active symbol table is to reset it or import a different one.
+In either case, we are likely to discard very valuable symbols along with those that are infrequently
+referenced. Re-adding the valuable symbols later wastes processing time and adds to the size of the
+stream. Being more selective about which symbols are added to the symbol table would mitigate these
+issues.
+
+### Serializing maps, dictionaries, and associative arrays
+
+Ion's `struct` data type closely resembles the classes, records, and structs found in a variety
+of programming languages. This makes it a natural choice as a serialized representation for such
+values.
+
+Consider this example of a Java class:
+
+```java
+public class Quux {
+  public String foo;
+  public String bar;
+  public int baz;
+
+  // ... rest of the class
+}
+```
+
+In Ion streams that contain several serialized instances of this class, we are guaranteed to recoup
+the data size cost of adding `foo`, `bar`, and `baz` to the symbol table. The fields in the class
+are a fixed set, so every serialized instance will need to reference each field.
+
+Ion's `struct` data type also acts as a mapping from text keys to values of any data type. This
+closely (though often imperfectly) aligns with common data types available in a variety of
+programming languages, including Java's `Map<String, Object>`, JavaScript's `Object`, and so on.
+
+Consider this example of a Python dictionary:
+
+```python
+logins_by_user_id = {
+  'f354c586' : ["2020-11-05T12:01:17.900Z", "2020-11-06T21:08:16.965Z"],
+  'd3f712ea' : ["2019-07-21T11:24:55.000Z", "2020-05-11T19:44:01.044Z"],
+  '467cba51' : ["2020-10-03T01:51:30.011Z", "2020-10-29T22:40:40.331Z"],
+}
+```
+
+Unlike classes and their analogs, the keys used in a series of maps written to an Ion stream are not
+guaranteed to be the same over time. Maps using high-cardinality values like UUIDs and datetimes as
+keys will have keys that never repeat.
+
+In these cases, creating symbol IDs for these keys pollutes the symbol table; it creates additional
+entries that readers and writers must store in memory but which do not provide any data size
+savings. Once added, the only way to remove these values from the symbol table is to reset it. This
+discards valuable symbols in the process, requiring them to be detected again downstream.
+
+## Inline symbols
+
+Inline symbols are an optional encoding which allow the text of a symbol, annotation, or field name to
+be specified within the value itself rather than as an entry in the active symbol table. Conceptually,
+they are analogous to the way symbols are encoded in text Ion.
+
+Inline symbols give writers the flexibility to decide whether and when to add a given string to the
+symbol table, allowing them to make trade-offs in data size, throughput, and memory consumption.
+
+### Text encoding
+
+*This RFC does not propose any syntax changes for Ion text. This description of Ion text is provided
+for convenient comparison.*
+
+In Ion text, symbols can be written out in full mid-stream without first adding them to the symbol table.
+
+For example, in this Ion struct:
+
+```js
+{
+    foo: [1, 2, 3],
+    bar: true,
+    baz: quux
+}
+```
+
+the symbols `foo`, `bar`, `baz`, and `quux` are all defined at their usage site. These "inline"
+symbols are not in the symbol table and do not have a symbol ID.
+
+[`SymbolToken`](http://amzn.github.io/ion-docs/guides/symbols-guide.html#structures)
+representations of such symbols
+[contain the inline text and an undefined import
+location](http://amzn.github.io/ion-docs/guides/symbols-guide.html#reading-symboltokens).
+
+### Binary encoding
+
+The following sections describe how an inline symbol would be encoded in each of the following use
+cases:
+
+1. Symbol values
+2. Struct field names
+3. Annotations
+   1. Sequences of annotations on a value
+   2. A single annotation on a value
+
+#### `0xF3`: Inline symbol values
+
+This encoding is identical to 
+[the `VarUInt`-length encoding for string values](http://amzn.github.io/ion-docs/docs/binary.html#8-string)
+with the exception of the leading type descriptor byte (`0x8E` vs `0xF3`).
+
+A `Length` field containing the number of bytes in the UTF8 representation must always be provided.
+If the `Length` is zero, the symbol's text is the empty string. It is not possible to encode
+`symbol.null` using type descriptor `0xF3`.
+
+```
+        7       4 3       0
+        +---------+---------+
+        |   15    |    3    |
+        +---------+---------+======+
+        |     Length [VarUInt]     |
+        +--------------------------+
+        |     UTF8 Representation  |
+        +--------------------------+
+```
+
+While many applications simply treat strings and symbols as different encoding options for text,
+others rely on the distinction between the types for their business logic. Providing this encoding
+ensures that a symbol value can be written as cheaply as a string value in any use case.
+
+#### `0xF4`: Structs with inlineable field names
+
+Structs with inlineable field names can encode their field names as either symbol IDs or
+inline text. The encoding need not be homogenous; some fields can use symbol IDs while
+others use inline text.
+
+Inline symbol structs have a type descriptor byte of `0xF4`. A `Length` field containing the number
+of bytes in the struct's representation must always follow the type descriptor byte. Because
+non-empty structs must have at least one field and one value, a `Length` of 1 is illegal. (This
+constrasts with Ion 1.0's `0xC1` struct encoding, which guarantees that fields are ordered by symbol
+ID. For more details, see the section [*Structs with ordered fields*](#structs-with-ordered-fields).
+
+Inlineable field names are encoded as a `VarInt` (not a `VarUInt`). The sign bit is used
+to indicate whether the field name has been encoded as a symbol ID or as inline text.
+
+
+```
+        7       4 3       0
+        +---------+---------+
+        |   15    |    4    |
+        +---------+---------+======+
+        :     length [VarUInt]     :
+        +==========================+=================================================+
+        : field name [VarInt + optional UTF8 representation]  :        value         :
+        +============================================================================+
+                    ⋮                     ⋮
+```
+
+If the `VarInt` is positive, then its magnitude represents a symbol ID. Its text can be found in
+the active symbol table.
+
+If the `VarInt` is negative, then its magnitude represents the length of the field name's UTF8
+representation, which follows immediately after the `Length`.
+
+For example, this struct:
+```js
+{
+    $37: 5,
+    foo: 9
+} 
+```
+could be encoded as:
+```
+                   f  o  o
+F4 89 A5 21 05 C3 66 6f 6f 21 09
+ ^  ^  ^  ^  ^  ^           ^  ^---- Value for "foo": 9
+ |  |  |  |  |  |           +------- 1-byte positive integer
+ |  |  |  |  |  +------------------- 0b1100_0011 / VarInt -3 / a 3-byte inline text field name ("foo")
+ |  |  |  |  +---------------------- Value for $37: 5
+ |  |  |  +------------------------- 1-byte positive integer
+ |  |  +---------------------------- 0b1010_0101 / VarInt 37 / symbol ID field name ($37)
+ |  +------------------------------- VarUInt Length: 9 bytes
+ +---------------------------------- Type descriptor: inline symbol struct
+```
+
+If the sign bit is positive and the magnitude is zero, the field name is symbol ID `0`.
+
+If the sign bit is negative and the magnitude is zero, the field name is the empty string.
+
+The [existing rules for NOP padding in struct
+fields](http://amzn.github.io/ion-docs/docs/binary.html#nop-padding-in-struct-fields) also apply to
+this representation.
+
+#### Inlineable annotations
+
+This RFC adds two new encodings for annotations:
+1. An encoding that (like [Ion 1.0's
+   encoding](http://amzn.github.io/ion-docs/docs/binary.html#annotations)) supports annotation
+   sequences of arbitrary size.
+2. An encoding that is optimized for the most common case, in which there is a single annotation.
+
+Both encodings support inline symbol definitions using the `VarInt` encoding scheme described in
+the section [*Structs with inlineable field names*](#0xf4-structs-with-inlineable-field-names).
+
+Because Ion 1.0's wrapper encoding cannot be shorter than 3 bytes, `0xE1` and `0xE2` were not legal
+type descriptor bytes. We use them in Ion 1.1 to represent our new encodings.
+
+##### `0xE2`: Multiple inlineable annotations
+
+[Ion 1.0's annotations encoding](http://amzn.github.io/ion-docs/docs/binary.html#annotations) uses a
+'wrapper' to associate a set of annotations with a given value. The wrapper contains a sequence of
+annotation symbol IDs followed by the value being annotated. The wrapper's header provides separate
+length fields for both:
+
+1. **The total size of the encoded symbol ID sequence and the encoded value.** This allows the
+   reader to skip over the entire wrapper if needed, moving to the next value in the Ion stream.
+2. **The size of the encoded symbol ID sequence.** This allows the reader to skip over the
+   annotations, moving to the annotated value inside the wrapper.
+
+For example, this value:
+```js
+Author::"Ernest Hemingway"
+```
+would be encoded in binary Ion as:
+```
+                   E  r  n  e  s  t     H  e  m  i  n  g  w  a  y
+EE 94 81 8A 8E 90 45 72 6E 65 73 74 20 48 65 6D 69 6E 67 77 61 79 
+ ^  ^  ^  ^  ^  +---- Length: 16 bytes
+ |  |  |  |  +------- String w/VarUInt Length
+ |  |  |  +---------- Annotation symbol ID $10
+ |  |  +------------- Annotations Length: 1 byte
+ |  +---------------- Annotations+Value Length: 20 bytes 
+ +------------------- Annotation wrapper w/VarUInt Length
+```
+*(The above assumes that 'Author' is already in the active symbol table as `$10`.)*
+
+In practice, few readers leverage the ability to skip the entire wrapper due in part to their need
+to read the wrapped value's type descriptor byte before deciding whether the value can be skipped.
+Additionally, the wrapped values themselves already specify their encoded size. This means that two
+of the four annotation header bytes in the above example are of dubious value.
+
+In contrast, Ion 1.1's `0xE2` encoding includes a single `length` field containing the number of
+bytes used to encode the annotation sequence. This field not include the length of the value itself,
+which provides a length encoding of its own.
+
+```
+            7       4 3       0
+            +---------+---------+
+Value with  |   14    |    2    |
+multiple    +---------+---------+=======================================+
+annotations :     length [VarUInt]                                      :
+            +-----------------------------------------------------------+
+            :     annotation [VarInt + optional UTF8 representation]    : ...
+            +-----------------------------------------------------------+
+            :     value                                                 :
+            +-----------------------------------------------------------+
+```
+
+To skip an annotation sequence, the reader must read the `VarUInt` `length` and skip that number of
+bytes. The reader will then be positioned over the annotated value, which it can read or skip as
+needed.
+
+Annotations are written using the `VarInt`-based encoding described in the section [*Structs with
+inlineable field names*](#0xf4-structs-with-inlineable-field-names). They do not need to be encoded
+homogenously; writers can write some annotations in the sequence as symbol IDs and others as inline
+text.
+
+##### `0xE1`: Single inlineable annotation
+
+An unscientific review of sample Ion 1.0 data found that values that have annotations at all most
+commonly have a single annotation. Our `0xE1` encoding is optimized for this use case by not
+requiring a `Length` field at all.
+
+```
+            7       4 3       0
+            +---------+---------+
+Value with  |   14    |    1    |
+singleton   +---------+---------+=======================================+
+annotation  :     annotation [VarInt + optional UTF8 representation]    :
+            +-----------------------------------------------------------+
+            |                           value                           |
+            +-----------------------------------------------------------+
+```
+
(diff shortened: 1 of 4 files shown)


thokevi:
I have been pondering a similar idea, emanating from a use case to be able to annotate any part of an Ion document with metadata, e.g. audit data describing the actor that created/update the document node, version numbers, etc. In this use case, many nodes in an Ion document could share the same metadata. This seems like a good case for using type annotations (which can apply to any node) and the ability for type annotations to represent arbitrary Ion values, rather than just strings. Simplifying the ask further, this could be done by allowing *symbols* to represent arbitrary Ion values, rather than just strings.

This RFC *almost* enables this use case, via its first stated purpose:
> Allowing any valid Ion value to be added to the table, not just strings.

However, I notice an important restriction:
> Template invocations can only appear in positions where an Ion value could appear. (e.g. You cannot use a template invocation as an annotation or field name)

Could we entertain the idea of relaxing this constraint, particularly for type annotations?

cameronsstone:
I like the idea, but I fear it will lead to very confusing Ion. The [Bigger Org-tree example](https://github.com/amzn/ion-docs/blob/bb4966de309128f5813cfa818f3386abe81b2953/rfcs/ion_templates.md#bigger-org-tree-example) has only 8 templates, but it's still somewhat confusing even with the included comments.

Suggestions:
1. Make the templates be specified as a structure, rather than a list. That is, in the `$ion_template_table` the `templates` field would contain a struct with each template referred to by a symbol. This would force users to think about what that template is for, aiding in maintainability. It would also make the template-using elements clearer, as the template could have a meaningful name (symbol) instead of a number.
1. I'd also like the template-using elements to take named arguments, instead of a list. The argument names could be specified in the template using `@argname` instead of `@_`. Your example with three `@_` is fine, but when there are more it would start to have the same problems as methods with many parameters of the same type. 

These changes will obviously increase the size of the text representation, but the binary representation of the structs would be similar in size as the symbols won't use up much space.

zslayton:
Hi @thokevi!

> I have been pondering a similar idea, emanating from a use case to be able to annotate any part of an Ion document with metadata, e.g. audit data describing the actor that created/update the document node, version numbers, etc. In this use case, many nodes in an Ion document could share the same metadata. This seems like a good case for using type annotations (which can apply to any node) and the ability for type annotations to represent arbitrary Ion values, rather than just strings.

Is there any reason this couldn't be done by wrapping each value in a `struct` with all of the desired metadata? Ion templates should still make this easy and cheap:

```
// Declare a new template that wraps the provided value in a metadata struct
$ion_template_table::{
    templates: [
        { // Template 0
            lastModified: {
                byUser: "thokevi",
                timestamp: 1970-01-01T00:00:00Z
            },
            value: @_
        }
    ]
}

// Invoke the template to wrap values as needed
@0({documentId: 123, content: "foo"})
@0({documentId: 456, content: "bar"})

// The above expands to:
{
    lastModified: {
        byUser: "thokevi",
        timestamp: 1970-01-01T00:00:00Z
    },
    value: {
        documentId: 123, 
        content: "foo"
    }
}
{
    lastModified: {
        byUser: "thokevi",
        timestamp: 1970-01-01T00:00:00Z
    },
    value: {
        documentId: 456, 
        content: "bar"
    }
}
```

I believe the main (only?) drawback to this is that the `Reader` needs to `stepIn()` to the wrapper struct instead of the metadata being truly independent from the value.

> Simplifying the ask further, this could be done by allowing _symbols_ to represent arbitrary Ion values, rather than just strings.

> Could we entertain the idea of relaxing this constraint, particularly for type annotations?

I think there's a strong case for doing this in a major version bump to Ion; I quite like the idea of 'annotations' really being arbitrary metadata attached to a value. However, I think it's too large a change to be in scope for this RFC. As currently described, Ion templates represent an change to the way Ion is encoded. Broadening the definition of an annotation entails a change to the Ion data model itself which has a much broader impact.

Ion templates may be complementary to a broader 'metadata' feature. Metadata values are likely to be sizable repeated composite entities; having templates already in the toolbox might make them an easier sell in a separate RFC.

rfcs/ion_templates.md
```diff
@@ -0,0 +1,472 @@
...
+
+```js
+$ion_template_table::{
+  templates : [
+    "sassafras", // template 0
+    1970-01-01T00:00:00Z, // template 1
+    3.14159265358979323846264338323 // template 2
+  ]
+}
+```
```
almann: Why not add `templates` as an optional field in the symbol table struct?  I could imagine this simplifying the processing logic around symbol tables, meaning we have one kind of entity that has two namespaces (e.g. `symbols` and `templates`).

Minimally, a rationale about why having different tables (and the associated definitions around imports/shared/local/system variants) is preferred over this would be good.  I could see tradeoffs either way, but adding templates into symbol tables on first blush seems like a simpler approach to me (but I might be missing something).

rfcs/ion_templates.md
```diff
@@ -0,0 +1,472 @@
...
+           :     length [VarUInt]     :
+           +--------------------------+
+           |     template ID [VarUInt]|
+           +--------------------------+
+           :           value          :
+           +==========================+
+                         ⋮
+```
+
+* `L` values of `0` and `null` are available to be used as sentinels.
```
almann: What is an `L` value of `null`?  Is that `L = 15` which below this is reserved.

rfcs/ion_templates.md
```diff
@@ -0,0 +1,472 @@
...
+           +--------------------------+
+           :           value          :
+           +==========================+
+                         ⋮
+```
+
+* `L` values of `0` and `null` are available to be used as sentinels. 
+* This document treats `L`=0 as a template blank: `@_`
+* `L`=14 is a variable `length` encoding.
+* `L`=15 is reserved for future specs to accommodate not-yet-designed template invocation methods. (e.g. “splicing”)
```
almann: The problem I have with this definition is that it implies that `L=15` is reserved for just template related things, and if it is, then my largest problem with this proposal is that it "consumes" too much of the encoding space for future expansion.

rfcs/ion_templates.md
```diff
@@ -0,0 +1,472 @@
...
+}  
+```
+
+### Template invocation inside an s-expression
+
+```js
+// Invoke template 0 from within an s-expression.
+($ion_template::@0(123))
+// ((123))
+```
```
almann: I *think* this has some lexical implications that really need to be spelled out.

To wit, `@` is an *operator* token on its own, `_` is an identifier token on its own, `0` is an integer token on its own, so does this mean that any or all of the following are equivalent?

```js
($ion_template::@    _)
($ion_template::'@' /*ignore me!*/ '_')
($ion_template::'@_')
```
and

```js
($ion_template::@  0)
($ion_template::'@'  '0')
($ion_template::'@0')
```

rfcs/ion_templates.md
```diff
@@ -0,0 +1,472 @@
...
+
+we can cut the overhead by 80%.
+
+The same technique could be used for local template table appends.
+
+## Shared Templates
+
+Shared template table imports are handled in the same way as shared symbol table imports.
+
```
almann: I think this is a bit light on details on how template tables exist relative to their symbol table counterparts.  I really think this specification needs to spell this out better.  For example, do template tables live in a distinct namespace from symbol tables (i.e. the template table named `foo` is distinct from and has *no* relation to the symbol table named `foo`)?

This is one of the reasons why I prefer adding templates to symbol tables, it inherits the logic, including resolution under the same specification, which streamlines this specification and possibly makes implementing this in existing implementations easier (no need to create parallel APIs for template tables, for example).

zslayton:
> I like the idea, but I fear it will lead to very confusing Ion. The [Bigger Org-tree example](https://github.com/amzn/ion-docs/blob/bb4966de309128f5813cfa818f3386abe81b2953/rfcs/ion_templates.md#bigger-org-tree-example) has only 8 templates, but it's still somewhat confusing even with the included comments.

Thanks for raising this. I should clarify: it is not expected that anyone will use templates in Ion's text encoding. It will only be used in earnest by binary Ion writers/readers. The RFC describes a text encoding for the sake of human readability. The syntax used (e.g. `@_`, `@0(...)`, etc) can be considered "system-level"--technically valid in an Ion document, but never leveraged by users. For other examples of this, see [the Ion spec's section on `symbol`s](http://amzn.github.io/ion-docs/docs/symbols.html), which uses text local symbol tables (`$ion_symbol_table::{...}`) and text symbol IDs (`$1`, `$2`, `$3`). In practice, text Ion streams rarely include explicit local symbol tables and the symbols themselves are written out in full rather than as ID references. Ion text is meant to be human-friendly and does not prioritize compactness.

> Make the templates be specified as a structure, rather than a list. That is, in the $ion_template_table the templates field would contain a struct with each template referred to by a symbol. This would force users to think about what that template is for, aiding in maintainability. It would also make the template-using elements clearer, as the template could have a meaningful name (symbol) instead of a number.

In the binary use case, requiring templates to have `symbol` names would cause `symbol`s and `template`s to share the same ID space. For example:

```
// Reference to symbol $2 ('$ion_1_0')
0x71 0x02

// Invoke template named $2 ('$ion_1_0') without any parameters
0xF1 0x82
```
(See [System Symbols](http://amzn.github.io/ion-docs/docs/symbols.html#system-symbols) for the built-in symbol list.)

This has a few effects:
* Declaring a new template would require that the `symbol` naming it has already been declared. That means that most `template` table appends will have to follow a corresponding symbol table append (or that `template` appends can implicitly modify the `symbol` table). This would increase the binary size and require more bookkeeping on the part of the `Writer`.
* When reading a `template` invocation, the `Reader` would have to first look up the ID in the `symbol` table (to make that `symbol` ID exists) and then in the `template` table (to make sure there's a template with that name).
* If you declare 127 templates, you'll consume 127 symbol IDs in the process. Any new symbols will take 2 or more bytes to represent when used as struct field names or type annotations. If you keep the concepts separate, more `template` invocations and `symbol`s can be represented in a single byte.
* Writers that automatically emit a template declaration would need to automatically generate a meaningful textual name for it, which may or may not be straightforward.

> I'd also like the template-using elements to take named arguments, instead of a list. The argument names could be specified in the template using @argname instead of @_. Your example with three @_ is fine, but when there are more it would start to have the same problems as methods with many parameters of the same type.

I believe the primary driver for this suggestion is improving text readability.

Thanks for taking the time to write up your suggestions! Let me know if there's anything I can expand on.

almann (changes requested):
I like where this is going, I do think there are some significant details to flesh out below.

rfcs/ion_templates.md
```diff
@@ -0,0 +1,472 @@
+# Ion 1.1 draft RFC
+
+Ion `template`s generalize Ion 1.0’s concept of `symbol`s by:
+
+1. Allowing any valid Ion value to be added to the table, not just `string`s.
```
therapon: nit: I think I know what you mean by "the table" here but for an RFC I would expect to be specific, which table are you talking about? Ion 1.0 has a few (or at least the spec uses a few names for instances of tables) and this RFC is adding a new one.

rfcs/ion_templates.md
```diff
@@ -0,0 +1,472 @@
...
+    year: @_,
+    frame: sedan,
+    numberOfWheels: 4,
+    transmission: automatic,
+    hasAirbags: true
+  }]
+}
+```
+
+Template definitions can refer to other templates, but only those templates that have already been defined (i.e. templates with a lower ID) in order to cheaply avoid cycles.
```
therapon: nit: remove "cheaply".

I think we want a clear statement here of intent. 

We want to disallow cycles. The restriction imposed by the RFC that templates can only refer to other templates with a lower ID is the means to achieve it. 

Justification for the decision should be for disallowing cycles not to the mechanism by which we achieve this restriction.

rfcs/ion_templates.md
```diff
@@ -0,0 +1,472 @@
...
+
+
+## Invoking a template
+
+### Invocation syntax
+
+```js
+// Invoke template 0 with the following parameters
```
therapon: nit: on line 6 the doc uses the word "reference" here the section uses "invocation". Be consistent pick one English word to denote the action and use it through out.

rfcs/ion_templates.md
```diff
@@ -0,0 +1,472 @@
...
+@0("Toyota", "Camry", 2017)
+```
+
+Parameters to a template invocation can themselves be template invocations. There is no restriction on which templates can be invoked as parameters.
```
therapon: I would add an example of a template invocation with a parameter that is another template invocation. 

I would also comment on the following: 

1. What are valid and invalid invocation scenarios, Let  `|@X|` to denote the number of `@_` in the template with ID `X` and let `N` be the number of parameters passes at template invocation then what does the spec say about 
    *  `|@X| = N` -- this is the case you already have and what I think is the expectation 
    *  `|@X| < N` -- invocation provides smaller number of parameters than template's `@_`
    *  `|@X| > N` -- invocation provides larger number of parameters than the template's `@_`      
1. Order of parameter template invocations
    * Given the restriction that template bodies can only call templates with lower IDs **and** that a template invocation clones and populates the template definition, order of evaluation seems to not matter. But I might be missing something.

rfcs/ion_templates.md
therapon: Another idea is to annotate s-expressions to indicate that the s-expression that follows is Ion Template related syntax. Here is a rough idea 

Template invocations 

`$ion_template::(1  a b c)` 

Borrowing from the way function calls are typically encoded in an s-expression, first element is the template ID the remaining elements are the parameters. 

`$ion_template::(1 $ion_template(2 a) c)`  an example with template invocation as a parameter. 

Template definition 
```
$ion_template_table::{
  templates : [{ // template 0
    $ion_template::(_)  // drops the `@` 
  }]
}  
```

Taking the idea a little further, the syntax for templates in Ion text (outside of s-expressions) could also be moved to this style, e.g. 

```
$ion_template_table::{
  templates : $ion_template[{ // template 0
    make: @_,
    model: @_,
    year: @_,
    frame: sedan,
    numberOfWheels: 4,
    transmission: automatic,
    hasAirbags: true
  }]
}
```
Given that the context is now specified through an annotation we get some more freedom on the syntax, for example I would drop the `@` signs in the above example and have `_` only. e.g., 

```
$ion_template_table::{
  templates : $ion_template[{ // template 0
    make: _,
    model: _,
    year: _,
    frame: sedan,
    numberOfWheels: 4,
    transmission: automatic,
    hasAirbags: true
  }]
}
```

Personally I would have one annotation for template definition ( `$ion_template`) and another for template invocation (`$ion_template_call`) but that might be overkill.

rfcs/ion_templates.md
```diff
@@ -0,0 +1,472 @@
...
+($ion_template::@0(123))
+// ((123))
+```
+
+
+
+## Restrictions
+
+* Ion templates perform simple expansions. They involve no branching, looping, or recursion. They are guaranteed to terminate.
+* A template always expands to a single, complete Ion value.
```
therapon: Aha. From this then I assume that when a template is invoked and the number of parameters passed is less than the number of `@_` in the template's definition is thus an error. Is that correct? 

This still does not answer the case of passing more parameters than needed.

rfcs/ion_templates.md
```diff
@@ -0,0 +1,472 @@
...
+  occupation: SDE
+}
+```
+
+This feature, paired with [support for extending containers](#support-for-extending-containers) below, allows users to leverage existing template definitions to encode complex values rather than having to define a collection of very similar templates.
+
+
+## Support for extending containers
+
+If a template is a container type, invocations can provide enough values to fill in the template's blanks and then an extra parameter: a container of the same type as the template container. Values from this extra container will be appended to the end of the expanded container.
```
therapon: Ah even more variation of the case of extra params provided. We should really be explicit about it if we are to rely on these variations and spell out all cases.

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+## Summary
+
+This RFC introduces syntax for defining *inline symbols* in binary Ion.
+
+Inline symbols make it possible to write new struct field names, annotations, and symbols to a
+binary Ion stream without first having to modify the active symbol table. This functionality is
+already supported in Ion text, which has the option of either indexing into the symbol table
+(e.g. `$10`) or defining the symbol inline (e.g. `foo` or `'foo'`).
+
+This capability streamlines the writing process, reduces the memory footprint of both readers and
```
toddjonker, line 38:
```suggestion
This capability streamlines the writing process, enables reducing the memory footprint of both readers and
```
I don't think it inherently reduces memory, but allows documents to be constructed with less memory or read with less memory.  (Though I'm interested to see how the latter is accomplished.)

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+writers, and can shrink the overall size of the stream.
+
+The changes described in this document are part of the larger [Ion 1.1 RFC](ion_1_1.md#rfc-ion-11).
+
+-----
+
+## Motivation
+
+All Ion streams have a *symbol table*. A symbol table is a list of known strings which can be
```
toddjonker, line 47: "have" is imprecise, and "a" suggests singular cardinality.

```suggestion
Each byte of an Ion stream is within the scope of a single active *symbol table*.
A symbol table is a list of known strings which can be
```

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+text each time.
+
+Ion 1.0's [binary encoding spec](https://amzn.github.io/ion-docs/binary.html) requires that all
+symbol values, struct field names, and annotations be encoded as [symbol
+IDs](http://amzn.github.io/ion-docs/docs/symbols.html). This means that before a given text value
+can be used as a symbol, field name, or annotation, it *must* be added to the symbol table first.
+
+This constraint incurs a number of expenses:
+
+* Writers must examine each top-level value that they intend to emit (however deeply nested) to
```
toddjonker, line 58:
```suggestion
* Writers must examine each value that they intend to emit (however deeply nested) to
```

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+  determine which of its symbols, field names, and annotations are not yet in the symbol table.
+* If any such elements are found, a new Local Symbol Table (LST) or LST-append must be written
+  to the stream establishing a new symbol ID for each of those elements.
+* Writing the top-level value itself must be deferred until the new symbols are added to the table, 
+  necessitating additional buffering of encoded output.
+* Both readers and writers must hold the complete symbol table in memory. The more entries there
+  are in the symbol table, the more memory it will consume.
```
toddjonker, line 65: If the content wasn't in the symbol table, it'd be in the values, perhaps many times. I don't feel that this bullet supports its implication that symtabs increase memory consumption.

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+* Symbol IDs are encoded as variable-length unsigned integers. The more symbol IDs there are, the
+  more bytes it will take to encode the newest symbol IDs.
```
toddjonker, line 67: I don't see why this is an "expense", since even the longest encoded symbol ID is highly likely to be shorter than the equivalent content it replaces.

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+
+In long-lived Ion streams, the symbol ID encoding mandate can cause the symbol table to become quite
+large.  This causes two problems:
+
+1. Both Readers and Writers are required to hold the entire symbol table in memory. The larger it
+   becomes, the more memory it consumes. Left unchecked, this can lead to program crashes, making it
+   a potential denial-of-service vector.
+2. Symbol IDs are encoded as variable-length unsigned integers; higher ID numbers require more bytes
+   to encode. Adding infrequently referenced text to the symbol table means that more valuable
+   symbols will be costlier to write out.
```
toddjonker, line 142: If you know in advance which symbols are more-frequently used, it's clear how to promote them to smaller IDs.  If you don't know that data, than I'm not sure how this proposal will be able to do any better.

Further, if you do have frequency data, then I think you're better off using a shared symtab to capture that.  Then a stream-reset is less of an issue.

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+In some cases, this can be partially addressed by using a Shared Symbol Table at the cost of
+complexity. See the section [*Inline symbols' relationship to Shared Symbol
+Tables*](#inline-symbols-relationship-to-shared-symbol-tables) for more information.
+
+### Long-lived streams
+
+In long-lived Ion streams, the symbol ID encoding mandate can cause the symbol table to become quite
+large.  This causes two problems:
+
+1. Both Readers and Writers are required to hold the entire symbol table in memory. The larger it
```
toddjonker, line 137: Long-lived streams and singular up-front symtabs are in contradiction. The (reasonable, IMO) long-stream expectation is that you'd reset/flush the stream regularly, and symtab-append was designed to make that approach more feasible.

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+  '467cba51' : ["2020-10-03T01:51:30.011Z", "2020-10-29T22:40:40.331Z"],
+}
+```
+
+Unlike classes and their analogs, the keys used in a series of maps written to an Ion stream are not
+guaranteed to be the same over time. Maps using high-cardinality values like UUIDs and datetimes as
+keys will have keys that never repeat.
+
+In these cases, creating symbol IDs for these keys pollutes the symbol table; it creates additional
+entries that readers and writers must store in memory but which do not provide any data size
```
toddjonker, line 191: This is the most clear-cut and convincing rationale so far in the document, but it's a _better_ argument for a more-generic map type.  (The name "struct" was carefully chosen with consideration for its connotation of structured data types and their repetitive nature.)

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+
+The Ion 1.0 spec describes system-level values like symbol tables in terms of high-level constructs.
+For example: a local symbol table is a top-level struct whose first annotation is
+`$ion_symbol_table`. If present, the `imports` and `symbols` fields of such structs have special
+meaning.
+
+Because binary Ion 1.0 only had one way to encode these constructs, reader implementations would
+often hardcode low-level, encoding-specific checks for symbol table processing. For example:
+
+* Checking for type code `13` to test for a struct.
```
toddjonker, line 677: I'd be shocked if this has ever been done. Any sane author would write the LST detector by calling something more abstract like `isStruct`.

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+```js
+{
+    foo: [1, 2, 3],
+    bar: true,
+    baz: quux
+}
+```
+
+the symbols `foo`, `bar`, `baz`, and `quux` are all defined at their usage site. These "inline"
+symbols are not in the symbol table and do not have a symbol ID.
```
toddjonker, line 222: This statement assumes that the section above represents the entire document.  I think less-absolute language would be appropriate.

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+        +---------+---------+======+
+        |     Length [VarUInt]     |
+        +--------------------------+
+        |     UTF8 Representation  |
+        +--------------------------+
+```
+
+While many applications simply treat strings and symbols as different encoding options for text,
+others rely on the distinction between the types for their business logic. Providing this encoding
+ensures that a symbol value can be written as cheaply as a string value in any use case.
```
toddjonker, line 263: Not true: there's a one-byte penalty for text under 15 bytes, which covers something like the p50 of strings in practice.   (Such `string` encodings would have no Length subfield at all.)

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+```
+        7       4 3       0
+        +---------+---------+
+        |   15    |    4    |
+        +---------+---------+======+
+        :     length [VarUInt]     :
+        +==========================+=================================================+
+        : field name [VarInt + optional UTF8 representation]  :        value         :
+        +============================================================================+
+                    ⋮                     ⋮
```
toddjonker, line 290: The use of sign bit feels a bit fancy compared to the overall Ion spec, but the encodings here are otherwise sensible and straightforward.

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+ |  |  |  |  +------- String w/VarUInt Length
+ |  |  |  +---------- Annotation symbol ID $10
+ |  |  +------------- Annotations Length: 1 byte
+ |  +---------------- Annotations+Value Length: 20 bytes 
+ +------------------- Annotation wrapper w/VarUInt Length
+```
+*(The above assumes that 'Author' is already in the active symbol table as `$10`.)*
+
+In practice, few readers leverage the ability to skip the entire wrapper due in part to their need
+to read the wrapped value's type descriptor byte before deciding whether the value can be skipped.
```
toddjonker, line 372: Without data, I find this statement hard to believe.

I'm willing to bet that a robust majority of skipping is based on field names, in which case the total-size in the wrapper is valuable.

Beyond that, I'd expect that it's _at least_ as common to skip a value based on its annotation as on its type.

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+Additionally, the wrapped values themselves already specify their encoded size. This means that two
+of the four annotation header bytes in the above example are of dubious value.
```
toddjonker, line 374: "dubious" feels judgemental.  It's a time vs space tradeoff.  The annotation design ensures that all values can be skipped with a single length-decode, at the cost of some redundancy in the encoding.  `0xE2` doubles the cost of skipping annotated values.

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+
+In contrast, Ion 1.1's `0xE2` encoding includes a single `length` field containing the number of
+bytes used to encode the annotation sequence. This field not include the length of the value itself,
+which provides a length encoding of its own.
```
toddjonker, line 378: This should be more explicit in the diagram below.

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+            +-----------------------------------------------------------+
+            :     annotation [VarInt + optional UTF8 representation]    : ...
+            +-----------------------------------------------------------+
+            :     value                                                 :
+            +-----------------------------------------------------------+
+```
+
+To skip an annotation sequence, the reader must read the `VarUInt` `length` and skip that number of
+bytes. The reader will then be positioned over the annotated value, which it can read or skip as
+needed.
```
toddjonker, line 395: This is so irregular WRT to the rest of the encoding, it feel prone to bugs. I expect we'll need to make a number of modifications throughout the spec to accommodate this exception. At this point in my reading of the document, I'm not convinced this is net-positive.

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+This RFC adds two new encodings for annotations:
+1. An encoding that (like [Ion 1.0's
+   encoding](http://amzn.github.io/ion-docs/docs/binary.html#annotations)) supports annotation
+   sequences of arbitrary size.
+2. An encoding that is optimized for the most common case, in which there is a single annotation.
+
+Both encodings support inline symbol definitions using the `VarInt` encoding scheme described in
+the section [*Structs with inlineable field names*](#0xf4-structs-with-inlineable-field-names).
+
+Because Ion 1.0's wrapper encoding cannot be shorter than 3 bytes, `0xE1` and `0xE2` were not legal
```
toddjonker, line 339: Probably worth mentioning the meaning of `0xE0` as well.

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+	$16: 2020-10-22T16:00:00Z
+  }
+}
+```
+
+This eliminates the need to write the UTF8 bytes of each expected symbol's text in the datagram
+itself, shrinking the total size of the data. This benefit comes at the cost of complexity; the
+stream is no longer self-contained. Readers and writers must coordinate to ensure that they both
+have access to the same SSTs. If the set of symbols changes, the writer will need to
+ensure that a new version of the SST is made available.
```
toddjonker, line 529: It's long been observed that the downsides here can be eased in many cases by prepending or inserting the shared symbol table into the stream.  In fact originally the `$ion_shared_symbol_table` annotation, when encountered at top level, _meant_ that the reader was to add that table to its catalog.  We decided to pull that out of the primary spec and to leave such behavior to another protocol level.  Such behavior would be easy to spec out and would address a number of the concerns presented by this RFC.

So the "stream is no longer self-contained" is a red-herring; it could easily be made self-contained if the application wanted to accomplish that. It's simply not _required_.

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+
+In the case of *expected symbols*--symbols which a writer is likely to reference in a given
+stream--SSTs and inline symbols can offer similar benefits. Both allow the writer to avoid writing a
+Local Symbol Table definition at the beginning of an Ion stream. Streams that repeatedly reference
+the same symbols and systems that reuse symbols across many streams will both benefit more from an
+SST than from inline symbols if the complexity cost is acceptable.
```
toddjonker, line 535: The costs aren't just complexity, it's also time/space.

SST costs the complexity of communicating the SST, but pays-back the size of the data, which at scale can be very high payback. The memory-cost to the writer is amortized-to-negligible for most server cases.

Inline symbols cost space and time when a symbol is repeated, pay-back space and time when a symbol is not repeated, and bring complexity to the writing application which has to figure out which symbols to inline without costing too much.

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+can be used as a symbol, field name, or annotation, it *must* be added to the symbol table first.
+
+This constraint incurs a number of expenses:
+
+* Writers must examine each top-level value that they intend to emit (however deeply nested) to
+  determine which of its symbols, field names, and annotations are not yet in the symbol table.
+* If any such elements are found, a new Local Symbol Table (LST) or LST-append must be written
+  to the stream establishing a new symbol ID for each of those elements.
+* Writing the top-level value itself must be deferred until the new symbols are added to the table, 
+  necessitating additional buffering of encoded output.
```
toddjonker, line 63: When a stream has only one TLV, which is the most common case, the entire output must be buffered regardless of symbol encoding, due to length-prefixing of the TLV.

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+```
+
+If we use inline symbol definitions for `foo`, `bar`, `baz`, and `com.example.project.Quux` in our
+template definition, we do not need to add them to the symbol table. By allocating a single template
+ID, we can produce a compact representation of a struct composed of several symbols without growing
+the symbol table.
+
+If the template is frequently used, it can be placed in a Shared Symbol Table.
+
+## Alternatives considered
```
toddjonker, line 609: I think you should consider a very different alternative: using a new IVM to switch the _entire stream_ (up to the next IVM) to inline-only.  This eliminates a lot of either-or complexity, you could probably reuse all the existing type-codes and just make tweaks to the int-encoding of SIDS like you've done here.

That's also the most manageable option from a UX perspective.   I'm really afraid to start thinking about the API complexity that's going to be needed in order to make use of these encoding options.

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+annotations they encounter in the stream.
+
+Writers have no way of knowing in advance whether the active symbol table contains *all* of the
+symbols, field names, and annotations that they will be asked to write. As such, the full symbol
+discovery phase of writing still has to happen for every value, even if no new symbols will ever be
+found. Inline symbols make it possible to eliminate some of this processing overhead. For example, a
+writer leveraging inline symbols may opt to perform lookups to see if a value is already in the
+symbol table, but avoid the work of adding it to the table if it is not. Alternatively, a writer
+could avoid *all* symbol table lookups for portions of a stream by always writing the text of any
+symbols, field names, and annotations inline.
```
toddjonker, line 552: This seems to be the only content addressing how this feature is used, and it feels short-shrift.

By itself, inline symbols don't solve the problems above. That's because making inlining a token-by-token _option_ means that you aren't preventing new SID-encoded symbols from coming down the line, which means you _still_ can't flush the symtab.  You can't flush the symtab until the application tells you it's safe to do so, at which point any undefined symbols must either cause failure or be inlined.

I don't think it's meaningful to say that the writer has a value-local choice in the matter, that applications "may opt to ... avoid the work of adding" an undefined symbol to the LST.  The LST has either been flushed or not, and if so, there's no choice in the matter.

But now there's another problem: when starting a struct (say), I don't know in advance whether there will be any un-interned field name, so I don't know which encoding to use.  So either the writer must always use the new variant, or it still needs to perform the "discovery phase".  In practice the "discovery phase" is just the encoding phase, the output being buffered in order to calculate the length-prefix.  So I'm missing how what meaningful "processing overhead" can be eliminated.

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+}
+com.example.project.Quux::{
+  foo: 7,
+  bar: 8,
+  baz: 9,
+}
+```
+
+If we use inline symbol definitions for `foo`, `bar`, `baz`, and `com.example.project.Quux` in our
+template definition, we do not need to add them to the symbol table. By allocating a single template
```
toddjonker, line 603: You _must_ encode those symbols inline, or you'd need to add a preceding LST to define them.

From this perspective, inline-symbols are a great boon to open content, which is something that should probably get more play in this document.

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+
+For example, add inline symbol structs but not inline symbol values. This would conserve type
+descriptor values and reduce the scope of the changes needed for Ion 1.1. However, it would
+eliminate a key feature for writers.
+
+Inline symbol definitions make it possible for writers to serialize a number of values without
+having to modify the active symbol table. This makes it possible to:
+1. Enforce restrictions on the size of the active symbol table (either by number of entries or data
+size) without having to resort to resetting it.
+2. Batch changes to the symbol table to avoid frequent LST appends.
```
toddjonker, line 621: Is frequent LST append seen as a problem?  They are fairly low-overhead.

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+
+## Alternatives considered
+
+### Add some new encodings, but not others
+
+For example, add inline symbol structs but not inline symbol values. This would conserve type
+descriptor values and reduce the scope of the changes needed for Ion 1.1. However, it would
+eliminate a key feature for writers.
+
+Inline symbol definitions make it possible for writers to serialize a number of values without
```
toddjonker, line 617: These use cases should be way up near the start of the document.  (Really the doc should start with use cases and requirements in any case, to make it clear what specific problems are addressed.)

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+having to modify the active symbol table. This makes it possible to:
+1. Enforce restrictions on the size of the active symbol table (either by number of entries or data
+size) without having to resort to resetting it.
+2. Batch changes to the symbol table to avoid frequent LST appends.
+3. Guarantee that all changes to the symbol table happen at particular offsets. (For example,
+symbol table appends only appear every 8MB in a stream.) This makes it possible to skip-scan
+over huge streams while only checking at each 8MB offset to see if there are symbol table changes
+that need to be processed.
```
toddjonker, line 625: This could use more explanation, I don't understand why one would do this.

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+```js
+[foo, 1, bar, 2, baz, 3]
+```
+
+Such solutions are implemented as a layer on top of the Ion data model itself, which presents
+additional challenges. All applications relying on a custom representation of a map must perform
+their own validation to guarantee that every key has an associated value and that each key is a
+string or symbol value. Tools such as [PartiQL](https://partiql.org/) that are designed to work with
+Ion will not recognize custom representations as maps, meaning that convenient syntax meant to work
+with key/value data will not function as expected.
```
toddjonker, line 647: This is a weak argument, since there are a number of higher-level data types, notably set/bag, that are not well-supported by Ion and that require special handling at higher layers.

But really this section doesn't solve the problems noted, it just moves it around.

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+states:
+
+> Binary-encoded structs support a special case where the fields are known to be sorted such that
+> the field-name integers are increasing. This state exists when L is one.
+
+We cannot trivially offer the same special case for [structs with inlineable field
+names](#0xf4-structs-with-inlineable-field-names). Field names which are defined as inline text will
+not have a symbol ID by which they can be ordered. While other encodings could be substituted (for
+example: "fields appear ordered lexicographically by their text"), the value proposition of such
+alternatives is unknown. As such, Ion 1.1 reserves the `Length=1` encoding for a future version of
```
toddjonker, line 664: WRT 0xF4, right?

rfcs/ion_1_1/feature-inline_symbols.md
```diff
@@ -0,0 +1,755 @@
...
+
+Allowing inlineable encodings to be used in symbol tables requires that readers check for higher-level
+constructs in the stream, which can involve multiple low-level checks:
+
+* Checking for type code `13` (a struct) OR type descriptor byte `0xF4` (struct with inlineable field names).
+* Checking whether the annotations list started with a symbol with text `$ion_symbol_table`,
+  regardless of its encoding.
+* Looking for struct fields with the name `imports` or `symbols` regardless of their encoding.
+
+This adds a modest amount of flexibility at the expense of implementation complexity. It may be
```
toddjonker, line 689: Not really: the implementation will simply not bypass the general purpose abstractions in the name of optimization.

toddjonker (commented):
I'm generally in favor of this. But some of the analysis feels hand-wavy, and there's not enough discussion of how this gets used in practice.  Like ordered-structs, it feels like these features will be hard to retrofit onto APIs that generally work very hard to be output-format-agnostic.

I'm also not sold on the orthogonal change to annotation encoding, as discussed there.

The encoding itself is reasonable, but it really screws up the hard-won symmetry and regularity of the 1.0 encoding.  I kinda want to see what you'd propose if you could change the existing typedescs....

zslayton:
This draft RFC is very out of date. Closing this PR in favor of a new draft that will be published in the coming weeks.

Issue #103, opened 2020-03-04
LOB_START and LOB_END are not properly rendered on the docs site


siler:
See http://amzn.github.io/ion-docs/docs/text.html

I see the following for LOB_START and LOB_END:

```antlr
fragment LOB_START    : ';
fragment LOB_END      : ';
```

It's not clear to me how this file is derived from the grammar branch or I'd make a pull request. If it is a manual transcription I can submit one.

tgregg:
Thanks for bringing this to our attention. We will work to fix the rendering issue. In the meantime, the grammar can be viewed here: http://amzn.github.io/ion-docs/grammar/IonText.g4.txt

jonwilsdon:
Resolved by https://github.com/amzn/ion-docs/pull/110

PR #110, opened 2020-03-14
Removes the verbatim copy of the grammar in text.md.

tgregg:
*Issue #, if available:*
#103 

*Description of changes:*

There are two problems with copying the grammar for display on text.md:
1. When changes are made to the grammar, we have to remember to change text.md.
2. The copy can't actually be made verbatim, because this leads to rendering problems (see #103). So we'd have to remember not only to copy the new grammar to text.md, but also to fix the rendering (assuming it's fixable).

Since the authoritative source of the text grammar is only one click away, I think the best solution is to simply remove the copy. If it were possible to transclude from the grammar page that would be another option, but I haven't found a good way to do this.


jonwilsdon (approved):
It looks like it would be possible to use {% include_relative ... %} (https://jekyllrb.com/docs/includes/) except that we did not put the grammar file in the same directory as this md file. I couldn't figure out a quick way to get it done, approving.

PR #112, opened 2020-04-03
Updates ion-java documentation link

pbcornell:

### DIFF
--- a/libs.md
+++ b/libs.md
@@ -11,16 +11,16 @@ description: "The latest news about Amazon Ion and the Amazon Ion community."
 
 | Name | Latest Version | Repository | Documentation |
 |------|----------------|------|---------------|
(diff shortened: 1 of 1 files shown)


jonwilsdon (approved):
Can you also change the GitHub releases link to point to `https://github.com/amzn/ion-java/releases/latest`? We changed it away from that when we had both c.a.i and s.a.i, now that we only have c.a.i. again, we can keep the release pointing to `latest`.

pbcornell:
> Can you also change the GitHub releases link to point to `https://github.com/amzn/ion-java/releases/latest`? We changed it away from that when we had both c.a.i and s.a.i, now that we only have c.a.i. again, we can keep the release pointing to `latest`.

Done, I've made this change for all the libraries that point to their release.

PR #114, opened 2020-04-07
Updating C# example to use HasAnnotation

### LINKED ISSUE #85: Update cookbook examples to use new annotations methods
Specifically, use them in the Sparse Reads example in the cookbook, replacing the inline implementation that is currently there.

_As noted by @tgregg in https://github.com/amzn/ion-dotnet/issues/68#issuecomment-583176210_

carolinemengbq:
resolves https://github.com/amzn/ion-dotnet/issues/85

Updating the Sparse Reads example in the cookbook for C#. Removing the inline implementation to check if the annotation is present to use the new `bool HasAnnotation(string annotation)` implemented in IIonReader.

### DIFF
--- a/guides/cookbook.md
+++ b/guides/cookbook.md
@@ -1012,7 +1012,7 @@ class SparseReads
         IonType type;
         while ((type = reader.MoveNext()) != IonType.None)
         {
-            if (type == IonType.Struct && HasAnnotation(reader, "foo"))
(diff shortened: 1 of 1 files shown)


pbcornell (approved):

pbcornell:
Merges to the gh-pages branch are automatically published to https://amzn.github.io/ion-docs/, so I'd suggest we merge this _after_ the next release of ion-dotnet.

PR #137, opened 2020-11-10
Added inline symbols RFC

zslayton:
*Description of changes:*

In Ion 1.0's binary format, all symbols, struct field names, and annotations must be encoded as symbol IDs.

This addition to the Ion 1.1 RFC proposes an encoding for symbols, struct field names, and annotations that allows the relevant text to be specified inline for one-off usages. This allows writers to decide which symbols are worth adding to the symbol table according to their use case.

### ==> [Rendered View](https://github.com/amzn/ion-docs/blob/066f58aaae84dacb2c4a6973d9586e7877c65003/rfcs/inline_symbols.md#rfc-inline-symbols) <==

### DIFF
--- a/rfcs/inline_symbols.md
+++ b/rfcs/inline_symbols.md
@@ -1,4 +1,4 @@
- RFC: Inline Symbols
+# RFC: Inline Symbols
 
 * [Summary](#summary)
 * [Motivation](#motivation)
@@ -59,7 +59,7 @@ Consider this short-lived Ion text stream representing data collected from a wea
   type: sensorData,
   reading: {
     temperature: celsius::12.5,
-	time: 2020-10-22T16:00:00Z
+    time: 2020-10-22T16:00:00Z
   }
 }
 ```
@@ -113,7 +113,8 @@ This causes two problems:
 The only way to remove entries from the active symbol table is to reset it or import a different one.
 In either case, we are likely to discard very valuable symbols along with those that are infrequently
 referenced. Re-adding the valuable symbols later wastes processing time and adds to the size of the
-stream.
+stream. Being more selective about which symbols are added to the symbol table would mitigate these
+issues.
 
 ### Serializing maps, dictionaries, and associative arrays
 
@@ -137,9 +138,9 @@ In Ion streams that contain several serialized instances of this class, we are g
 the data size cost of adding `foo`, `bar`, and `baz` to the symbol table. The class's fields are a
 fixed set, so every serialized instance will need to reference each field.
 
-Ion's `struct` data type also act as a mapping from text keys to values of any data type. This
+Ion's `struct` data type also acts as a mapping from text keys to values of any data type. This
 closely (though often imperfectly) aligns with common data types available in a variety of
-programming languages, including Java's `Map<String, Object>`, Javascript's `Object`, and so on.
+programming languages, including Java's `Map<String, Object>`, JavaScript's `Object`, and so on.
 
 Consider this example of a Python dictionary:
 
@@ -157,9 +158,9 @@ keys will have keys that never repeat at all.
 
 In these cases, creating symbol IDs for these keys pollutes the symbol table; it creates additional
 entries that readers and writers must store in memory but which do not provide any data size
-psavings. Once added, the only way to remove these values from the symbol table is to reset it
-altogether. This discards valuable symbols in the process, requiring them to be detected again.
-
+savings. Once added, the only way to remove these values from the symbol table is to reset it
+altogether. This discards valuable symbols in the process, requiring them to be detected again
+downstream.
 
 ## Inline symbols
 
@@ -200,8 +201,8 @@ cases:
 1. Symbol values
 2. Struct field names
 3. Annotations
-  a. Singular annotation on a value
-  b. Collections of annotations on a value
+  1. Singleton annotation on a value
+  2. Collections of annotations on a value
 
 #### `0xF3`: Inline symbol values
 
@@ -280,9 +281,9 @@ F4 89 A5 21 05 C3 66 6f 6f 21 09
  +---------------------------------- Type descriptor: inline symbol struct
 ```
 
-If the sign bit is positive and the magnitude is zero, the field name is the empty string.
+If the sign bit is positive and the magnitude is zero, the field name is symbol ID `0`.
 
-If the sign bit is negative and the magnitude is zero, the field name is symbol ID `0`.
+If the sign bit is negative and the magnitude is zero, the field name is the empty string.
 
 The [existing rules for NOP padding in struct
 fields](http://amzn.github.io/ion-docs/docs/binary.html#nop-padding-in-struct-fields) also apply to
@@ -328,13 +329,13 @@ This RFC adds two new encodings for annotations:
 1. An encoding that is optimized for the most common case, in which there is a single annotation.
 2. An encoding that is optimized for the case in which a value has multiple annotations.
 
-Both encodings support inline symbol definitions using the `VarInt` encoding scheeme described in
+Both encodings support inline symbol definitions using the `VarInt` encoding scheme described in
 [Inline symbol structs](#inline-symbol-structs).
 
 Because Ion 1.0's wrapper encoding cannot be shorter than 3 bytes, type descriptor bytes `0xE1` and
 `0xE2` were not legal type descriptor bytes. We use them in Ion 1.1 to represent our new encodings.
 
-##### `0xE1`: Single inline symbol annotation
+##### `0xE1`: Singleton inline symbol annotation
 
 ```js
             7       4 3       0
@@ -364,8 +365,8 @@ would be encoded either using inline symbol text:
        A  u  t  h  o  r        E  r  n  e  s  t     H  e  m  i  n  g  w  a  y
 E1 C6 41 75 74 68 6f 72 8E 90 45 72 6E 65 73 74 20 48 65 6D 69 6E 67 77 61 79 
  |  |                    |  +--- VarUInt Length: 16 bytes
- |	|					 +------ String w/VarUInt Length
- |	+--------------------------- VarInt -6: a 6-byte inline symbol
+ |  |                    +------ String w/VarUInt Length
+ |  +--------------------------- VarInt -6: a 6-byte inline symbol
  +------------------------------ Singleton annotation	
 ```
 
@@ -414,10 +415,9 @@ needed.
 
 ### Combining inline symbols with Ion templates
 
-When combined with 
-[Ion Templates](https://github.com/amzn/ion-docs/blob/ion-templates-rfc/rfcs/ion_templates.md#rfc-ion-templates),
-	inline symbols can further reduce the number of symbol IDs that must be added to the symbol
-	table.
+When combined with [Ion
+Templates](https://github.com/amzn/ion-docs/blob/ion-templates-rfc/rfcs/ion_templates.md#rfc-ion-templates),
+inline symbols can further reduce the number of symbol IDs that must be added to the symbol table.
 	
 Consider a template representing a class called `com.example.project.Quux`:
 
@@ -459,7 +459,6 @@ com.example.project.Quux::{
   bar: 8,
   baz: 9,
 }
-
 ```
 
 If we use inline symbol definitions for `foo`, `bar`, `baz`, and `com.example.project.Quux` in our
--- a/rfcs/inline_symbols.md
+++ b/rfcs/inline_symbols.md
@@ -172,9 +172,8 @@ keys will have keys that never repeat.
 
 In these cases, creating symbol IDs for these keys pollutes the symbol table; it creates additional
 entries that readers and writers must store in memory but which do not provide any data size
-savings. Once added, the only way to remove these values from the symbol table is to reset it
-altogether. This discards valuable symbols in the process, requiring them to be detected again
-downstream.
+savings. Once added, the only way to remove these values from the symbol table is to reset it. This
+discards valuable symbols in the process, requiring them to be detected again downstream.
 
 ## Inline symbols
 
--- a/rfcs/inline_symbols.md
+++ b/rfcs/inline_symbols.md
@@ -1,9 +1,23 @@
 # RFC: Inline Symbols
 
-* [Summary](#summary)
-* [Motivation](#motivation)
-* [Inline symbols](#inline-symbols)
-* [Usage with templates](#usage-with-templates)
+- [RFC: Inline Symbols](#rfc-inline-symbols)
+  - [Summary](#summary)
+  - [Motivation](#motivation)
+    - [Short-lived streams](#short-lived-streams)
+    - [Long-lived streams](#long-lived-streams)
+    - [Serializing maps, dictionaries, and associative arrays](#serializing-maps-dictionaries-and-associative-arrays)
+  - [Inline symbols](#inline-symbols)
+    - [Text encoding](#text-encoding)
+    - [Binary encoding](#binary-encoding)
+      - [`0xF3`: Inline symbol values](#0xf3-inline-symbol-values)
+      - [`0xF4`: Inline symbol structs](#0xf4-inline-symbol-structs)
+      - [Inline annotations](#inline-annotations)
+        - [`0xE1`: Singleton inline symbol annotation](#0xe1-singleton-inline-symbol-annotation)
+        - [`0xE2`: Multiple inline symbol annotations](#0xe2-multiple-inline-symbol-annotations)
+    - [Combining inline symbols with Ion templates](#combining-inline-symbols-with-ion-templates)
+  - [Alternatives considered](#alternatives-considered)
+    - [Add some new encodings, but not others](#add-some-new-encodings-but-not-others)
+    - [Encode maps, dicts, etc as something other than a struct](#encode-maps-dicts-etc-as-something-other-than-a-struct)
 
 ## Summary
 
@@ -201,8 +215,8 @@ cases:
 1. Symbol values
 2. Struct field names
 3. Annotations
-  1. Singleton annotation on a value
-  2. Collections of annotations on a value
+   1. Singleton annotation on a value
+   2. Collections of annotations on a value
 
 #### `0xF3`: Inline symbol values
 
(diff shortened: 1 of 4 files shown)


rfcs/inline_symbols.md
```diff
@@ -0,0 +1,508 @@
...
+
+Consider this short-lived Ion text stream representing data collected from a weather station:
+
+```js
+{
+  sensorId: 12345,
+  type: sensorData,
+  reading: {
+    temperature: celsius::12.5,
+	time: 2020-10-22T16:00:00Z
```
tgregg: Misaligned (is this a tab?).

rfcs/inline_symbols.md
```diff
@@ -0,0 +1,508 @@
...
+
+  // ... rest of the class
+}
+```
+
+In Ion streams that contain several serialized instances of this class, we are guaranteed to recoup
+the data size cost of adding `foo`, `bar`, and `baz` to the symbol table. The class's fields are a
+fixed set, so every serialized instance will need to reference each field.
+
+Ion's `struct` data type also act as a mapping from text keys to values of any data type. This
```
tgregg: act -> acts

rfcs/inline_symbols.md
```diff
@@ -0,0 +1,508 @@
...
+}
+```
+
+Unlike classes and their analogs, the keys used in a series of maps written to an Ion stream are not
+guaranteed to be the same over time. Maps using high-cardinality values like UUIDs and datetimes as
+keys will have keys that never repeat at all.
+
+In these cases, creating symbol IDs for these keys pollutes the symbol table; it creates additional
+entries that readers and writers must store in memory but which do not provide any data size
+psavings. Once added, the only way to remove these values from the symbol table is to reset it
```
tgregg: psavings -> savings

rfcs/inline_symbols.md
```diff
@@ -0,0 +1,508 @@
...
+ |  |  |  |  +---------------------- Value for $37: 5
+ |  |  |  +------------------------- 1-byte positive integer
+ |  |  +---------------------------- 0b1010_0101 / VarInt 37 / symbol ID field name ($37)
+ |  +------------------------------- VarUInt Length: 9 bytes
+ +---------------------------------- Type descriptor: inline symbol struct
+```
+
+If the sign bit is positive and the magnitude is zero, the field name is the empty string.
+
+If the sign bit is negative and the magnitude is zero, the field name is symbol ID `0`.
```
tgregg: Should these be switched around? Sign bit positive means the magnitude is a symbol ID, so it would make sense for positive + 0 to mean SID 0 and negative + 0 to mean empty string.

rfcs/inline_symbols.md
```diff
@@ -0,0 +1,508 @@
...
+In practice, few readers leverage this capability due in part to their need to read the wrapped
+value's type descriptor byte before deciding whether the value can be skipped. Additionally, the
+wrapped values themselves already specify their encoded size. This means that two of the four
+Annotation header bytes in the above example are of dubious value.
+
+This RFC adds two new encodings for annotations:
+1. An encoding that is optimized for the most common case, in which there is a single annotation.
+2. An encoding that is optimized for the case in which a value has multiple annotations.
+
+Both encodings support inline symbol definitions using the `VarInt` encoding scheeme described in
```
tgregg: scheeme -> scheme

rfcs/inline_symbols.md
```diff
@@ -0,0 +1,508 @@
...
+Author::"Ernest Hemingway"
+```
+
+would be encoded either using inline symbol text:
+
+```js
+       A  u  t  h  o  r        E  r  n  e  s  t     H  e  m  i  n  g  w  a  y
+E1 C6 41 75 74 68 6f 72 8E 90 45 72 6E 65 73 74 20 48 65 6D 69 6E 67 77 61 79 
+ |  |                    |  +--- VarUInt Length: 16 bytes
+ |	|					 +------ String w/VarUInt Length
```
tgregg: The formatting got messed up on this line.

rfcs/inline_symbols.md
```diff
@@ -0,0 +1,508 @@
...
+To skip an annotation sequence, the reader must read the `VarUInt` `length` and skip that number of
+bytes. The reader will then be positioned over the annotated value, which it can read or skip as
+needed.
+
+### Combining inline symbols with Ion templates
+
+When combined with 
+[Ion Templates](https://github.com/amzn/ion-docs/blob/ion-templates-rfc/rfcs/ion_templates.md#rfc-ion-templates),
+	inline symbols can further reduce the number of symbol IDs that must be added to the symbol
+	table.
```
tgregg: Extra indentation.

rfcs/inline_symbols.md
```diff
@@ -0,0 +1,508 @@
...
+
+## Alternatives considered
+
+### Add some new encodings, but not others
+
+For example, add inline symbol structs but not inline symbol values. This would conserve type
+descriptor values and reduce the scope of the changes needed for Ion 1.1. However, it would
+eliminate a key feature for writers.
+
+Inline symbol definitions make it possible for writers to serialize a nunmber of values without
```
tgregg: nunmber -> number

rfcs/inline_symbols.md
```diff
@@ -0,0 +1,508 @@
...
+having to modify the active symbol table. This makes it possible to:
+1. Enforce restrictions on the size of the active symbol table (either by number of entries or data
+size) without having to resort to resetting it.
+2. Batch changes to the symbol table to avoid frequent LST appends.
+3. Guarantee that all changes to the symbol table happen on at particular offsets. (For example,
```
tgregg: on at -> at

tgregg (commented):
Looks good to me.

rfcs/inline_symbols.md
zslayton: Good catch, this is a holdover from an earlier draft that inverted the meaning of the sign bit. Fixed.

rfcs/inline_symbols.md
```diff
@@ -0,0 +1,521 @@
...
+
+Inline symbols make it possible to write new struct field names, annotations, and symbols to a
+binary Ion stream without first having to modify the active symbol table. This functionality is
+already supported in Ion text, which has the option of either indexing into the symbol table
+(e.g. `$10`) or defining the symbol inline (e.g. `foo` or `'foo'`).
+
+This capability streamlines the writing process, reduces the memory footprint of both readers and
+writers, and can shrink the overall size of the stream.
+
+The changes described in this document are part of the larger Ion 1.1 RFC. TODO(add link)
```
jonwilsdon: Does this link exist yet?

rfcs/inline_symbols.md
```diff
@@ -0,0 +1,521 @@
...
+
+## Motivation
+
+All Ion streams have a *symbol table*. A symbol table is a list of known strings which can
```
jonwilsdon: `which can referenced` => `which can be referenced`

rfcs/inline_symbols.md
```diff
@@ -0,0 +1,521 @@
...
+  }
+}
+```
+
+The symbol ID encoding requirement forces us to process our sensor data twice: once to detect any
+symbols, annotations, and field names that needed to be added to the symbol table and once to write
+the sensor data itself. However, because none of the values being added to the symbol table appear
+more than once, we do not benefit from the compactness of the symbol ID encodings. The overhead
+associated with adding each of them to the symbol table slows down the writing process and inflates
+the overall size of the data.
```
jonwilsdon: Is it worth mentioning shared symbol tables here? What about this example makes it so that shared symbol tables is not a better solution?

rfcs/inline_symbols.md
```diff
@@ -0,0 +1,521 @@
...
+
+| Format | Size in Bytes |
+|:--|:--:|
+| Text | 94 |
+| Binary | 100 |
+
+### Long-lived streams
+
+In long Ion streams, the symbol ID encoding mandate can cause the symbol table to become quite large.
```
jonwilsdon: Should this be `long-lived Ion streams`? That would match the short-lived example.

rfcs/inline_symbols.md
```diff
@@ -0,0 +1,521 @@
...
+  public String foo;
+  public String bar;
+  public int baz;
+
+  // ... rest of the class
+}
+```
+
+In Ion streams that contain several serialized instances of this class, we are guaranteed to recoup
+the data size cost of adding `foo`, `bar`, and `baz` to the symbol table. The class's fields are a
```
jonwilsdon: Maybe change `The class's fields` to `The fields in the class` to get rid of any hesitation readers have over possessive forms.

rfcs/inline_symbols.md
```diff
@@ -0,0 +1,521 @@
...
+```
+
+Unlike classes and their analogs, the keys used in a series of maps written to an Ion stream are not
+guaranteed to be the same over time. Maps using high-cardinality values like UUIDs and datetimes as
+keys will have keys that never repeat at all.
+
+In these cases, creating symbol IDs for these keys pollutes the symbol table; it creates additional
+entries that readers and writers must store in memory but which do not provide any data size
+savings. Once added, the only way to remove these values from the symbol table is to reset it
+altogether. This discards valuable symbols in the process, requiring them to be detected again
```
jonwilsdon: Is `altogether` necessary here? Can you leave it at `is to reset it.`

rfcs/inline_symbols.md
```diff
@@ -0,0 +1,521 @@
...
+#### `0xF4`: Inline symbol structs
+
+Inline symbol structs are structs whose field names can be encoded as either symbol IDs or
+inline text. The encoding need not be homogenous; some fields can use symbol IDs while
+others use inline text.
+
+Inline symbol structs have a type descriptor byte of `0xF4`. A `Length` field containing the number
+of bytes in the struct's representation must always follow the type descriptor byte.
+
+Inline symbol structs' field names are encoded as a `VarInt` (not a `VarUInt`). The sign bit is used
```
jonwilsdon: Should this be: `Inline symbol struct field names` instead?

rfcs/inline_symbols.md
```diff
@@ -0,0 +1,521 @@
...
+  foo: 7,
+  bar: 8,
+  baz: 9,
+}
+```
+
+If we use inline symbol definitions for `foo`, `bar`, `baz`, and `com.example.project.Quux` in our
+template definition, we do not need to add them to the symbol table. By allocating a single template
+ID, we can produce a compact representation of a struct composed of several symbols without growing
+the symbol table at all.
```
jonwilsdon: Is `at all` needed here? Can you leave it as `without growing the symbol table.`?

rfcs/inline_symbols.md
```diff
@@ -0,0 +1,521 @@
...
+logins_by_user_id = {
+  'f354c586' : ["2020-11-05T12:01:17.900Z", "2020-11-06T21:08:16.965Z"],
+  'd3f712ea' : ["2019-07-21T11:24:55.000Z", "2020-05-11T19:44:01.044Z"],
+  '467cba51' : ["2020-10-03T01:51:30.011Z", "2020-10-29T22:40:40.331Z"],
+}
+```
+
+Unlike classes and their analogs, the keys used in a series of maps written to an Ion stream are not
+guaranteed to be the same over time. Maps using high-cardinality values like UUIDs and datetimes as
+keys will have keys that never repeat at all.
```
jonwilsdon: Is `at all` needed here? Can this be left as `will have keys that never repeat.`?

jonwilsdon (commented):
Some minor comments.

rfcs/inline_symbols.md
zslayton: It doesn't, no. I plan to merge this PR into the `ion-templates-rfc` branch, then follow up with another commit that adds the Ion 1.1 meta-RFC and updates the inline symbols and templates docs to point back to it.

zslayton: I'll remove the `TODO` though, since I'm not likely to forget that step during the next PR.

rfcs/inline_symbols.md
zslayton: Fixed, thanks.

rfcs/inline_symbols.md
zslayton: > What about this example makes it so that shared symbol tables is not a better solution?

When you say "this example" are you referring to the example below, in _Short-lived streams_?

Shared Symbol Tables (SSTs) address a small portion of the problem. Specifically, they make it possible to omit LST appends for _expected symbols,_ which can save on data size. (All of the symbols used in _Short-lived streams_ are expected symbols, so an SST would work well there.)

However, writers have no way of knowing whether the active symbol table (shared, local, or system) contains _all_ of the symbols, field names, and annotations that they will be asked to write. As such, the full 'discovery' phase of writing still has to happen for every value, even if it happens to come up empty-handed each time. For many Ion streams, it's impossible to define all of the symbols you'll need in advance. Serializing maps is likely the biggest driver of this, as the set of struct fields that might be referenced is unknown until runtime.

Additionally, using SSTs involves a commitment to some amount of additional `Catalog` infrastructure. Ion streams relying on SSTs are no longer self-contained, so sharing Ion datagrams becomes more complex. Readers on other hosts must have a way of locating and accessing the declared SST dependencies to read the data. This extra coordination can be worthwhile, but it creates both friction and pitfalls for simpler use cases.

I'll add a paragraph about this at the end of _Short-lived streams_ illustrating the distinction between expected and unexpected symbols.

rfcs/inline_symbols.md
zslayton: Just bein' melodramatic. 🎭

rfcs/inline_symbols.md
zslayton: I've [added a new section](https://github.com/amzn/ion-docs/pull/137/commits/5230db4896768132f506ba7bbd9ba3d96e51982e#diff-748521366873d3b7ee89e9d8ef0b1dbff910dd869c817aa9296284a826e2ac0b) to attempt to address this.

rfcs/inline_symbols.md
```diff
@@ -0,0 +1,624 @@
...
+ensures that a symbol value can be written as cheaply as a string value in any use case.
+
+#### `0xF4`: Inline symbol structs
+
+Inline symbol structs are structs whose field names can be encoded as either symbol IDs or
+inline text. The encoding need not be homogenous; some fields can use symbol IDs while
+others use inline text.
+
+Inline symbol structs have a type descriptor byte of `0xF4`. A `Length` field containing the number
+of bytes in the struct's representation must always follow the type descriptor byte.
```
almann: Is size 1 invalid (and potentially a reserved value)?

rfcs/inline_symbols.md
```diff
@@ -0,0 +1,624 @@
...
+1. An encoding that is optimized for the most common case, in which there is a single annotation.
+2. An encoding that is optimized for the case in which a value has multiple annotations.
+
+Both encodings support inline symbol definitions using the `VarInt` encoding scheme described in
+the section [*Inline symbol structs*](#0xf4-inline-symbol-structs).
+
+Because Ion 1.0's wrapper encoding cannot be shorter than 3 bytes, type descriptor bytes `0xE1` and
+`0xE2` were not legal type descriptor bytes. We use them in Ion 1.1 to represent our new encodings.
+
+##### `0xE1`: Singleton inline symbol annotation
```
almann: Maybe better phrased as "Single, optionally inline annotation" due to the modality of the encoding.

rfcs/inline_symbols.md
```diff
@@ -0,0 +1,624 @@
...
+        +--------------------------+
+        |     UTF8 Representation  |
+        +--------------------------+
+```
+
+While many applications simply treat strings and symbols as different encoding options for text,
+others rely on the distinction between the types for their business logic. Providing this encoding
+ensures that a symbol value can be written as cheaply as a string value in any use case.
+
+#### `0xF4`: Inline symbol structs
```
almann: Maybe better phrased as "Structs with optionally inlined field names"

rfcs/inline_symbols.md
```diff
@@ -0,0 +1,624 @@
...
+ |  |  +------- String w/VarUInt Length
+ |  +---- VarInt 10: symbol ID $10
+ +------- Singleton annotation
+```
+
+To skip a singleton annotation, the reader must read the `VarInt`. If it is negative, the reader
+must skip that number of bytes to move beyond the UTF8 bytes that follow it. The reader will then
+be positioned over the annotated value, which it can read or skip as needed.
+
+##### `0xE2`: Multiple inline symbol annotations
```
almann: Maybe call this "Optionally inline annotations" or some such to line up with the modality of the encoding.

rfcs/inline_symbols.md
```diff
@@ -0,0 +1,624 @@
...
+```js
+$ion_1_0
+$ion_symbol_table::{ // Add all of our symbols, field names, and annotations to the symbol table
+  sensorId,   // $10
+  type,       // $11
+  sensorData, // $12
+  reading,    // $13
+  temperature,// $14
+  celsius,    // $15
+  time,       // $16
```
almann: Shouldn't this be `symbols: [ ... ]`?

Actually, this makes me wonder if we need or have text to emphasize that inline symbols are not allowed as symbol values in the table.  The text in the [symbols doc][1] says:

>The symbols field should be a list of strings. If the field is missing or has any other type, it is treated as if it were an empty list.

[1]: https://amzn.github.io/ion-docs/docs/symbols.html#local-symbol-tables

rfcs/inline_symbols.md
```diff
@@ -0,0 +1,624 @@
...
+  name: "com.example.weather.symbols",
+  version: 1,
+  symbols: [ // Create a Shared Symbol Table with the necessary symbols
+    sensorId,
+    type,
+    sensorData,
+    reading,
+    temperature,
+    celsius,
+    time,
```
almann: Should be string values, unless we allow inline symbols here...

almann (commented):
Overall I think this is good, I have some minor questions around what is allowed in symbol table `symbols` list.

rfcs/inline_symbols.md
zslayton: > Shouldn't this be symbols: [ ... ]?

Good catch! This was a text editing flub on my part, not a bold new proposal 😛. I think the existing text in the symbols doc is sufficient.

rfcs/inline_symbols.md
zslayton: Fixed, thanks.

rfcs/inline_symbols.md
zslayton: Per offline discussion, I'm using the phrase `inlineable`.

rfcs/inline_symbols.md
```diff
@@ -0,0 +1,754 @@
...
+Mandating a single encoding for system-level constructs requires usages of the disallowed encodings
+lead to one of the following behaviors:
+
+1. [The symbol table is treated as user-level data
+   instead.](#invalid-symbol-table-encodings-are-treated-as-user-data)
+2. [The symbol table is ignored.](#the-symbol-table-is-ignored)
+3. [An error is raised by the reader.](#an-error-is-raised-by-the-reader)
+
+None of these alternatives are particularly appealing. Given the costs involved, the format symmetry
+offered by allowing lineeable encodings everywhere seems like a worthwhile benefit.
```
tgregg: lineeable -> inlineable

PR #150, opened 2021-06-30
Make the L = 14 struct size case explicit

jobarr-amzn:
*Description of changes:* I think this phrasing of struct size semantics is a little more clear, opening this PR to see if anyone else agrees.

### DIFF
--- a/docs/binary.md
+++ b/docs/binary.md
@@ -580,12 +580,12 @@ _L_ is one. Thus:
     nested fields.
   * When _L_ is 1, the struct has at least one symbol/value pair, the _length_
     field exists, and the field name integers are sorted in increasing order.
+  * When _L_ is 14, the _length_ field exists, and no assertion is made about field
+    ordering.
   * When _L_ is 15, the value is `null.struct`, and there's no _length_ or
     nested fields.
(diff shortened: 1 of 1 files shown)


docs/binary.md
```diff
@@ -580,12 +580,12 @@ _L_ is one. Thus:
     nested fields.
   * When _L_ is 1, the struct has at least one symbol/value pair, the _length_
     field exists, and the field name integers are sorted in increasing order.
+  * When _L_ is 14, the _length_ field exists, and no assertion is made about field
+    ordering.
   * When _L_ is 15, the value is `null.struct`, and there's no _length_ or
     nested fields.
-  * When _1 < L < 14_ then there is no _length_ field as _L_ is enough to represent
+  * Otherwise _1 < L < 14_ then there is no _length_ field as _L_ is enough to represent
     the struct size, and no assertion is made about field ordering.
```
zslayton: There's a "then" without a leading "if" here. How's this?
```suggestion
  * Otherwise, _1 < L < 14_  and there is no _length_ field as _L_ is enough to represent
    the struct size. No assertion is made about field ordering.
```

zslayton (approved):
I do think this is clearer. 👍 One minor suggestion below.

PR #153, opened 2021-07-07
Allow non-compact binary encodings of decimal zero

PeytonT:
*Issue #, if available:* https://github.com/amzn/ion-docs/issues/152

*Description of changes:*

Updates the binary decimal encoding specification to clarify that the value 0d0 is not required to be encoded with the single byte 0x50.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the Apache License 2.0.


tgregg (approved):
Looks good to me. I'd like to get one more approval before merging.

zslayton (approved):

PR #157, opened 2021-09-20
Slightly clarify where annotations are valid

jobarr-amzn:
*Description of changes:*
Since "content" is not well-defined in the Ion spec, this clarifies that annotations can be applied to values. 

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -673,9 +673,10 @@ value are separated by double-colons:
 ```
 {% raw %}
 int32::12                                // Suggests 32 bits as end-user type
+degrees::'celsius'::100                  // You can have multiple annotaions on a value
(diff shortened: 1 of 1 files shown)


docs/spec.md
```diff
@@ -667,7 +667,8 @@ When multiple annotations are present, the Ion processor will maintain
 their order. Duplicate annotation symbols are allowed but discouraged.
 
 In the text format, type annotations are denoted by a non-null symbol
-token and double-colons preceding any content:
+token and double-colons preceding any value. Multiple annotations on the same
+value are separated by double-colons:
```
zslayton, line 671: I notice the example list just below this only includes a single example of a value with multiple annotations and it's a struct field (which is already a bit confusing for Ion newcomers).

Could you add a plain value with multiple annotations to the examples? (`degrees::'celsius'::100`?)

zslayton (approved):

PR #158, opened 2021-10-04
A few tweaks to docs

jobarr-amzn:
*Description of changes:*
###Use hexadecimal numbers for type codes
We use hex code to represent binary encoding in our examples, and so do binary viewing and editing tools.  Labeling type codes with the hexadecimal numbers fits better. This commit also removes bit numbering in the binary examples, except for signaling the high and low bits of individual bytes. 

###Adopt reference-style links
This makes doc maintenance and citation easier. 

* Remove an aside about annotation(annotation(value)) restriction being possible in grammar. As the aside says, we didn't do that.

### DIFF
--- a/docs/binary.md
+++ b/docs/binary.md
@@ -4,7 +4,7 @@ title: Ion Binary Encoding
 description: "An explanation of the Amazon Ion binary encoding."
 ---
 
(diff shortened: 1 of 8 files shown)


zslayton (commented):
I like it 👍

PR #161, opened 2021-11-22
Explicit null.decimal binary spec

Pyrolistical:
Updated `null.decimal` to be as explicit as `null.float` just above.
Spec matches the test https://github.com/amzn/ion-tests/blob/master/iontestdata/good/nullDecimal.10n


jobarr-amzn (approved):
Thanks for the contribution :)

PR #165, opened 2022-02-11
Adds GitHub Workflow to automatically create news posts for new releases

popematt:
**Issue #, if available:**

None.

**Description of changes:**

Adds a daily GitHub Actions workflow that checks for new releases and creates a new post for each new release it finds.

Here is an example of what it produces: https://github.com/popematt/github-actions-test/commit/9683d0f2b5b0c230d010851d2e8de14f76d16c43

_By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice._


.github/workflows/release-news.yml
```diff
@@ -0,0 +1,36 @@
+name: Release News Workflow
+
+on:
+  # Allow manually starting the workflow incase we need human intervention.
+  workflow_dispatch:
+
+  # Runs at 08:17 UTC / 00:17 PST / 01:17 PDT
+  # 17 is an arbitrarily chosen to avoid running the workflow during the peak at the beginning of the hour
+  schedule:
+    - cron: '17 8 * * *'
```
jobarr-amzn, line 10: Is arbitrary from your meat brain, dice, or what? :D `echo $(($RANDOM % 60))` if in doubt :D

This immediately inspired me to look up whether it was possible to jitter a GitHub Actions schedule- I didn't find anything.

_scripts/generate_release_news.sh
```diff
@@ -0,0 +1,115 @@
...
+ion-java-benchmark-cli
+ion-java-path-extraction
+ion-js
+ion-kotlin-builder
+ion-python
+ion-rust
+ion-schema
+ion-schema-kotlin
+ion-schema-rust"
+
```
jobarr-amzn:
```suggestion
ion-schema-rust
"

```

_scripts/generate_release_news.sh
```diff
@@ -0,0 +1,115 @@
...
+    git add "$news_item_file_path"
+    echo "Generated '$news_item_file_path'"
+    commit_msg_body=$(printf '%s\n%s' "$commit_msg_body" "* $repo_name $tag")
+  fi
+
+  [[ $GITHUB_ACTIONS ]] && echo "::endgroup::"
+done
+
+readonly NUM_NEW_POSTS=$(git status -s -uno | grep -ce .)
```
jobarr-amzn:
```suggestion
readonly NUM_NEW_POSTS=$(git status -s -uno | wc -l)
```

Looks like you could use `wc -l` to count lines here instead of `grep -ce .`. I might prefer that because it's a little more explicit about what you're doing.

jobarr-amzn (approved):

.github/workflows/release-news.yml
popematt: From my brain.

I tried making a variable schedule, but the `cron` value needs to be a static string or else GitHub says it's an invalid workflow.

_scripts/generate_release_news.sh
popematt: Done

_scripts/generate_release_news.sh
popematt: Done

PR #178, opened 2022-09-12
Clarify Expectations for ANTLR Text Grammar

rmarrowstone:
Adds to description of ANTLR Grammar to make clear that it is intended
to be illustrative, not authoritative.

### DIFF
--- a/grammar/IonText.g4.txt
+++ b/grammar/IonText.g4.txt
@@ -1,15 +1,21 @@
 // Ion Text 1.0 ANTLR v4 Grammar
 //
-// The following grammar does not encode all of the Ion semantics, in particular:
-//
+// The following grammar does not encode the Ion Text semantics completely or
+// unambiguously.
+
+// Known gaps:
 //   * Timestamps are syntactically defined but the rules of ISO 8601 need to be
 //     applied (especially regarding day rules with months and leap years).
 //   * Non $ion_1_0 version markers are not trapped (e.g. $ion_1_1, $ion_2_0)
+//   * The semantics of Symbol Ids, Local Symbol Table defines and imports are not validated
 //   * Edge cases around Unicode semantics:
 //      - ANTLR specifies only four hex digit unicode escapes and on Java operates
 //        on UTF-16 code units (this is a flaw in ANTLR).
 //      - The grammar doesn't validate unpaired surrogate escapes in symbols or strings
 //        (e.g. "\udc00")
+//   * There are some ambiguities involving symbol operators within s-expressions
+//      - Do '-' and '+' (possibly others) right or left associate?
+//      - Do symbol operators terminate special float symbols (nan, -inf, +inf)?
 
 grammar IonText;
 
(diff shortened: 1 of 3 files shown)


zslayton (approved):
Thanks for putting this together. Please consider adding a similar disclaimer to [the top of the ANTLR grammar itself](https://amzn.github.io/ion-docs/grammar/IonText.g4.txt).

popematt (approved):

rmarrowstone:
Yea.. Adding to the comment is a great idea. I'm slightly embarrassed I didn't just do that in the first place!

tgregg:
Do we think it's a non-goal for the grammar to be authoritative? Or do we want it to be authoritative, but we're just not sure it's completely correct yet?

zslayton:
> Do we think it's a non-goal for the grammar to be authoritative? Or do we want it to be authoritative, but we're just not sure it's completely correct yet?

I'd like it to be authoritative but am not sure it's completely correct. By "authoritative," I mean that:
* The grammar's acceptance/rejection of an input should determine whether that input is considered valid Ion.
* Readers must understand the input in the same way that the grammar does. (Two readers may agree that `(a--b)` is valid Ion, but disagree about what values the s-expression contains. The grammar's interpretation would be the definitive answer.)
* The grammar's AST for a given input is _not_ mandatory and is only one (good) example of how a text reader may be implemented. Readers are free to use equivalent grammars or bespoke, hand-written parsers.

tgregg (approved):
I agree with Zack's assessment, and I'm fine with adding this disclaimer until we're sure the grammar is correct.

rmarrowstone:
> Do we think it's a non-goal for the grammar to be authoritative? Or do we want it to be authoritative, but we're just not sure it's completely correct yet?

It _should_ be authoritative, but we know it is not complete and do not know if it is correct for some of the more esoteric bits of the spec. I had been shy to put such wording into the description, but I will give it a shot, along with a statement that the known gaps are described in comments within the grammar itself.

rmarrowstone:
I updated this PR per the conversation. I would like a re-review prior to merging.

grammar/IonText.g4.txt
```diff
@@ -10,6 +12,10 @@
 //        on UTF-16 code units (this is a flaw in ANTLR).
 //      - The grammar doesn't validate unpaired surrogate escapes in symbols or strings
 //        (e.g. "\udc00")
+//   * Symbol Tables nor valid Symbol Number ranges are defined.
```
tgregg: I'm not really sure what this one means. Can you give some examples?

tgregg: Discussed offline; an example is a symbol identifier that is out of range of the current symbol table.

The following is invalid:

```$ion_symbol_table::{symbols:["foo"]} $99```

While the following is valid:

```$ion_symbol_table::{symbols:["foo"]} $10```

rmarrowstone: Yes. I tried the most subtle rewrite of the comment to better clarify. I think it helps.

tgregg (approved):

PR #182, opened 2022-10-12
Amend Supported Unicode Text Encodings

rmarrowstone:
This change removes the stated support for UTF-16 and UTF-32,
reflecting a similar change in the JSON specification over time.

RFC 7159 Sec. 8.1 specified that JSON text SHALL be encoded in UTF-8,
UTF-16, or UTF-32. In 2017 RFC 8259 revised Sec. 8.1 to reflect that
JSON text MUST be encoded as UTF-8 when "exchanged between systems
that are not part of a closed ecosystem"

No Ion implementations support non-UTF-8 encodings.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docs/text.md
+++ b/docs/text.md
@@ -7,8 +7,7 @@ description: "An explanation of the Amazon Ion text encoding."
 # [Docs][docs]/ {{ page.title }}
 
 A [value stream][glossary-value-stream] in the text encoding must be a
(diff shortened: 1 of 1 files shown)


popematt (approved):

jobarr-amzn (approved):
Thank you :D

PR #187, opened 2023-01-05
Correct Explanation of Problem with Storing Decimal as Float

rmarrowstone:
The previous explanation used the terms "rational" and "irrational"
and was not correct. A rational number is rational in all bases.

This is my attempt at a more correct *brief* description.

### Description of changes:

----

### DIFF
--- a/docs/decimal.md
+++ b/docs/decimal.md
@@ -7,9 +7,9 @@ description: "Amazon Ion supports a decimal numeric type to allow accurate repre
 # [Docs][docs]/ {{ page.title }}
 
 Amazon Ion supports a decimal numeric type to allow accurate representation
-of base-10 floating point values such as currency amounts. This
-representation preserves significant trailing zeros when converting
-between text and binary forms.
+of base-10 floating point values such as currency amounts. An Ion Decimal
+has arbitrary precision and scale. This representation preserves significant
+trailing zeros when converting between text and binary forms.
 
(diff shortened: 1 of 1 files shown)


popematt (approved):
A couple of things that could _possibly_ be very minor improvements:
* Could be useful to include a link to https://en.wikipedia.org/wiki/Repeating_decimal for further information.
* We could also call out the fact that Floats have fixed number of bits available, whereas Decimals support arbitrary precision. It serves to further distinguish the two data types, but it is bringing a new topic into the discussion here, which may be distracting from the main point.

However, I think this is great, and I have absolutely no qualms signing off on it.

rmarrowstone:
> We could also call out the fact that Floats have fixed number of bits available, whereas Decimals support arbitrary precision. It serves to further distinguish the two data types, but it is bringing a new topic into the discussion here, which may be distracting from the main point.

It does seem somehow lacking that arbitrary precision is not explicitly mentioned in the overview! Will update.

> Could be useful to include a link to https://en.wikipedia.org/wiki/Repeating_decimal for further information.

I spent some time researching this to ensure I was clear and I had a hope to capture one really good link, but couldn't find a satisfying one. I agree that understanding the above is useful to understand that rational numbers are not necessarily terminating, but for someone who has never thought about how floats are encoded they need to start at the top. The best (shortest and most illuminating) was actually the python docs: https://docs.python.org/3/tutorial/floatingpoint.html

PR #188, opened 2023-01-13
Remove mention of comma from sexp description.

toddjonker:
Commas are not valid operator characters.

This documentation defect dates back to 2007-04-24 when we removed comma from the set of operator characters. We felt it risked confusion and ambiguity for `[1, 2]` to have two elements but `(1, 2)` to have three.  (We simultaneously removed colon, for similar (but arguably less convincing) reasons.)

----


tgregg (approved):
Wow! Hard to believe that went unnoticed for so long.

PR #190, opened 2023-01-19
Adds two small improvements to GH Pages site

popematt:
### Issue #, if available:

None

### Description of changes:

* Adds Ion Rust link to the web page footer
* Replaces the hard coded list of repos in favor of using the GitHub CLI to list all public repos in the `amazon-ion` org.

----

### DIFF
--- a/_scripts/generate_release_news.sh
+++ b/_scripts/generate_release_news.sh
@@ -16,39 +16,7 @@
 # This script does not generate news items for releases marked as a pre-release
 # version.
 
-# TODO: See if github actions can use the github cli to automatically list all Ion repositories.
-# readonly REPO_NAMES="$(gh api teams/2323876/repos --jq '.[] | select(.visibility == "public") | .name')"
-
-# This should be kept up-to-date with all PUBLIC Ion repositories.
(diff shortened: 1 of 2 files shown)


_scripts/generate_release_news.sh
```diff
@@ -16,39 +16,7 @@
...
-ion-java-path-extraction
-ion-js
-ion-kotlin-builder
-ion-python
-ion-rust
-ion-schema
-ion-schema-kotlin
-ion-schema-rust
-"
+readonly REPO_NAMES="$(gh repo list amazon-ion --json name --visibility=Public --limit 100 -q '.[].name')"
```
popematt, line 19: Default value for `--limit` is 30, but we have 34 public repos, so we need to use a higher `--limit`. I figure 100 is safe for now.

jobarr-amzn: Would be worth making the above a comment in the source I think, but on the other hand I think we have a real problem if we have 100 repos in the org. 🤷🏻‍♂️

jobarr-amzn (approved):

PR #195, opened 2023-04-10
Add Macros by Example chapter.

toddjonker:
### Description of changes:

Added a narrative chapter explaining macro features incrementally.  Linked to it from `main.adoc`.

----
_**By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.**_

PR #198, opened 2023-04-13
Replace `for-each` special form with more expressive `for`.

toddjonker:
### Description of changes:
Feedback on #195 rightfully observed that `for_each` was limiting and confusing in its handling of the local name.  Here I replace it with a more expressive form called `for`.

----

### DIFF
--- a/src/macros-by-example.adoc
+++ b/src/macros-by-example.adoc
@@ -249,6 +249,9 @@ named `literal`, not to this special form.  As an aside, there is no need for su
 E-expressions, because in that context symbols and S-expressions are not “evaluated”, and
 everything is literal except for E-expressions (which are not data, but encoding artifacts).
 
+NOTE: Ion 1.1 defines a number of built-in macros and special forms. While this document covers
+the highlights, it is not a complete reference to all features.
+
 
 === Parameter Types
 
@@ -462,25 +465,42 @@ types are verified:
 TODO: demonstrate splicing in TDL macro invocations
 
 
-=== Mapping Templates Over Streams
+=== Mapping Templates Over Streams: `*for*`
 
-Another way to produce a stream is via a mapping form. The `*for_each*` special form evaluates a
-template once for each value provided by a macro parameter.  Each time, the parameter name is
-shadowed and rebound to the next value on the stream.
+Another way to produce a stream is via a mapping form.  The `*for*` special form evaluates a
+template once for each value provided by a stream or streams.  Each time, a local variable is
+created and bound to the next value on the stream.
 
 [{mrk}]
 ----
 (*public* prices
-  [(*symbol* currency), (*number\...* amount)]
-  (*for_each* amount
-    (price amount currency)))
+  [(*symbol* currency), (*number\...* amounts)]
+  (*for* [(amt amounts)]
+    (price amt currency)))
 ----
+
+The list following `*for*` contains pairs of variable names and template expressions, so here
+each value from `amounts` is given the name `amt` before the `price` is expanded.
+
 ----
 (:prices GBP 10 9.99 12.)
   ⇒ {amount:10, currency:GBP} {amount:9.99, currency:GBP} {amount:12., currency:GBP}
 ----
 
-TODO: Consider the more-general binding form
+More than one stream can be iterated in parallel, and iteration terminates when any stream
+becomes empty.
+
+[{mrk}]
+----
+(*public* zip [(**any* **front), (**any* **back)]
+  (*for* [(f front),
+        (b back)]
+    [f, b]))
+----
+----
+(:zip [1, 2, 3] [a, b])
+  ⇒ [1, a] [2, b]
+----
 
 
 === Empty Streams: `*void*`
--- a/src/macros-by-example.adoc
+++ b/src/macros-by-example.adoc
@@ -479,8 +479,9 @@ created and bound to the next value on the stream.
     (price amt currency)))
 ----
 
-The list following `*for*` contains pairs of variable names and template expressions, so here
-each value from `amounts` is given the name `amt` before the `price` is expanded.
+The list immediately following `*for*` contains S-expressions pairing variable names with
+template expressions.  Here, each value from the template `amounts` is given the name `amt`
+before the `price` invocation is expanded.
 
 ----
 (:prices GBP 10 9.99 12.)


src/macros-by-example.adoc
```diff
@@ -462,25 +465,42 @@ types are verified:
...
+created and bound to the next value on the stream.
 
 [{mrk}]
 ----
 (*public* prices
-  [(*symbol* currency), (*number\...* amount)]
-  (*for_each* amount
-    (price amount currency)))
+  [(*symbol* currency), (*number\...* amounts)]
+  (*for* [(amt amounts)]
```
popematt, line 478: Would you consider `[amt::amounts]`, `[(amt in amounts)]`, or some other syntax so that it doesn't look quite so much like `[(symbol currency)]`?

I don't think there's a lot of value in being terse in the template expressions for Ion text encoding. Adding a few extra words is a small price to pay for readability (and by extension, ease of learning/use).

src/macros-by-example.adoc
```diff
@@ -462,25 +465,42 @@ types are verified:
...
+    (price amt currency)))
 ----
+
+The list following `*for*` contains pairs of variable names and template expressions, so here
```
popematt: > The list following `*for*` contains pairs of variable names and template expressions

This wording is ambiguous to me. Do you mean to say that I can use template expressions in this list?
```
(*public* prices
  [(*symbol* currency), (*number\...* amounts)]
  (*for* 
    [(amt amounts)]   // <--------------------- here?
    (price amt currency)))
```
Or are you using "list" in a more general sense, and referring to the `price` invocation on the next line?

src/macros-by-example.adoc
```diff
@@ -462,25 +465,42 @@ types are verified:
...
+each value from `amounts` is given the name `amt` before the `price` is expanded.
+
 ----
 (:prices GBP 10 9.99 12.)
   ⇒ {amount:10, currency:GBP} {amount:9.99, currency:GBP} {amount:12., currency:GBP}
 ----
 
-TODO: Consider the more-general binding form
+More than one stream can be iterated in parallel, and iteration terminates when any stream
+becomes empty.
```
popematt, line 492: Very nice!

Is there any way to do something with any remaining items in the non-empty stream(s)?

src/macros-by-example.adoc
toddjonker: When I use the term "list" I always mean "Ion list".  So "the list following *for*" references the list `[(amt amounts)]` and "pairs of variable names and template expressions" references `(amt amounts)`.  And yes you can use any expression after the variable name, where `amounts` occurs in this example.

How about:
```
The list immediately following `*for*` contains pairs of variable names and template expressions.  Here,
each value from the template `amounts` is given the name `amt` before the `price` invocation is expanded.
```

src/macros-by-example.adoc
toddjonker: Not as specified.  The shorter streams could start producing void, but I think that would get quite complicated to handle properly and to debug.  Do you have a use case in mind?

popematt: I have a sort-of use case—I can't think of a use case where we would want to _silently_ drop the end of the longer stream, and so being able to see that there are more elements remaining in stream could allow me (as a macro author) to raise an error when it occurs.

Is there any particular reason for it _not_ to be an error when the streams are different lengths?

src/macros-by-example.adoc
toddjonker: TBH as soon as this example came together I saw the inconsistency and thought that its the _signature_ that's the problem here.

This `for` is following longstanding Lisp convention, in fact this syntax is [lifted from Racket](https://docs.racket-lang.org/reference/for.html). I'm inclined to continue that tradition, because it's proven itself worthy in terms of extensibility (meaning there's lots of syntactic "room" to add new capabilities to existing forms) and consistency (meaning similarly-behaving forms will have similar syntax).  Relevant here is the convention that variable declarations (ie binding forms) put the variable name first, followed by the expression to be evaluated.

The current signature syntax goes against this principle, following C-like conventions. If the order of macro parameter-specs were reversed (eg strawman `[(currency symbol)]` then these would be consistent in following in idiom that's more likely to be usable for the other forms we need to add.

I'm open to considering other syntax adornments, but it requires a fair amount of thought to ensure its not painting ourselves into a corner and that the new idiom will be generally applicable in other parts of the language where similar needs arise.

src/macros-by-example.adoc
toddjonker: It's worth considering making uneven lengths an error. I cut issue #201 to track the idea.

FWIW I've been trying to avoid adding user-space error-handling mechanisms; I think its a sign that the language (or a macro author) is trying to do too much. It also opens the door to complexity necessary to do things like construct error messages, more-complex predicates, etc.

src/macros-by-example.adoc
toddjonker: TBH as soon as this example came together I thought that its the _signature_ that's the problem here.

This is following longstanding Lisp convention, in fact this syntax is [lifted from Racket](https://docs.racket-lang.org/reference/for.html). I'm inclined to continue that tradition, because it's proven itself worthy in terms of extensibility (meaning there's lots of syntactic "room" to add new capabilities to existing forms) and consistency (meaning similarly-behaving forms will have similar syntax).  Relevant here is the convention that variable declarations (ie binding forms) put the variable name first, followed by the expression to be evaluated.

The current signature syntax goes against this principle, following C-like conventions. If the order of macro parameter-specs were reversed (eg strawman `[(currency symbol)]` then these would be consistent in following in idiom that's more likely to be usable for the other forms we need to add.

almann (approved):
🚢

src/macros-by-example.adoc
popematt: What I was trying to say is that the `(inner_binding expression)` syntax looks an awful lot like the `(type param_name)` syntax (regardless of order). The fact that both are a list of 2-element s-expressions means that they look similar. Adding a keyword—or _anything_—to the text format of the `for` binding (e.g. `(inner_binding of outer_var)`) provides a local contextual clue. Sure, it's not _necessary_, but I'm thinking from an ease-of-use perspective.

I'm not trying to advocate for any particular syntax here—I just think it's worth considering the tradeoffs between purity/minimalism and approachability.

toddjonker: I get that, and will close by saying that the similar shape is an intentional syntactic pattern.  It's safe to say that different folks have different feelings about whether the minimalism is a benefit or a drawback.

Since all of this syntax is provisional and nothing's been implemented, I ask you bear with your (reasonable!) syntactic reservations as more of the language comes up for review. If things still don't gel in your mind, we can have the discussion with the full grammar at hand.

popematt (approved):

toddjonker:
Thanks for the review! I don't have write-access to this repo, so please merge.

PR #199, opened 2023-04-14
Added a glossary.

toddjonker:
### Description of changes:
Certainly incomplete, but a good start.

----

### DIFF
--- a/src/glossary.adoc
+++ b/src/glossary.adoc
@@ -24,7 +24,7 @@ The association of a name with an entity (for example, a module or macro). See a
 all declarations are definitions: some introduce new names for existing entities.
 
 definition::
-The specification of a new entity. Most (but not all) definitions are declarations.
+The specification of a new entity.
 
 directive::
 A unit of data in an Ion document that affects the encoding context, and thus the way the
@@ -45,7 +45,7 @@ The 1.0 `$ion_symbol_table` directive is effectively a less capable alternative 
 
 encoding environment::
 The context-specific data maintained by an Ion implementation while encoding or decoding data. In
-Ion 1.0 this consists of the local symbol table; in Ion 1.1 it also has the local module table
+Ion 1.0 this consists of the local symbol table; in Ion 1.1 this is expanded to include the local module table
 and a collection of available modules.
 
 encoding expression::
--- a/src/glossary.adoc
+++ b/src/glossary.adoc
@@ -1,0 +1,116 @@
+[[sec:glossary]]
+[glossary]
+= Glossary
+
+[glossary]
+
+arity::
+The number of arguments declared by or passed to a macro.  A macro can be _fixed_ or _variable_
+arity; the number of subforms in an invocation is its _actual arity_.
+
+argument::
+A single subform within an encoding expression, corresponding to one of the macro’s parameters.
+This is a one-to-one relation except in the case of rest parameters, where any number of trailing
+arguments are passed to the same parameter.
+
+cardinality::
+The number of elements accepted by a single argument of a macro, or produced by a macro. The
+arguments and results of a macro are streams, and each can be constrained to have exactly-one
+(`*!*`), zero-or-one (`*?*`), zero-or-more (`*{asterisk}*`, `*\...*`), or one-or-more (`*+*`)
+elements.
+
+declaration::
+The association of a name with an entity (for example, a module or macro). See also _definition_. Not
+all declarations are definitions: some introduce new names for existing entities.
+
+definition::
+The specification of a new entity. Most (but not all) definitions are declarations.
+
+directive::
+A unit of data in an Ion document that affects the encoding context, and thus the way the
+document’s data is decoded. In Ion 1.0 there are two directives: the Ion Version Marker, and
+top-level `$ion_symbol_table` structs. Ion 1.1 adds the `$ion_encoding` directive.
+
+document::
+A stream of octets conforming to either the Ion text or binary specification. Can consist of
+multiple _segments_, perhaps using varying versions of the Ion specification. A document doesn’t
+necessarily exist as a file, and isn’t necessarily finite.
+
+E-expression::
+See _encoding expression_.
+
+encoding directive::
+In Ion 1.1, a top-level `$ion_encoding` struct that controls the segment immediately following it.
+The 1.0 `$ion_symbol_table` directive is effectively a less capable alternative syntax.
+
+encoding environment::
+The context-specific data maintained by an Ion implementation while encoding or decoding data. In
+Ion 1.0 this consists of the local symbol table; in Ion 1.1 it also has the local module table
+and a collection of available modules.
+
+encoding expression::
+The invocation of a macro in encoded data, aka E-expression. Starts with a macro reference
+denoting the function to invoke. The Ion text format uses “smile syntax” `(:__macro__ …)` to denote
+E-expressions. Ion binary devotes a large number of opcodes to E-expressions, so they can be
+compact.
+
+macro::
+A transformation function that accepts some number of streams of values, and produces a stream of
+values.
+
+macro definition::
+Specifies a macro in terms of a _signature_ and a _template_.
+
+macro reference::
+Identifies a macro for invocation, renaming, or exporting. Must always be unambiguous. Lexically
+scoped, and never a “forward reference” to a macro that’s declared later in the document.
+
+module::
+The data entity that defines and exports both symbols and macros. Modules are imported by
+encoding directives then installed into the local symbol and/or macro tables.
+
+parameter::
+A named input to a macro, as defined by its signature. At expansion time a parameter produces a
+stream of values.
+
+qualified macro reference::
+A macro reference that consists of a module name and either a macro name exported by that module,
+or a numeric address within the range of the module’s exported macro table. In text, these look
+like `:__module-name__:__name-or-address__`.
+
+rest parameter::
+A macro parameter—always the final parameter—declared with the `*\...*` modifier, that accepts all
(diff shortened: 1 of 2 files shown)


src/glossary.adoc
```diff
@@ -0,0 +1,116 @@
...
+like `:__module-name__:__name-or-address__`.
+
+rest parameter::
+A macro parameter—always the final parameter—declared with the `*\...*` modifier, that accepts all
+remaining arguments to the macro. Similar to “varargs” parameters in Java and other languages.
+
+segment::
+A contiguous portion of a document that uses the same encoding environment. Segment boundaries
+are caused by directives: an IVM starts a new segment, while `$ion_symbol_table` and `$ion_encoding`
+directives end segments (with a new one starting immediately afterwards).
```
popematt: Doesn't an IVM also end segments? (In Ion 1.0, it effectively resets the symbol table, right?)

popematt (commented):
A few general comments:

* It would be good to expand all of the contractions to make the document more accessible.
* Just curious—is it possible to have intra-page links within the glossary so that, e.g. under the _argument_ entry, you could click on _rest parameters_ to jump to that definition?
* Can you add definitions for _subform_ and _opcode_? Both of those terms are used in the glossary, but would not necessarily be commonly known to the target audience.

src/glossary.adoc
```diff
@@ -0,0 +1,116 @@
+[[sec:glossary]]
+[glossary]
+= Glossary
+
+[glossary]
+
+arity::
+The number of arguments declared by or passed to a macro.  A macro can be _fixed_ or _variable_
+arity; the number of subforms in an invocation is its _actual arity_.
```
almann: Is this just the count of all arguments or is this the number of parameters that a macro has?

src/glossary.adoc
```diff
@@ -0,0 +1,116 @@
...
+
+argument::
+A single subform within an encoding expression, corresponding to one of the macro’s parameters.
+This is a one-to-one relation except in the case of rest parameters, where any number of trailing
+arguments are passed to the same parameter.
```
almann: I don't think this is quite right givven our current vernacular.  Isn't an argument a sub-expression bound to a parameter, and each parameter may have zero or more arguments depending on its _argument cardinality_?

src/glossary.adoc
```diff
@@ -0,0 +1,116 @@
...
+
+cardinality::
+The number of elements accepted by a single argument of a macro, or produced by a macro. The
+arguments and results of a macro are streams, and each can be constrained to have exactly-one
+(`*!*`), zero-or-one (`*?*`), zero-or-more (`*{asterisk}*`, `*\...*`), or one-or-more (`*+*`)
+elements.
```
almann: I think this is _argument cardinality_ and we also have _stream cardinality_ which is the number of values that are associated with a parameter (really any evaluation).

src/glossary.adoc
```diff
@@ -0,0 +1,116 @@
...
+
+declaration::
+The association of a name with an entity (for example, a module or macro). See also _definition_. Not
+all declarations are definitions: some introduce new names for existing entities.
+
+definition::
+The specification of a new entity. Most (but not all) definitions are declarations.
```
almann: *Minor*: Editorial suggestion, I don't think it is adding much to say most definitions are declarations (and it technically may not be true).

```suggestion
The specification of a new entity.
```

src/glossary.adoc
```diff
@@ -0,0 +1,116 @@
...
+
+directive::
+A unit of data in an Ion document that affects the encoding context, and thus the way the
+document’s data is decoded. In Ion 1.0 there are two directives: the Ion Version Marker, and
+top-level `$ion_symbol_table` structs. Ion 1.1 adds the `$ion_encoding` directive.
```
almann: I contend that the IVM is not data, so I would not say a unit of data.  Maybe "syntax or unit of data?"

src/glossary.adoc
```diff
@@ -0,0 +1,116 @@
...
+See _encoding expression_.
+
+encoding directive::
+In Ion 1.1, a top-level `$ion_encoding` struct that controls the segment immediately following it.
+The 1.0 `$ion_symbol_table` directive is effectively a less capable alternative syntax.
+
+encoding environment::
+The context-specific data maintained by an Ion implementation while encoding or decoding data. In
+Ion 1.0 this consists of the local symbol table; in Ion 1.1 it also has the local module table
+and a collection of available modules.
```
almann: *Minor*: Editorial suggestion

```suggestion
encoding environment::
The context-specific data maintained by an Ion implementation while encoding or decoding data. In
Ion 1.0 this consists of the local symbol table; in Ion 1.1 this is expanded to include the local module table
and a collection of available modules.
```

src/glossary.adoc
```diff
@@ -0,0 +1,116 @@
...
+encoding directives then installed into the local symbol and/or macro tables.
+
+parameter::
+A named input to a macro, as defined by its signature. At expansion time a parameter produces a
+stream of values.
+
+qualified macro reference::
+A macro reference that consists of a module name and either a macro name exported by that module,
+or a numeric address within the range of the module’s exported macro table. In text, these look
+like `:__module-name__:__name-or-address__`.
```
almann, line 104: The *`:name-or-address`* form is also a qualified reference, yes?  Is this maybe a _local qualified macro reference_?

almann (changes requested):
Looks pretty good--I have some minor questions/suggestions below.

src/glossary.adoc
toddjonker: I guess we need to define "data"!

src/glossary.adoc
toddjonker: There's an unanswered question lurking in this terminology.
```
(public mac [(any* p), (any... q)] /*template*/)
(:mac [elt1, elt2] elt3 elt4)
```
Are `elt1` and `elt2` each arguments, or is `[elt1, elt2]` the argument?  If the former is correct, what do we call the latter?  And what about the rest-arguments?

I've failed to coalesce on terminology for all the parts here.

src/glossary.adoc
toddjonker: I'm not sure that "local" works well, since `:name` could refer to any module, including `$ion`.

src/glossary.adoc
toddjonker: My thinking is that a document that starts with an IVM has that IVM within its first segment, and it defines the encoding version of that segment.

Otherwise, those bytes are the end of the preceding segment, which could be a different version. Since all Ion docs are 1.0 at the start, then a leading 1.1 IVM is part of an Ion 1.0 segment. That seems odd to me.

almann (approved):
🚢 LGTM

src/glossary.adoc
popematt: I didn't make myself clear. An IVM is not the end of a segment; it causes the current segment to end. So rather than make a distinction between IVMs and encoding directives, why not say something like this?

> An IVM or encoding directive _causes_ the current segment to end and _is_ the start of a new segment. 

I _think_ it's the same in practice as what you said, but it avoids special-casing them. 

(Technically, my phrasing _might_ imply that every Ion text document starts with a possibly empty Ion 1.0 segment, but I don't think that's a problem since implementations have to treat it that way anyway.)

toddjonker:
FYI: There's still discussion and editing needed here but I'm prioritizing the module-example document #205 for the moment.

src/glossary.adoc
toddjonker: A segment is a non-overlapping, contiguous range of bytes, and the bytes encoding a directive lie within exactly one segment.  I'm trying to be clear about which segment a directive resides in, because that determines how that directive is encoded. 

IVMs are inherently a special case compared to other directive, because they are not encoded as data but as a special keyword (in text) or opcode (in binary).   And because `$ion_1_1` is not valid keyword in 1.0, it must necessarily be part of a new segment.

I think the statement "every Ion text document starts with a possibly empty Ion 1.0 segment" is true and reasonable, assuming that we allow for empty segments.  Consider:
```
// first line
$ion_1_1  // second line
// third line
```
The first line is within a 1.0 segment, because that's the default for Ion documents, and the second and third lines are within a 1.1 segment.

PS: I'm not fond of "causes" because it implies an active role to passive data.  I think it's better to focus on the _meaning_ (denotation) of the data and how its structured.  (I'm sure I still use active phrasing, but I try to avoid it.)

src/glossary.adoc
almann: Here is my current thinking (which may not be right):

* `elt1`, `elt2`, `elt3` are all _arguments_.  `elt1`, for example is a single _argument_.
* `[elt1, elt2]` (the grouping) are the _arguments_ for a _parameter_.  I think this could be shorthanded as _parameter arguments_ or maybe _parameter argument group_.

toddjonker: Cool, that's aligned with my thinking.  I like "argument group" a lot, and need to see how it flows with the related narratives.

src/glossary.adoc
almann: True, but isn't it unambiguous when `:name` refers to a macro versus a module (I cannot invoke a module)?  What do we call that sort of identifier?

For *macro references*, I see three kinds:

* _unqualified macro references_ - these are denoted without a `:` prefix.  E.g., `foobar`.
* _qualified macro references_ - these are denoted with _module name_ and prefixed/delimited with `:`.  E.g., `:something:foobar`
* _partially qualified macro references_ (not sold on that term) - these are denoted with the `:` prefix and no module name.  Such references are resolved against the _current macro table_.  E.g., `:foobar`

PR #203, opened 2023-04-20
Use Unicode arrows consistently.

toddjonker:
While AsciiDoc will replace `=>` with ⇒ automatically, it doesn't do so in all contexts, so using the Unicode character is more foolproof.

----


zslayton (approved):

toddjonker:
@zslayton Could you pull this please?

PR #205, opened 2023-04-22
Add first portion of new "Modules By Example" chapter.

toddjonker:
This covers encoding context, module basics, and installation of symbols.

----

### DIFF
--- a/src/modules-by-example.adoc
+++ b/src/modules-by-example.adoc
@@ -16,8 +16,7 @@ context maintained by an Ion implementation while encoding or decoding a documen
 The Ion 1.0 encoding environment is just the local symbol table.
 
 The encoding environment is controlled by _directives_ embedded in the document at top-level.
-These directives, also known as “system values”, are encoding artifacts and are not part of the
-application data model.
+These are encoding artifacts and are not part of the application data model.
 
 Ion 1.0 has two forms of directives:
 
@@ -29,7 +28,8 @@ struct closes.
 A _segment_ is a contiguous portion of a document that uses the same encoding environment.
 Segment boundaries are caused by directives: an IVM starts a new segment, while an
 `$ion_symbol_table` struct ends a segment, defining a new one that starting immediately
-afterwards.
+afterwards.  As a result, non-IVM directives are always encoded using the environment of the
+segment that contains them.
 
 TODO Ion text docs always start with a 1.0 segment until an IVM is encountered.
 
@@ -41,10 +41,11 @@ The logical interface to a module has three main components: a spec version, a s
 table, and a macro table.
 
 A module's _spec version_ indicates which Ion specification it uses. This ensures the module has
-stable semantics over time.  Spec versions are expressed as Ion version markers, and a module can
+stable semantics over time.  A module can
 only be used in segments encoded with that version or later.
 
-// TODO The above may be too strict; use solely for symbols could be more relaxed.
+IMPORTANT: **Discussion**: The above may be too strict; use solely for symbols could be more
+relaxed.
 
 A module's _exported symbol table_ is simply a sequence of strings. These denote the
 text of symbols, and are equivalent in meaning to the `symbols` list of an Ion 1.0 shared symbol
@@ -68,11 +69,14 @@ declare symbols and not macros.
 
 === Ion 1.1 Encoding Environment
 
-In Ion 1.1, the encoding environment is enhanced to include:
+In Ion 1.1, the encoding environment includes:
 
   * The current Ion version, because a document may have segments using different Ion versions.
   * The _available modules_, a name to module mapping.
+  * The local symbol table, assembled from a subset of the available modules.
   * The local macro table, assembled from a subset of the available modules.
+
+NOTE: In Ion 1.0, the local symbol table _is_ the encoding environment.
 
 Upon encountering the `$ion_1_1` IVM, the environment is reset to the default state, in which:
 
@@ -171,7 +175,7 @@ back to 1.0. First, turn the `local` module into the equivalent shared symbol ta
 [{nrm}]
 ----
 **$ion_shared_symbol_table**::{
-  name: "local",
+  name: "extracted",
   version: 1,
   symbols: ["s1", "s2"]
 }
@@ -183,11 +187,11 @@ Then translate `install_symbols:[local]` into its 1.0 equivalent:
 ----
 *$ion_1_0*
 **$ion_symbol_table**::{
-  imports: [{ name: "local", version: 1, max_id: 2 }]
+  imports: [{ name: "extracted", version: 1, max_id: 2 }]
 }
 ----
 
-NOTE:  Even ignoring Ion 1.1, this is how you would hoist local symbols into a
+NOTE:  Even ignoring Ion 1.1, this is how you would extract local symbols into a
 new shared symbol table.
 
 The latter imports-only document has this 1.1 equivalent:
@@ -198,7 +202,7 @@ The latter imports-only document has this 1.1 equivalent:
 *$ion_1_1*
 **$ion_encoding**::{
   modules: [
-    (*import* local "local" 1 2)
+    (*import* local "extracted" 1 2)
   ],
   install_symbols: [ local ]
 }
@@ -208,15 +212,14 @@ Here we see a new form inside the `modules` field that imports a module into the
 environment and assigns it a name.
 The `*import*` keyword starts an S-expression that expects three or four arguments. The first is
 a symbolic name that we can use later to refer to the imported module.  The remaining arguments
-are effectively the `name`, `version` and `max_id` fields of the 1.0 `imports` struct.
+are effectively the `name`, `version` and `max_id` fields of the 1.0 `imports` struct, with only
+the max_id being optional in this form.
 
 TIP: From the perspective of Ion 1.1, shared symbol tables _are_ encoding modules.
 
 
 === Declaring Multiple Modules
 
-Within a directive's `modules` field, both `*module*` and `*import*` clauses have the same
-ultimate effect, declaring a name->module mapping for the encoding environment.
 Let’s look at a scenario with both imported and locally-defined symbols:
 
 [{nrm}]
@@ -242,8 +245,8 @@ Here’s the Ion 1.1 equivalent in terms of symbol allocation order:
 }
 ----
 
-Just as in the 1.0 version, this allocates symbol IDs starting with ten from `shared1`, twenty
-from `shared2`, then the two locally-defined symbols.
+Just as in the 1.0 version, this allocates ten symbol IDs for `shared1` (as requested by its
+max_id argument), twenty symbol IDs for `shared2`, then the two locally-defined symbols.
 
 By decoupling symbol-table importing from installation, Ion 1.1 allows some encoding techniques
 that are not possible in 1.0.  For example, we can give local symbols smaller IDs than imported
@@ -316,4 +319,3 @@ Here again, Ion 1.1 enables a new technique: we can prepend new symbols to the e
   install_symbols: [local2, local]                    // 'local2' is first
 }
 ----
-
--- a/src/modules-by-example.adoc
+++ b/src/modules-by-example.adoc
@@ -1,0 +1,320 @@
+[[sec:moduleexample]]
+== Modules by Example
+
+:nrm: subs="+normal"
+
+Within the spectrum of Ion 1.1's new encoding facilities, macros lie at the bottom, affecting the
+serialized byte-stream.  The prior chapter explored macro definitions while ignoring the contexts
+within which those definitions exist.  This chapter covers that context top-down.
+
+
+=== Encoding Environment 1.0
+
+An Ion _document_ is a stream of octets conforming to either the Ion text or binary specification.
+(For our purposes here, a document does not necessarily exist as a file, and isn’t necessarily
+finite.)  The interpretation of those octets is guided by an _encoding environment_, the
+context-specific data maintained by an Ion implementation while encoding or decoding a document.
+The Ion 1.0 encoding environment is effectively just the local symbol table.
+
+The encoding environment is controlled by _directives_ embedded in the document at top-level.
+These directives, also known as “system values”, are encoding artifacts and are not part of the
+application data model.
+
+Ion 1.0 has two forms of directives:
+
+* An Ion Version Marker (IVM) resets the environment to the default provided by that version of
+Ion.
+* An `$ion_symbol_table` struct defines a new environment that takes effect immediately after the
+struct closes.
+
+A _segment_ is a contiguous portion of a document that uses the same encoding environment.
+Segment boundaries are caused by directives: an IVM starts a new segment, while an
+`$ion_symbol_table` struct ends a segment, defining a new one that starting immediately
+afterwards.
+
+TODO Ion text docs always start with a 1.0 segment until an IVM is encountered.
+
+
+=== Modules from the Outside
+
+In Ion 1.1, you define, share, and install symbols and macros using _encoding modules_.
+The logical interface to a module has three main components: a spec version, a symbol
+table, and a macro table.
+
+A module's _spec version_ indicates which Ion specification it uses. This ensures the module has
+stable semantics over time.  Spec versions are expressed as Ion version markers, and a module can
+only be used in segments encoded with that version or later.
+
+// TODO The above may be too strict; use solely for symbols could be more relaxed.
+
+A module's _exported symbol table_ is simply a sequence of strings. These denote the
+text of symbols, and are equivalent in meaning to the `symbols` list of an Ion 1.0 shared symbol
+table.
+
+A module's _exported macro table_ is a sequence of <name, macro> pairs.  Names can be
+null, in which case the corresponding macro can be referenced by its zero-based index in the
+table, known as its _exported address_.  Non-null names in the table must all be unique, so that
+a name-to-macro mapping function is well-defined.
+
+TIP: Macros have their own "object identity" independent of the names that map to them.  It's
+possible for the same macro to have multiple addresses and/or names.
+
+To reuse macros across documents, _shared modules_ subsume the capabilities
+of shared symbol tables while remaining backwards-compatible with their current schema and catalog
+semantics.
+
+IMPORTANT: Any existing Ion shared symbol table is a valid encoding module, albeit one that only
+declares symbols and not macros.
+
+
+=== Encoding Environment 1.1
+
+In Ion 1.1, the encoding environment is enhanced to include:
+
+  * The current Ion version, because a document may have segments using different Ion versions.
+  * The _available modules_, a name->module mapping.
+  * The local macro table, assembled from a subset of the available modules.
+
+Upon encountering the `$ion_1_1` IVM, the environment is reset to the default state, in which:
+
+  * The current Ion version is 1.1.
+  * The available modules contains only the `$ion` module, version 2 (v1 being Ion 1.0).
+  * The current local macro table is empty.
+  * The current symbol table is the Ion 1.1 system symbol table.
+
+To customize this environment, we use an _encoding directive_: a
+top-level struct annotated with `$ion_encoding`. Like `$ion_symbol_table`, this directive defines a
+new encoding environment that goes into effect immediately after the struct closes.
+
+The general syntax of an encoding directive is as follows:
+
+[{nrm}]
+----
+**$ion_encoding**::{
+  modules:         [ /* module declarations \*/ ],
+  install_symbols: [ /* names declared in 'modules' field \*/ ],
+  install_macros:  [ /* names declared in 'modules' field */ ]
+}
+----
+
+The three fields here correspond directly to the main components of the environment.
+The `modules` field enumerates the available modules, while the `install_symbols`
+and `install_macros` fields specify the layout of the local symbol and macro tables.
+
+Let’s look at some examples illustrating the relation between `$ion_symbol_table` and
+`$ion_encoding`.
+
+
+=== Defining Local Symbols
+
+The most basic Ion encoding scenario uses only locally-defined symbols.
+In Ion 1.0, this is expressed as follows:
+
+[{nrm}]
+----
+*$ion_1_0*
+**$ion_symbol_table**::{
+  symbols: ["s1", "s2"]
+}
+----
+
+Here’s an Ion 1.1 document that’s equivalent, in the sense that it allocates symbol IDs in the
+same order. (The IDs will be different, though, due to new system symbols.)
+
+[{nrm}]
+----
+*$ion_1_1*
+**$ion_encoding**::{
+  modules: [
+    (*module* local
+      (*symbols* [ "s1", "s2" ]))
+  ],
+  install_symbols: [ local ]
+}
+----
+
+The definition of the local symbol table has been refactored into two parts. First, the list of
+symbols is expressed inside a module named `local`. Then, the symbols from that module are
+installed to form the new local symbol table. Compared to the behavior of `$ion_symbol_table`,
+this is akin to defining a named symbol table “inline” to hold local symbols, then defining the
+local symbol table only via `imports` and no `symbols` field.
+
+Let's look more closely at the definition of `local`:
+
+[{nrm}]
+----
+(*module* local
+  (*symbols* ["s1", "s2"]))
+----
+
+The `*module*` keyword starts an S-expression that defines a new module with the given name.
+The `*symbols*` keyword starts a subform that defines the module's exported symbol table.
+This clause expects a list of strings, and both the syntax and semantics are the same as the
+`symbols` field of `$ion_shared_symbol_table`.
+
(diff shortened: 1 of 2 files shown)


src/modules-by-example.adoc
```diff
@@ -0,0 +1,320 @@
...
+Upon encountering the `$ion_1_1` IVM, the environment is reset to the default state, in which:
+
+  * The current Ion version is 1.1.
+  * The available modules contains only the `$ion` module, version 2 (v1 being Ion 1.0).
+  * The current local macro table is empty.
+  * The current symbol table is the Ion 1.1 system symbol table.
+
+To customize this environment, we use an _encoding directive_: a
+top-level struct annotated with `$ion_encoding`. Like `$ion_symbol_table`, this directive defines a
+new encoding environment that goes into effect immediately after the struct closes.
```
popematt, line 90: Does this mean that things in the encoding directive cannot be used within the encoding directive itself?

src/modules-by-example.adoc
```diff
@@ -0,0 +1,320 @@
...
+
+Here’s an Ion 1.1 document that’s equivalent, in the sense that it allocates symbol IDs in the
+same order. (The IDs will be different, though, due to new system symbols.)
+
+[{nrm}]
+----
+*$ion_1_1*
+**$ion_encoding**::{
+  modules: [
+    (*module* local
```
popematt, line 132: Is there anything special about the name `local`, or is it just convention, or even just a convenient name since we were just talking about Ion 1.0 local symbol tables?

src/modules-by-example.adoc
```diff
@@ -0,0 +1,320 @@
...
+  ],
+  install_symbols: [ local ]
+}
+----
+
+Here we see a new form inside the `modules` field that imports a module into the encoding
+environment and assigns it a name.
+The `*import*` keyword starts an S-expression that expects three or four arguments. The first is
+a symbolic name that we can use later to refer to the imported module.  The remaining arguments
+are effectively the `name`, `version` and `max_id` fields of the 1.0 `imports` struct.
```
popematt: If it's "three or four", which one(s) are optional?

src/modules-by-example.adoc
```diff
@@ -0,0 +1,320 @@
...
+*$ion_1_1*
+**$ion_encoding**::{
+  modules: [(*import* n1 "shared1" 1 10),
+            (*import* n2 "shared2" 2 20),
+            (*module* local (*symbols* ["s1", "s2"]))],
+  install_symbols: [n1, n2, local]
+}
+----
+
+Just as in the 1.0 version, this allocates symbol IDs starting with ten from `shared1`, twenty
```
popematt: Just a thought—consider whether something like "starting with SID 10" or "starting with symbol id `$10`" might be easier to read/parse than "starting with ten". My mind tries to continue "starting with ten _of something_", but upon re-reading it a few times, I think you are referring to the number ten.

src/modules-by-example.adoc
```diff
@@ -0,0 +1,320 @@
+[[sec:moduleexample]]
+== Modules by Example
+
+:nrm: subs="+normal"
+
+Within the spectrum of Ion 1.1's new encoding facilities, macros lie at the bottom, affecting the
+serialized byte-stream.  The prior chapter explored macro definitions while ignoring the contexts
+within which those definitions exist.  This chapter covers that context top-down.
```
almann: I am not sure what the first sentence is trying to say, (especially "macros lie at the bottom").  I would cut this sentence all together.

```suggestion
The prior chapter explored macro definitions while ignoring the contexts
within which those definitions exist.  This chapter covers that context top-down.
```

src/modules-by-example.adoc
```diff
@@ -0,0 +1,320 @@
...
+
+
+=== Encoding Environment 1.0
+
+An Ion _document_ is a stream of octets conforming to either the Ion text or binary specification.
+(For our purposes here, a document does not necessarily exist as a file, and isn’t necessarily
+finite.)  The interpretation of those octets is guided by an _encoding environment_, the
+context-specific data maintained by an Ion implementation while encoding or decoding a document.
+The Ion 1.0 encoding environment is effectively just the local symbol table.
```
almann: *Minor*: Editorial suggestion.

```suggestion
context maintained by an Ion implementation while encoding or decoding a document.
The Ion 1.0 encoding environment is just the local symbol table.
```

src/modules-by-example.adoc
```diff
@@ -0,0 +1,320 @@
...
+
+The encoding environment is controlled by _directives_ embedded in the document at top-level.
+These directives, also known as “system values”, are encoding artifacts and are not part of the
```
almann: FWIW, I do not consider the IVM a "system value," but I do think it is a directive maybe we just cut the parenthetical aside?

src/modules-by-example.adoc
```diff
@@ -0,0 +1,320 @@
...
+
+* An Ion Version Marker (IVM) resets the environment to the default provided by that version of
+Ion.
+* An `$ion_symbol_table` struct defines a new environment that takes effect immediately after the
+struct closes.
+
+A _segment_ is a contiguous portion of a document that uses the same encoding environment.
+Segment boundaries are caused by directives: an IVM starts a new segment, while an
+`$ion_symbol_table` struct ends a segment, defining a new one that starting immediately
+afterwards.
```
almann: Is it worth mentioning that the non-IVM directives evaluate/process in the segment that contains them.  I think this is not a particularly intuitive aspect of Ion.

src/modules-by-example.adoc
```diff
@@ -0,0 +1,320 @@
...
+
+=== Modules from the Outside
+
+In Ion 1.1, you define, share, and install symbols and macros using _encoding modules_.
+The logical interface to a module has three main components: a spec version, a symbol
+table, and a macro table.
+
+A module's _spec version_ indicates which Ion specification it uses. This ensures the module has
+stable semantics over time.  Spec versions are expressed as Ion version markers, and a module can
+only be used in segments encoded with that version or later.
```
almann: > Spec versions are expressed as Ion version markers

This is a confusing clause.  Are we saying that spec versions are defined by the Ion version of the segment the module is defined in, or is this  explicitly something in the module's definition?  IVMs only exist at the top-level so I assume the former, but I can't tell reading this.

src/modules-by-example.adoc
```diff
@@ -0,0 +1,320 @@
...
+
+// TODO The above may be too strict; use solely for symbols could be more relaxed.
```
almann: Maybe expose as  discussion point call out?

src/modules-by-example.adoc
```diff
@@ -0,0 +1,320 @@
...
+
+A module's _exported symbol table_ is simply a sequence of strings. These denote the
+text of symbols, and are equivalent in meaning to the `symbols` list of an Ion 1.0 shared symbol
+table.
+
+A module's _exported macro table_ is a sequence of <name, macro> pairs.  Names can be
+null, in which case the corresponding macro can be referenced by its zero-based index in the
+table, known as its _exported address_.  Non-null names in the table must all be unique, so that
+a name-to-macro mapping function is well-defined.
```
almann, line 57: I am a little uncomfortable with "null" here because that is a value in Ion and also symbols have *undefined* text (which are the syntax of these things).  I wonder if we should use a term like "undefined" which could be encoded as not specified in the definition, a literal `null`, or even `$0` if we so choose.

src/modules-by-example.adoc
```diff
@@ -0,0 +1,320 @@
...
+
+TIP: Macros have their own "object identity" independent of the names that map to them.  It's
+possible for the same macro to have multiple addresses and/or names.
```
almann, line 60: I really am uncomfortable with the term "object identity."  Is it even germane for anything that the user experiences?  In other words, why is this not an implementation detail?

I only think we're saying that Macros may be aliases by more than one name.  Anything else is implementation detail including re-using such a definition internally or copying it if an implementation so chooses.

src/modules-by-example.adoc
```diff
@@ -0,0 +1,320 @@
...
+
+To reuse macros across documents, _shared modules_ subsume the capabilities
+of shared symbol tables while remaining backwards-compatible with their current schema and catalog
+semantics.
+
+IMPORTANT: Any existing Ion shared symbol table is a valid encoding module, albeit one that only
+declares symbols and not macros.
```
almann: *Minor*: Editorial suggestion.

```suggestion
IMPORTANT: All existing Ion shared symbol tables **are** encoding modules.  Such modules only
declare symbols and not macros.
```

src/modules-by-example.adoc
```diff
@@ -0,0 +1,320 @@
...
+
+
+=== Encoding Environment 1.1
```
almann: *Minor*: Editorial suggestion.

```suggestion
=== Ion 1.1 Encoding Environment
```

src/modules-by-example.adoc
```diff
@@ -0,0 +1,320 @@
...
+
+In Ion 1.1, the encoding environment is enhanced to include:
+
+  * The current Ion version, because a document may have segments using different Ion versions.
+  * The _available modules_, a name->module mapping.
```
almann: *Minor*: Editorial suggestion.

```suggestion
  * The _available modules_, a name to module mapping.
```

src/modules-by-example.adoc
```diff
@@ -0,0 +1,320 @@
...
+  * The local macro table, assembled from a subset of the available modules.
```
almann, line 77: I would add the _local symbol table_ here as well and change the sentence that opens this bullet list to just declare what it is (I found it confusing that it wasn't here and I had to re-read it to figure out why).

Maybe something like:

> * The _local symbol table_, assembled from a subset of the available modules.
>
> Note, in Ion 1.0, the local symbol table **is** the encoding environment.

src/modules-by-example.adoc
```diff
@@ -0,0 +1,320 @@
...
+  * The current Ion version is 1.1.
+  * The available modules contains only the `$ion` module, version 2 (v1 being Ion 1.0).
+  * The current local macro table is empty.
+  * The current symbol table is the Ion 1.1 system symbol table.
+
+To customize this environment, we use an _encoding directive_: a
+top-level struct annotated with `$ion_encoding`. Like `$ion_symbol_table`, this directive defines a
+new encoding environment that goes into effect immediately after the struct closes.
+
+The general syntax of an encoding directive is as follows:
```
almann, line 92: Are Ion 1.0 encoding directives allowed in Ion 1.1 (we should say explicitly, unless I missed it)?

src/modules-by-example.adoc
```diff
@@ -0,0 +1,320 @@
...
+}
+----
+
+Here we see a new form inside the `modules` field that imports a module into the encoding
+environment and assigns it a name.
+The `*import*` keyword starts an S-expression that expects three or four arguments. The first is
+a symbolic name that we can use later to refer to the imported module.  The remaining arguments
+are effectively the `name`, `version` and `max_id` fields of the 1.0 `imports` struct.
+
+TIP: From the perspective of Ion 1.1, shared symbol tables _are_ encoding modules.
```
almann, line 218: It might be worth noting that while shared symbol tables are encoding modules, Ion 1.0 local symbol tables are encoding directives (which implicitly have a locally defined module).  This distinction is subtle and not obvious.

src/modules-by-example.adoc
```diff
@@ -0,0 +1,320 @@
...
+
+
+=== Declaring Multiple Modules
+
+Within a directive's `modules` field, both `*module*` and `*import*` clauses have the same
+ultimate effect, declaring a name->module mapping for the encoding environment.
+Let’s look at a scenario with both imported and locally-defined symbols:
```
almann: These paragraph seems a little out of place--I was confused because of the Ion 1.0 example below.  Maybe set up the 1.0 example then move this down.

src/modules-by-example.adoc
```diff
@@ -0,0 +1,320 @@
...
+----
+
+Here’s the Ion 1.1 equivalent in terms of symbol allocation order:
+
+[{nrm}]
+----
+*$ion_1_1*
+**$ion_encoding**::{
+  modules: [(*import* n1 "shared1" 1 10),
+            (*import* n2 "shared2" 2 20),
```
almann, line 242: Is `n1`/`n2` required?  Can they be omitted if `shared1`/`shared2` are valid module names?

src/modules-by-example.adoc
```diff
@@ -0,0 +1,320 @@
...
+**$ion_encoding**::{
+  modules: [(*module* local (*symbols* ["s1", "s2"]))],
+  install_symbols: [local]
+}
+
+// ... application data ...
+
+**$ion_encoding**::{
+  modules: [
+    *{asterisk}*,                                      // retain all available modules
```
almann, line 301: This gets rendered as `*` right?

almann (commented):
Looking really great--comments/questions below.

src/modules-by-example.adoc
popematt: +1

I was going to ask what is the significance of this fact—can an Ion 1.1 encoding directive import an Ion 1.0 shared symbol table?

src/modules-by-example.adoc
popematt: Again, I was going to ask the same thing, but Almann is faster with the publish button. ;)

If it does render as __`*`__, is this special syntax or should it actually be single-quoted like a symbol? (If it's special syntax, we should call it out as such.)

popematt (commented):
I found that the comparisons to Ion 1.0 shared symbol tables throughout the document were not helpful because Ion 1.0 shared symbol tables are something that even I (someone who is employed to work on Ion) infrequently think about, and other people probably even less so. It effectively means that the Ion 1.0 spec is required pre-reading for this document.

I think it would make this chapter (and modules) more approachable if this document could introduce encoding directives/modules without discussing Ion 1.0 (at least not very much). Only after explaining modules in their own right, we could have a "Comparison with Ion 1.0 symbol tables" section.

What do you think?

src/modules-by-example.adoc
toddjonker: I think its more than an implementation detail, since the rule around reference ambiguity requires identity of macros.  Specifically, an alias to a macro has the same identity, as does an re-export from an importing module.  Copying the macro—or otherwise assigning a second identity with the same definition—would introduce ambiguity.

src/modules-by-example.adoc
almann: Yes--but more importantly a few things:

* Shared symbol tables in Ion 1.1 ***are*** shared encoding modules.
* The syntax for shared symbol tables is extended to support macros in Ion 1.1 (we don't have a convention for something like `$ion_shared_module::{...}`
* Ion 1.0 local symbol tables ***are*** encoding directives.

I think we need to be super crisp about these facts (or correct my understanding if these facts are wrong).

src/modules-by-example.adoc
toddjonker: Correct.  ~~An encoding~~ Any non-IVM directive is encoded using the environment of the segment it appears in.  The environment does not change—a new segment does not begin—until immediately after the end of the directive.

src/modules-by-example.adoc
toddjonker: I'm using the term "encoding directive" specifically for this 1.1 entity.  `$ion_symbol_table` and IVMs are different kinds of directives.

I'm not 100% satisfied by that delineation, but it's the best I've got today.  I def want a specific term for `$ion_encoding` so that annotation isn't needed everywhere its discussed.

src/modules-by-example.adoc
toddjonker: It's just a name, nothing special about it.  Here I just used it to connote the alignment with the entity formerly known as the local symbol table.

src/modules-by-example.adoc
toddjonker: I was referring to assignment of ten addressed from shared1 (as per its max_id) and then 20 addresses from shared2 (as per its max_id).   I'll rephrase.

toddjonker: How about:
```
Just as in the 1.0 version, this allocates ten symbol IDs for `shared1` (as requested by its 
max_id argument), twenty symbol IDs for `shared2`, then the two locally-defined symbols.
```

src/modules-by-example.adoc
toddjonker: The symbols are required, and are local names.  The second argument must be a string, just as in an `import` struct.

src/modules-by-example.adoc
toddjonker: Only the final `max_id` argument is optional.  Will update.

src/modules-by-example.adoc
almann: @popematt asks a really good question, `*` would not, unquoted, be valid _Ion_ syntax in a list here.  So is the syntax `'*'` or do we have a special form for this like `(*)` or `(inherit_context)` or whatever.

src/modules-by-example.adoc
toddjonker: Yes, an Ion 1.1 encoding directive can import an Ion 1.0 shared symbol table. That's exactly what's shown by the preceding example.

src/modules-by-example.adoc
toddjonker: `{asterisk}` is an AsciiDoc built-in, and this is the most reliable way I've found to render a bold-face asterisk.

@popematt The Ion data looks like `modules:[*, (module` and the asterisk....  oh heck, I see what you mean, operator-symbols are invalid as list elements, so it would need to be `['*', (module`.   Gross, we might want to change the syntax.

toddjonker: Cut #206 to track that.

popematt: What about `(module *)`? Or, I think that `(*)` could be a fine alternative, as Almann (sort of) suggested.

src/modules-by-example.adoc
toddjonker: Almann I agree with your first two bullets, but ATM the last one isn't correct. In my current terminology, the 1.0 LST struct is not an "encoding directive" but some other kind of directive for which I don't have a crisp term.  To clarify my earlier comment, I want to be able to say "the `modules` field of an encoding directive" and other things that don't apply to LST structs.

src/modules-by-example.adoc
toddjonker: I'll just kill that clause, it's poorly worded and it'll be easier to explain when it shows up later.

src/modules-by-example.adoc
toddjonker: Well, I used "null" b/c the way to define an anonymous macro is by using literal `null` as the name, via a definition like `(macro null ...)`.   So I really did mean null, or perhaps `null.symbol`.

Maybe here we just say "Macros can be anonymous, in which case..."  and then "Names in the table..." ?

toddjonker:
@popematt  From the big-picture perspective, you're absolutely right: we need to have content that's approachable by Ion newcomers and users that don't know all the details.  We're simply "not there yet". At this stage, this doc is very much aimed at stakeholders who understand Ion 1.0 spec and are giving feedback on the 1.1 work in progress.  The first chapter is explicitly "what's new in 1.1" and these tutorial-ish chapters are implicitly similar.

Once some key stakeholders signal alignment with the direction described by this document, we'll start collating the more-formal specification and can also evolve these narratives. (I expect they'll end up as a separate document.)  We have quite a few more PRs before the 1.1 documents are complete.

src/modules-by-example.adoc
toddjonker: Not discussed here is the ability to retain a selected set of available modules.  I'm inclined to go with clauses like `(retain *)` and `(retain mod1 mod3)` for this purpose.  I think a dedicated keyword is worthwhile since the behavior is so different from `import` and `module`.

Also not discussed here is the (tentative, TBD) parallel ability to retain previously-directed `install_symbols` and `install_modules` lists, so they don't need to be explicitly named.  I've also used `*` there (in other docs) and it's similarly defective.

src/modules-by-example.adoc
almann: Can it not be required--to avoid DRY when possible? `(import "shared1" 1 10)` => `(import shared1 "shared1" 1 10)`, I mean not a big deal, but a little annoying.

src/modules-by-example.adoc
almann: Hmm, then I do think we need some clarifying text here--because as it is the dissonance between Ion 1.0 local symbol tables, Ion 1.0/1.1 shared symbol tables (shared modules?), and Ion 1.1 encoding directives are muddled to me, and part of that muddling is the recycling of the shared symbol table schema (which practically makes sense, but really hurts my brain around the mental model)

src/modules-by-example.adoc
almann: I think we need some text that indicates that the LST is not an encoding directive and indicates if such a system value is allowed in Ion 1.1 or not.

almann (commented):
Latest revision looks great--I think there are some outstanding comments from prior review around text vis-a-vis Ion 1.0 local symbol table, Ion 1.0/1.1 shared symbol table/shared modules, and Ion 1.1 encoding directives which is the only reason why I am not hitting approve, but if you'd like to punt on that, I am willing to defer.

toddjonker:
@almann I've got the second half of this chapter ready to go, but I don't think I can submit that PR while this one is open.

I'd like to get this one merged so we can start discussing the second half.  I think the feedback still pending here can be addressed once the doc as a whole is in place.

almann (approved):
Approving since you will pull feedback for follow up PR

src/modules-by-example.adoc
```diff
@@ -0,0 +1,321 @@
...
+[{nrm}]
+----
+(*module* local
+  (*symbols* ["s1", "s2"]))
+----
+
+The `*module*` keyword starts an S-expression that defines a new module with the given name.
+The `*symbols*` keyword starts a subform that defines the module's exported symbol table.
+This clause expects a list of strings, and both the syntax and semantics are the same as the
+`symbols` field of `$ion_shared_symbol_table`.
```
zslayton, line 156: > This clause expects a list of strings....

I wonder if we should expect a list of text values instead, allowing existing symbol IDs to be used in defining a module. `$0` would lead to an entry with unknown text the way that non-`string` entities do today.

src/modules-by-example.adoc
```diff
@@ -0,0 +1,321 @@
...
+  install_symbols: [n1, n2, local]
+}
+----
+
+Just as in the 1.0 version, this allocates ten symbol IDs for `shared1` (as requested by its
+max_id argument), twenty symbol IDs for `shared2`, then the two locally-defined symbols.
+
+By decoupling symbol-table importing from installation, Ion 1.1 allows some encoding techniques
+that are not possible in 1.0.  For example, we can give local symbols smaller IDs than imported
+symbols by installing `local` first:
```
zslayton, line 253: Consider adding a sentence after the example explaining why smaller symbol IDs are beneficial.

src/modules-by-example.adoc
```diff
@@ -0,0 +1,321 @@
...
+
+[{nrm}]
+----
+*$ion_1_1*
+**$ion_encoding**::{
+  modules: [(*import* n1 "shared1" 1 10),
+            (*import* n2 "shared2" 2 20),
+            (*module* local (*symbols* ["s1", "s2"]))]
```
zslayton, line 261: I think we could use a bit of text explicitly stating that `max_id` (when supplied) _only_ impacts symbol ID allocation and has no effect on macros.

src/modules-by-example.adoc
toddjonker: In most cases, symbol table names are long and include qualifying prefixes, since its a global "namespace".  I think few will be usable as identifying symbols.

The syntax could infer the symbolic name, but it only works if the string satisfies the content constraint (ie it only uses identifier characters).  That doesn't work universally, and it's complex to explain, and the resulting schema is irregular in how the local module-names are expressed... so I'm not inclined.

toddjonker: I've submitted #209 which uses more realistic SST names.

src/modules-by-example.adoc
toddjonker: I agree that's worth considering.  Cut #210

src/modules-by-example.adoc
toddjonker: This is discussed in the new content where imported macros are introduced.  See L489 of https://github.com/amazon-ion/ion-docs/pull/208/files

src/modules-by-example.adoc
toddjonker: I think this would be a great topic for our meetup next week.  Resolving in favor of #211

src/modules-by-example.adoc
toddjonker: Resolving in favor of #212 since there's no further examples or explanation of anonymous macros.

src/modules-by-example.adoc
toddjonker: In #214 I change this to simply read "their own identity".

PR #208, opened 2023-04-26
Add module examples covering macros.

toddjonker:
----

### DIFF
--- a/src/modules-by-example.adoc
+++ b/src/modules-by-example.adoc
@@ -329,7 +329,8 @@ define modules that export macros, then you enumerate the modules whose macros y
 install. The lists of exported macros from each of those modules are concatenated to form a
 contiguous address space so that any macro can be referenced by an integer.
 
-We can now define our macros for two-dimensional geometry with full context:
+We can now define a small module for two-dimensional geometry, finally showing macro definitions
+in full context:
 
 [{nrm}]
 ----
@@ -396,6 +397,7 @@ Macros are most useful when they're shared across documents, and for that we use
 a generalization of Ion 1.0's shared symbol tables. As discussed in
 <<_modules_from_the_outside>>, they export both a symbol table and a macro table.
 
+TIP: In Ion 1.1, a shared symbol table _is_ a shared module.
 
 NOTE: We intend to propose a new schema for shared modules, akin to the new `$ion_encoding` schema.
 That should be easier to explain and understand than the format below.
@@ -486,8 +488,9 @@ the name is ambiguous, meaning that two installed modules map it to different ma
 (:point 17 28) ⇒ **error**: ':point' is ambiguous, exported by 'geo' and 'g3d'.
 ----
 
-Another thing to note in the directive used above is that the `**import** g3d` declaration includes a symbol
-table name and version, but no max_id argument.  As with legacy imports, absence of max_id forces
+Another thing to note in the directive used above is that the `**import** g3d` declaration
+includes a symbol table name and version, but no max_id argument.  As with imports in a local
+symbol table, absence of max_id forces
 the Ion implementation to acquire the symbol table entity with exactly the stated version.  While
 this is generally not best-practice for importing symbols, exact-match is a **requirement** for
 using the module in `install_macros`.  In other words, when a document is encoded using macros,
@@ -638,8 +641,7 @@ smile syntax does not apply.  Instead, use a quoted symbol:
 ----
 
 That has the intended effect of keeping `scatterplot` using 2D points, but it's somewhat awkward.
-A more ergonomic approach is to introduce an alias to disambugate:
-
+A more ergonomic approach is to introduce an alias to disambiguate:
 
 [{nrm}]
 ----
@@ -688,7 +690,16 @@ available to consumers of the module, and for that they can be exported:
 ----
 
 Exports can also be used to "pass through" selected macros from an imported module: `(*export*
-':g2d:line')` exports the name `line` from the enclosing module.
+':g2d:line')` exports the name `line` from the enclosing module.  The pass-through form is
+_almost_ the same as the pair of clauses:
+
+[{nrm}]
+----
+(*alias* line ':g2d:line')
+(*export* line)
+----
+
+..._except_ the latter declares a local name while the pass-through does not.
 
 IMPORTANT: The macro names exported by a module must be unique, regardless of whether they are
 exported implicitly via `*macro*` or explicitly via `*export*`.
@@ -729,7 +740,15 @@ the macro table as well:
 === Separate Installation
 
 The preceding example has some repetition between `install_symbols` and `install_macros`,
-illustrating that the symbol and macro tables are maintained independently.  There's no
+illustrating that the symbol and macro tables are maintained independently.
+The following is legal:
+
+----
+  install_symbols: [local, local2]
+  install_macros:  [local2, local]
+----
+
+There's no
 assumption that the document needs both symbols and macros from every module, or that the
 relative allocation of addresses should be the same.  If anything, we assume the opposite: that
 installing the macros from a module suggests that you don't need to install its symbols since
--- a/src/modules-by-example.adoc
+++ b/src/modules-by-example.adoc
@@ -40,6 +40,7 @@ In Ion 1.1, you define, share, and install symbols and macros using _encoding mo
 The logical interface to a module has three main components: a spec version, a symbol
 table, and a macro table.
 
+[#spec-version]
 A module's _spec version_ indicates which Ion specification it uses. This ensures the module has
 stable semantics over time.  A module can
 only be used in segments encoded with that version or later.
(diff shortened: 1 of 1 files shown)


src/modules-by-example.adoc
```diff
@@ -319,3 +320,472 @@ Here again, Ion 1.1 enables a new technique: we can prepend new symbols to the e
...
+
+
+=== Installing and Using Macros
+
+The local macro table works in essentially the same way as the local symbol table: you import or
+define modules that export macros, then you enumerate the modules whose macros you want to
+install. The lists of exported macros from each of those modules are concatenated to form a
+contiguous address space so that any macro can be referenced by an integer.
+
+We can now define our macros for two-dimensional geometry with full context:
```
almann: I don't recall this  context ever being set up in this document so the sentence is really jarring.  What "two-dimensional geometry" are we talking about?  I think you are referring to the by example section, right?

src/modules-by-example.adoc
```diff
@@ -319,3 +320,472 @@ Here again, Ion 1.1 enables a new technique: we can prepend new symbols to the e
...
+        [a, b]))
+  ],
+  install_macros: [ geo ]
+}
+(:point 17 28)
+(:line (1 2) (3 4))
+----
+
+This `geo` module defines macros instead of symbols, using the `*macro*` definition syntax
+explored throughout <<sec:macroexample>>.
```
almann, line 353: Ah, there it is--I think we need to put this above the sentence referring to it.

src/modules-by-example.adoc
```diff
@@ -319,3 +320,472 @@ Here again, Ion 1.1 enables a new technique: we can prepend new symbols to the e
...
+
+The `install_macros` field works much like `install_symbols`: it assembles a local macro
+table by concatenating the exported macro tables of the referenced modules, which must be
+declared within the adjacent `modules` field.
+
+With macros installed, the document can then invoke them using E-expressions, and the `point` and
+`line` invocations above produce results equivalent to:
```
almann, line 360: *Minor*: Editorial suggestion.

```suggestion
With macros installed, the document can then invoke them using E-Expressions, and the `point`/
`line` invocations above produce results equivalent to:
```

src/modules-by-example.adoc
```diff
@@ -319,3 +320,472 @@ Here again, Ion 1.1 enables a new technique: we can prepend new symbols to the e
...
+
+This topic is more interesting when more than one module is involved, so let's table this
+for now.
+
+
+=== Shared Modules
+
+Macros are most useful when they're shared across documents, and for that we use _shared modules_,
+a generalization of Ion 1.0's shared symbol tables. As discussed in
+<<_modules_from_the_outside>>, they export both a symbol table and a macro table.
```
almann, line 398: In Ion 1.1, a shared symbol table ***is*** a shared module, right?  Can we just say this explicitly, this text sort of says it, but I feel like this is something to really drill home.

src/modules-by-example.adoc
```diff
@@ -319,3 +320,472 @@ Here again, Ion 1.1 enables a new technique: we can prepend new symbols to the e
...
+      [a, b])
+    (*macro* poly  [(point first), (point... rest)]
+      [first, rest])
+  )
+}
+----
+
+The `module` field here is very similar to the `*module*` S-expression inside `$ion_encoding`.
+Here, no symbolic name is declared, since one will be assigned when the module is ``*import*``ed.
+No `*symbols*` clause is allowed, since those are expected to be in the legacy `symbols` field.
```
almann, line 430: Ugh, I know why this is here, but it is annoying that we have to deal with different parsing contexts and different rules for this stuff.

src/modules-by-example.adoc
```diff
@@ -319,3 +320,472 @@ Here again, Ion 1.1 enables a new technique: we can prepend new symbols to the e
...
+=== Using Shared Macros
+
+With a shared module at hand, we can import it and install its macros:
+
+[{nrm}]
+----
+*$ion_1_1*
+**$ion_encoding**::{
+  modules: [
+    (*import* g3d "com.example.graphics.3d" 1),     // Import it
```
almann, line 464: *Side Note*: I still dislike that if my shared module was called `g3d` that I would have to write something like:
```
(import g3d "g3d" 1)
```

src/modules-by-example.adoc
```diff
@@ -319,3 +320,472 @@ Here again, Ion 1.1 enables a new technique: we can prepend new symbols to the e
...
+
+In fact, we _must_ do so.  An E-expression with an un unqualified macro name is erroneous when
+the name is ambiguous, meaning that two installed modules map it to different macros.
+
+[{nrm}]
+----
+(:point 17 28) ⇒ **error**: ':point' is ambiguous, exported by 'geo' and 'g3d'.
+----
+
+Another thing to note in this directive is that the `**import** g3d` declaration includes a symbol
```
almann: *Minor*: Editorial suggestion.

```suggestion
Another thing to note in the directive used above is that the `**import** g3d` declaration includes a symbol
```

src/modules-by-example.adoc
```diff
@@ -319,3 +320,472 @@ Here again, Ion 1.1 enables a new technique: we can prepend new symbols to the e
...
+table name and version, but no max_id argument.  As with legacy imports, absence of max_id forces
```
almann: What is a "legacy import?" is that something that we define somewhere or do we mean  an import in an Ion 1.0 local symbol table?

src/modules-by-example.adoc
```diff
@@ -319,3 +320,472 @@ Here again, Ion 1.1 enables a new technique: we can prepend new symbols to the e
...
+the Ion implementation to acquire the symbol table entity with exactly the stated version.  While
+this is generally not best-practice for importing symbols, exact-match is a precondition for
+using the module in `install_macros`.  In other words, when a document is encoded using macros,
```
almann: *Minor*: Editorial suggestion.  I think we should be clearer on the requirement.

```suggestion
this is generally not best-practice for importing symbols, exact-match is a **requirement** for
using the module in `install_macros`.  In other words, when a document is encoded using macros,
```

src/modules-by-example.adoc
```diff
@@ -319,3 +320,472 @@ Here again, Ion 1.1 enables a new technique: we can prepend new symbols to the e
...
+**$ion_encoding**::{
+  modules: [(*import* chart "com.example.charts" 1)],
+  install_macros: [chart, geo]
+}
+  ⇒ **error**: no module named 'geo' is available for installation.
+----
+
+When the Ion implementation loads the 'chart' module, it will transitively load the geometry
+module as well, but the import of `com.example.geometry` by `com.example.charts` is
+_private_ to the importer.
```
almann: *Minor*: Editorial suggestion.

```suggestion
_not visible by name_ to the importer.
```

src/modules-by-example.adoc
```diff
@@ -319,3 +320,472 @@ Here again, Ion 1.1 enables a new technique: we can prepend new symbols to the e
...
+
+
+[{nrm}]
+----
+  modules: [
+    (*module* chart
+      (*import* geo "com.example.geometry" 1)
+      (*import* g3d "com.example.graphics.3d" 1)
+      (*alias* point2 ':geo:point')               // <1>
+      (*macro* scatterplot [(point2 \... points)]  // <2>
```
almann, line 653: I notice the clever use of `point2` to avoid shadowed  names.  Is this where `(':point2' ...)` could be used in the macro definition?

Also even with the shadowing, couldn't this lexical alias override the ambiguity?

I think this is worth showing as examples, because I am not sure I know what the rules are.

src/modules-by-example.adoc
```diff
@@ -319,3 +320,472 @@ Here again, Ion 1.1 enables a new technique: we can prepend new symbols to the e
...
+      (*export* point2 point3))
+  ],
+  install_macros: [ local, geo, g3d ]
+}
+(:point2 93 5)
+(:point3 0 12 33)
+----
+
+Exports can also be used to "pass through" selected macros from an imported module: `(*export*
+':g2d:line')` exports the name `line` from the enclosing module.
```
almann: I think this is different than:
```
(alias line ':g2d:line')
(export line)
```

And I think the difference is I don't get a local name, but is that right (and this question makes me think we should have text that illustrates if this is a difference or not).

src/modules-by-example.adoc
```diff
@@ -319,3 +320,472 @@ Here again, Ion 1.1 enables a new technique: we can prepend new symbols to the e
...
+  install_symbols: [local, local2]
+  install_macros:  [local, local2]
+}
+----
+
+
+=== Separate Installation
+
+The preceding example has some repetition between `install_symbols` and `install_macros`,
+illustrating that the symbol and macro tables are maintained independently.  There's no
```
almann: Maybe have an example where:

```
  install_symbols: [local, local2]
  install_macros:  [local2, local]
```

src/modules-by-example.adoc
```diff
@@ -319,3 +320,472 @@ Here again, Ion 1.1 enables a new technique: we can prepend new symbols to the e
...
+*$ion_1_1*
+**$ion_encoding**::{
+  modules: [
+    (*import* geo "com.example.geometry" 1),
+    (*import* g3d "com.example.graphics.3d" 1),
+    (*module* priority
+      (*import* g3d)
+      (*export* point tri))
+  ],
+  install_macros: [priority, geo, g3d]
```
almann, line 806: I think there needs to be text around the scoping of `import`, it surprised me that `(import g3d)` was needed in the locally defined module since you had imported it in the `modules` section.

almann (commented):
Looks pretty good--comments/questions below.

src/modules-by-example.adoc
toddjonker: This was an example from the later sections of the macro-by-example document.  I'll rephrase to not assume that dependency.

src/modules-by-example.adoc
toddjonker: I tried really hard to minimize any differences, this is the best I've been able to come up with.

I considered allowing a `symbols` clause inside the module, assuming there's nothing in the legacy `symbols` field (and thus no need to combine them), but then the overall SST isn't 1.0-meaningful.

src/modules-by-example.adoc
toddjonker: I've cut #213 to discuss.

src/modules-by-example.adoc
toddjonker: I mean the latter, so I'll just say that.

src/modules-by-example.adoc
toddjonker: Yes, you could write `(':point2' \... points)` in the signature, since that colon forces resolution among macro/alias declarations in this module.

But I kinda want to avoid getting into all the rules here.

PS I didn't think this was clever, its kind a the point of the example...

src/modules-by-example.adoc
toddjonker: You are correct. It has the same effect to the module's consumer, but there's no local name.  I'll add the example.

src/modules-by-example.adoc
toddjonker: TBH I've wavered on the necessity of this.  My last doc round eleminited the requirement, and the code in a module could see all macros exported from preceding `import` and `module` declarations.  But I think that will lead to more name collision and scoping confusion, and change-at-a-distance, with no way to "opt out" of that visibility.  In short, more qualified names and/or aliases just to workaround having things visible that you don't need or want.

When I wrote up this example I considered adding a new keyword for this mechanism, since this isn't really the same kind of "import".  My strawman would be perhaps `(use <module-name>+)`.   Would a different syntax make more sense?

Update: you could then `(use *)` to get the "everything is visible" state.

almann (approved):
🚢 LGTM

PR #209, opened 2023-04-26
Use more realistic names for shared symbol tables.

toddjonker:
### Description of changes:
Follow-up to discussion thread at https://github.com/amazon-ion/ion-docs/pull/205#discussion_r1176925931

----

### DIFF
--- a/src/modules-by-example.adoc
+++ b/src/modules-by-example.adoc
@@ -175,7 +175,7 @@ back to 1.0. First, turn the `local` module into the equivalent shared symbol ta
 [{nrm}]
 ----
 **$ion_shared_symbol_table**::{
-  name: "extracted",
+  name: "com.example.extracted",
   version: 1,
   symbols: ["s1", "s2"]
 }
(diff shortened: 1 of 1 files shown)


src/modules-by-example.adoc
```diff
@@ -175,7 +175,7 @@ back to 1.0. First, turn the `local` module into the equivalent shared symbol ta
 [{nrm}]
 ----
 **$ion_shared_symbol_table**::{
-  name: "extracted",
+  name: "com.example.extracted",
```
almann, line 178: TBH--I don't know why we like this awful Java style package naming--it is so verbose.

almann (approved):
LGTM 🚢

src/modules-by-example.adoc
toddjonker: I wouldn't say we "like" it, but IMO we should be encouraging use of name selection that's fairly globally-unique, because Ion offers no namespace mechanism.

Any sort of Ion Catalog Service at-scale would almost certainly have some enforced namespace and permissions mechanism to make sharing this data safe.   Just like open-source package managers have namespace conventions of various levels of expressiveness...

Would you prefer ARNs?  ;-)

PR #215, opened 2023-05-12
Ion 1.1 binary encoding specification

zslayton:
This PR adds an initial version the Ion 1.1 binary encoding specification. 

It does not include the binary encoding of E-expressions, which I will publish in a follow-on PR.

----

### DIFF
--- a/src/binary-encoding.adoc
+++ b/src/binary-encoding.adoc
@@ -326,7 +326,7 @@ The meanings of each opcode are organized loosely by their high and low nibbles.
 <|Delimited struct with `VarSym` field names start
 
 |`4`
-<|Reserved
+<|Variable length prefixed macro invocation
 
 |`5`
 <|Variable length integer
@@ -356,10 +356,10 @@ The meanings of each opcode are organized loosely by their high and low nibbles.
 <|Variable length struct with `VarSym` field names
 
 |`E`
-<|Reserved
+<|Variable length blob
 
 |`F`
-<|Variable length prefixed macro invocation
+<|Variable length clob
 
 |===
 
@@ -1047,24 +1047,329 @@ address that is decoded is biased by the number of addresses that can be encoded
 |16,641
 |===
 
-
 [[binary_data]]
 == Binary data
 
 [[blobs]]
 === Blobs
 
+Opcode `FE` indicates a blob of binary data.
+
+==== Example encoding of a 24-byte blob
+[source]
+----
+┌──── Opcode FE indicates a blob, VarUInt length follows
+│   ┌─── Length: VarUInt 24
+│   │
+FE 31 49 20 61 70 70 6c 61 75 64 20 79 6f 75 72 20 63 75 72 69 6f 73 69 74 79
+      └────────────────────────────────┬────────────────────────────────────┘
+                            24 bytes of binary data
+----
+
+
 [[clobs]]
 === Clobs
+
+Opcode `FF` indicates a clob--character data of an unspecified encoding.
+
+==== Example encoding of a 24-byte clob
+[source]
+----
+┌──── Opcode FF indicates a clob, VarUInt length follows
+│   ┌─── Length: VarUInt 24
+│   │
+FF 31 49 20 61 70 70 6c 61 75 64 20 79 6f 75 72 20 63 75 72 69 6f 73 69 74 79
+      └────────────────────────────────┬────────────────────────────────────┘
+                            24 bytes of binary data
+----
 
 [[containers]]
 == Containers
 
+Each of the container types (list, s-expression, and struct) has both a length-prefixed encoding and a delimited
+encoding.
+
+The length-prefixed encoding places more burden on the writer, but simplifies reading and enables skipping
+over uninteresting values in the data stream. In contrast, the delimited encoding is simpler and faster for
+writers, but requires the reader to visit each child value in turn to skip over the container.
+
 [[lists]]
 === Lists
+
+==== Length-prefixed encoding
+
+An opcode with a high nibble of `0xA_` indicates a length-prefixed list. The lower nibble of the
+opcode indicates how many bytes were used to encode the child values that the list contains.
+
+If the list's encoded byte-length is too large to be encoded in a nibble, writers may use the `0xFA` opcode
+to write a variable-length list. The `0xFA` opcode is followed by a
+<<varuint, `VarUInt`>> that indicates the list's byte length.
+
+===== Example length-prefixed encoding of an empty list (`[]`)
+[source]
+----
+┌──── An Opcode in the range 0xA0-0xAF indicates a list.
+│┌─── A low nibble of 0 indicates that the child values of this list took zero bytes to encode.
+A0
+----
+
+===== Example encoding of `[1, 2, 3]`
+[source]
+----
+┌──── An Opcode in the range 0xA0-0xAF indicates a list.
+│┌─── A low nibble of 0 indicates that the child values of this list took zero bytes to encode.
+A6 51 01 51 02 51 03
+   └─┬─┘ └─┬─┘ └─┬─┘
+     1     2     3
+----
+
+===== Example encoding of `["variable length list"]`
+[source]
+----
+┌──── Opcode 0xFA indicates a variable-length list. A VarUInt length follows.
+│  ┌───── Length: VarUInt 22
+│  │  ┌────── Opcode 0xF8 indicates a variable-length string. A VarUInt length follows.
+│  │  │  ┌─────── Length: VarUInt 20
+│  │  │  │   v  a  r  i  a  b  l  e     l  e  n  g  t  h     l  i  s  t
+FA 2d F8 29 76 61 72 69 61 62 6c 65 20 6c 65 6e 67 74 68 20 6c 69 73 74
+      └─────────────────────────────┬─────────────────────────────────┘
+                          Nested string element
+----
+
+==== Delimited encoding
+
+Opcode `0xF1` begins a delimited list, while opcode `0xF0` closes the most recently opened delimited container
+that has not yet been closed.
+
+===== Example delimited encoding of an empty list (`[]`)
+[source]
+----
+┌──── Opcode 0xF1 indicates a delimited list
+│  ┌─── Opcode 0xF0 indicates the end of the most recently opened container
+F1 F0
+----
+
+===== Example encoding of `[1, 2, 3]`
+[source]
+----
+┌──── Opcode 0xF1 indicates a delimited list
+│                    ┌─── Opcode 0xF0 indicates the end of
+│                    │    the most recently opened container
+F1 51 01 51 02 51 03 F0
+   └─┬─┘ └─┬─┘ └─┬─┘
+     1     2     3
+----
+
+===== Example encoding of `[1, [2], 3]`
+[source]
+----
+┌──── Opcode 0xF1 indicates a delimited list
+│        ┌─── Opcode 0xF1 begins a nested delimited list
+│        │        ┌─── Opcode 0xF0 closes the most recently
+│        │        │    opened delimited container: the nested list.
+│        │        │        ┌─── Opcode 0xF0 closes the most recently opened (and still open)
+│        │        │        │    delimited container: the outer list.
+│        │        │        │
+F1 51 01 F1 51 02 F0 51 03 F0
+   └─┬─┘    └─┬─┘    └─┬─┘
+     1        2        3
+----
 
 [[s_expressions]]
 === S-expressions
 
+==== Length-prefixed encoding
+
+An opcode with a high nibble of `0xB_` indicates a length-prefixed S-expression. The lower nibble of the
+opcode indicates how many bytes were used to encode the child values that the S-expression contains.
+
+If the S-expression's encoded byte-length is too large to be encoded in a nibble, writers may use
+the `0xFB` opcode to write a variable-length S-expression. The `0xFB` opcode is followed by a
+<<varuint, `VarUInt`>> that indicates the S-expression's byte length.
+
+===== Example length-prefixed encoding of an empty S-expression (`()`)
+[source]
+----
+┌──── An Opcode in the range 0xB0-0xBF indicates an S-expression.
+│┌─── A low nibble of 0 indicates that the child values of this S-expression took zero bytes to encode.
+B0
+----
+
+===== Example encoding of `(1 2 3)`
+[source]
+----
+┌──── An Opcode in the range 0xB0-0xBF indicates an S-expression.
+│┌─── A low nibble of 6 indicates that the child values of this S-expression took six bytes to encode.
+B6 51 01 51 02 51 03
+   └─┬─┘ └─┬─┘ └─┬─┘
+     1     2     3
+----
+
+===== Example encoding of `("variable length sexp")`
+[source]
+----
+┌──── Opcode 0xFB indicates a variable-length list. A VarUInt length follows.
+│  ┌───── Length: VarUInt 22
+│  │  ┌────── Opcode 0xF8 indicates a variable-length string. A VarUInt length follows.
+│  │  │  ┌─────── Length: VarUInt 20
+│  │  │  │   v  a  r  i  a  b  l  e     l  e  n  g  t  h     s  e  x  p
+FB 2d F8 29 76 61 72 69 61 62 6c 65 20 6c 65 6e 67 74 68 20 73 65 78 70
+      └─────────────────────────────┬─────────────────────────────────┘
+                          Nested string element
+----
+
+==== Delimited encoding
+
+Opcode `0xF2` begins a delimited S-expression, while opcode `0xF0` closes the most recently opened
+delimited container that has not yet been closed.
+
+===== Example delimited encoding of an empty S-expression (`()`)
+[source]
+----
+┌──── Opcode 0xF2 indicates a delimited S-expression
+│  ┌─── Opcode 0xF0 indicates the end of the most recently opened container
+F2 F0
+----
+
+===== Example encoding of `(1 2 3)`
+[source]
+----
+┌──── Opcode 0xF2 indicates a delimited S-expression
+│                    ┌─── Opcode 0xF0 indicates the end of
+│                    │    the most recently opened container
+F2 51 01 51 02 51 03 F0
+   └─┬─┘ └─┬─┘ └─┬─┘
+     1     2     3
+----
+
+===== Example encoding of `(1 (2) 3)`
+[source]
+----
+┌──── Opcode 0xF2 indicates a delimited S-expression
+│        ┌─── Opcode 0xF2 begins a nested delimited S-expression
+│        │        ┌─── Opcode 0xF0 closes the most recently
+│        │        │    opened delimited container: the nested S-expression.
+│        │        │        ┌─── Opcode 0xF0 closes the most recently opened (and still open)
+│        │        │        │    delimited container: the outer S-expression.
+│        │        │        │
+F2 51 01 F2 51 02 F0 51 03 F0
+   └─┬─┘    └─┬─┘    └─┬─┘
+     1        2        3
+----
+
 [[structs]]
 === Structs
+
+Structs have 3 available encodings:
+
+. Structs with symbol address field names
+. Structs with <<varsym, `VarSym`>> field names
+. Delimited structs with `VarSym` field names
+
+[[structs_with_symbol_address_field_names]]
+==== Structs with symbol address field names
+
+An opcode with a high nibble of `0xC_` indicates a struct with symbol address field names. The lower
+nibble of the opcode indicates how many bytes were used to encode all of its nested `(field name, value)`
+pairs.
+
+If the struct's encoded byte-length is too large to be encoded in a nibble, writers may use the `0xFC` opcode
+to write a variable-length struct with symbol address field names. The `0xFC` opcode is followed by a
+<<varuint, `VarUInt`>> that indicates the byte length.
+
+Each field in the struct is encoded as a <<varuint, `VarUInt`>> representing the address of the field name's
+text in the symbol table, followed by an opcode-prefixed value.
+
+===== Example encoding of an empty struct (`{}`)
+[source]
+----
+┌──── An opcode in the range 0xC0-0xCF indicates a struct with symbol address field names
+│┌─── A lower nibble of 0 indicates that the struct's fields took zero bytes to encode
+C0
+----
+
+===== Example encoding of `{$10: 1, $11: 2}`
+[source]
+----
+┌──── An opcode in the range 0xC0-0xCF indicates a struct with symbol address field names
+│  ┌─── Field name: VarUInt 10 ($10)
+│  │        ┌─── Field name: VarUInt 11 ($11)
+│  │        │
+C6 15 51 01 17 51 02
+      └─┬─┘    └─┬─┘
+        1        2
+----
+
+===== Example encoding of `{$10: "variable length struct"}`
+[source]
+----
+ ┌───────────── Opcode `FC` indicates a variable length struct with symbol address field names
+ │  ┌────────── Length: VarUInt 25
+ │  │  ┌─────── Field name: VarUInt 10 ($10)
+ │  │  │  ┌──── Opcode `F8` indicates a variable length string
+ │  │  │  │  ┌─ VarUInt: 22 the string is 22 bytes long
+ │  │  │  │  │  v  a  r  i  a  b  l  e     l  e  n  g  t  h     s  t  r  u  c  t
+FC 33 15 F8 2D 76 61 72 69 61 62 6c 65 20 6c 65 6e 67 74 68 20 73 74 72 75 63 74
+               └─────────────────────────────┬─────────────────────────────────┘
+                                        UTF-8 bytes
+----
+
+[[structs_with_varsym_field_names]]
+==== Structs with `VarSym` field names
+
+NOTE: This encoding is very similar to <<structs_with_symbol_address_field_names, structs with symbol address
+field names>>, but allows writers to choose between representing each field name as a symbol address
+(for example: `$10`) or inline UTF-8 bytes (for example: `"foo"`). This encoding is potentially less
+dense, but offers writers significant flexibility over whether and when field names are added to the
+symbol table.
+
+An opcode with a high nibble of `0xD_` indicates a struct with <<varsym, `VarSym`>> field names. The lower
+nibble of the opcode indicates how many bytes were used to encode all of its nested `(field name, value)`
+pairs.
+
+WARNING: Empty structs MUST be written using `0xC0`. `0xD0` is a reserved opcode.
+
+If the struct's encoded byte-length is too large to be encoded in a nibble, writers may use the `0xFD` opcode
+to write a variable-length struct with <<varsym, `VarSym`>> field names. The `0xFD` opcode is followed by a
+<<varuint, `VarUInt`>> that indicates the byte length.
+
+Each field in the struct is encoded as a  <<varsym, `VarSym`>> field name, followed by an
+opcode-prefixed value.
+
+===== Example encoding of `{"foo": 1, $11: 2}`
+[source]
+----
+┌─── Opcode with high nibble `D` indicates a struct with VarSym field names
+│┌── Length: 7
+││ ┌─ VarSym -3      ┌─ VarSym: 11 ($11)
+││ │   f  o  o       │
+D9 FD 66 6F 6F 51 01 17 91 02
+      └──┬───┘ └─┬─┘    └─┬─┘
+      3 UTF-8    1        2
+       bytes
+----
+
+// TODO: Demonstrate splicing macro values into the struct via VarSym escape code `0x00`.
+
+[[delimited_structs]]
+==== Delimited structs
+
+Opcode `0xF3` indicates the beginning of a delimited struct with `VarSym` field names,
+while opcode `0xF0` closes the most recently opened delimited container that has not yet been closed.
+
+NOTE: There is no delimited encoding for structs with symbol address field names.
+
+===== Example encoding of `{"foo": 1, $11: 2}`
+[source]
+----
+┌─── Opcode 0xF3 indicates the beginning of a delimited struct with `VarSym` field names.
+│
+│  ┌─ VarSym -3      ┌─ VarSym: 11 ($11)
+│  │                 │
+│  │                 │         ┌─ Opcode 0xF0 indicates the end of the most
+│  │   f  o  o       │         │  recently opened delimited container
+F3 FD 66 6F 6F 51 01 17 91 02 F0
+      └──┬───┘ └─┬─┘    └─┬─┘
+      3 UTF-8    1        2
+       bytes
+----
--- a/src/binary-encoding.adoc
+++ b/src/binary-encoding.adoc
@@ -159,12 +159,17 @@ expansion.
 
 A `VarSym` begins with a <<varint,`VarInt`>>; once this integer has been read, we can evaluate it to determine how to proceed. If the VarInt is:
 
-* *greater than zero*, it represents a symbol ID. The symbol’s associated text can be found in the local symbol table. No more bytes follow.
-* *less than zero*, its absolute value represents a number of UTF-8 bytes that follow the `VarInt`. These bytes represent the symbol’s text.
-* *exactly zero*, another byte follows that is an <<opcodes, opcode>>. The `VarSym` parser is not responsible for evaluating this opcode, only returning it—the caller will decide whether the opcode is legal in the current context. Example usages of the opcode include:
+* *greater than zero*, it represents a symbol ID. The symbol’s associated text can be found in the local symbol table.
+No more bytes follow.
+* *less than zero*, its absolute value represents a number of UTF-8 bytes that follow the `VarInt`. These bytes
+represent the symbol’s text.
+* *exactly zero*, another byte follows that is an <<opcodes, opcode>>. The `VarSym` parser is not responsible for
+evaluating this opcode, only returning it—the caller will decide whether the opcode is legal in the current context.
+Example usages of the opcode include:
 ** Representing SID `$0` as `0x70`
 ** Representing the empty string (`""`) as `0x80`
-** When used to encode a struct field name, the opcode can invoke a macro that will evaluate to a struct whose key/value pairs are spliced into the parent struct (TODO: Link)
+** When used to encode a struct field name, the opcode can invoke a macro that will evaluate to a struct whose key/value
+pairs are spliced into the parent struct (TODO: Link)
 ** In a delimited struct, terminating the sequence of (name, value) pairs with `0xAD`. (TODO: Link)
 
 ==== Example encoding of a `VarSym` with symbol ID `$10` ====
@@ -418,10 +423,12 @@ F4 05 50 fc
 Float values are encoded using the IEEE-754 specification, and can be serialized in four sizes:
 
 * 0 bits (0 bytes), representing the value 0e0 and indicated by opcode `0x5A`
-* 16 bits (2 bytes, link:https://en.wikipedia.org/wiki/Half-precision_floating-point_format[half precision]), indicated
-by opcode `0x5B`
-* 32 bits (4 bytes, link:https://en.wikipedia.org/wiki/Single-precision_floating-point_format[single precision]), indicated by opcode `0x5C`
-* 64 bits (8 bytes, link:https://en.wikipedia.org/wiki/Double-precision_floating-point_format[double precision]), indicated by opcode `0x5D`
+* 16 bits (2 bytes, link:https://en.wikipedia.org/wiki/Half-precision_floating-point_format[half precision]),
+indicated by opcode `0x5B`
+* 32 bits (4 bytes, link:https://en.wikipedia.org/wiki/Single-precision_floating-point_format[single precision]),
+indicated by opcode `0x5C`
+* 64 bits (8 bytes, link:https://en.wikipedia.org/wiki/Double-precision_floating-point_format[double precision]),
+indicated by opcode `0x5D`
 
 Note that in the Ion data model, float values are always 64 bits. However, if a value can be losslessly serialized
 in fewer than 64 bits, applications may choose to do so.
@@ -625,7 +632,9 @@ consume only enough bits to represent the fractional precision supported by the 
 
 ==== Opcode `0x73`: Hour+Minutes @ UTC or Unknown (4 bytes)
 
-NOTE: Each encoding for a precision greater than or equal to `Hour+Minutes` comes in two flavors: one that uses a single bit (`U`) to indicate UTC versus Unknown offset, and another that uses 7 bits (`o`) to encode the number of quarter-hours offset from `-14:00`.
+NOTE: Each encoding for a precision greater than or equal to `Hour+Minutes` comes in two flavors: one that uses a single
+bit (`U`) to indicate UTC versus Unknown offset, and another that uses 7 bits (`o`) to encode the number of quarter-hours
+offset from `-14:00`.
 
 [source]
 ----
@@ -661,9 +670,9 @@ NOTE: Each encoding for a precision greater than or equal to `Hour+Minutes` come
 ==== Opcode `0x77`: Milliseconds @ UTC or Unknown (6 bytes)
 [source]
 ----
-+=========+=========+=========+=========+=========+
-|YYYY:YYYM|MMMD:DDDD|HHHH:Hmmm|mmmU:ssss|ss--:----|
-+=========+=========+=========+=========+=========+
++=========+=========+=========+=========+=========+=========+
+|YYYY:YYYM|MMMD:DDDD|HHHH:Hmmm|mmmk:ssss|ssff:ffff|ffff:----|
++=========+=========+=========+=========+=========+=========+
 ----
 
 ==== Opcode `0x78`: Milliseconds @ Offset (7 bytes)
@@ -762,7 +771,8 @@ If the timestamp's length is greater than or equal to `8`, it has fractional sec
 `(coefficient, exponent)` pair, similar to a <<decimals, decimal>>. However, it is illegal for the fractional
 seconds value to be greater than or equal to `1.0` or less than `0.0`. For this reason, both the exponent and
 the coefficient are encoded using unsigned types. The included exponent `VarUInt` is implicitly negative, preventing
-the encoding of decimal numbers greater than `1.0`. The coefficient `FixedUInt` is unsigned to prevent the encoding of fractional seconds less than `0.0`. Note that validation is still required; namely:
+the encoding of decimal numbers greater than `1.0`. The coefficient `FixedUInt` is unsigned to prevent the encoding of
+fractional seconds less than `0.0`. Note that validation is still required; namely:
 
 * An exponent value of `0` is illegal, as that would result in a fractional seconds greater than `1.0` (a whole second).
 * If `coefficient * 10^-exponent > 1.0`, that `(coefficient, exponent)` pair is illegal.
--- a/src/binary-encoding.adoc
+++ b/src/binary-encoding.adoc
@@ -757,11 +757,19 @@ consume only enough bits to represent the fractional precision supported by the 
 | Unused
 |===
 
+NOTE: Timestamps are encoded in Little Endian byte order. In the diagrams below, the bytes are read from left to right
+(least significant to most significant), while the bits within each byte should be read from right to left. (Again
+least significant to most significant.) +
+While this encoding may complicate the diagrams somewhat, it guarantees that the timestamp's subfields (`year`, `month`,
+etc.) occupy the same bit indexes regardless of how many bytes there are overall. (The last subfield,
+`fractional_seconds`, always begins at the same bit index when present, but can vary in length according to the
+precision.)
+
 ==== Opcode `0x70`: Year (1 byte)
 [source]
 ----
 +=========+
-|YYYY:YYY-|
+|-YYY:YYYY|
 +=========+
 ----
 
@@ -769,7 +777,7 @@ consume only enough bits to represent the fractional precision supported by the 
 [source]
 ----
 +=========+=========+
-|YYYY:YYYM|MMM-:----|
+|MYYY:YYYY|----:-MMM|
 +=========+=========+
 ----
 
@@ -777,7 +785,7 @@ consume only enough bits to represent the fractional precision supported by the 
 [source]
 ----
 +=========+=========+
-|YYYY:YYYM|MMMD:DDDD|
+|MYYY:YYYY|DDDD:DMMM|
 +=========+=========+
 ----
 
@@ -790,15 +798,16 @@ offset from `-14:00`.
 [source]
 ----
 +=========+=========+=========+=========+
-|YYYY:YYYM|MMMD:DDDD|HHHH:Hmmm|mmmU:----|
+|MYYY:YYYY|DDDD:DMMM|mmmH:HHHH|----:Ummm|
 +=========+=========+=========+=========+
 ----
 
 ==== Opcode `0x74`: Hour+Minutes @ Offset (5 bytes)
 [source]
 ----
+
 +=========+=========+=========+=========+=========+
-|YYYY:YYYM|MMMD:DDDD|HHHH:Hmmm|mmmo:oooo|oo--:----|
+|MYYY:YYYY|DDDD:DMMM|mmmH:HHHH|oooo:ommm|----:--oo|
 +=========+=========+=========+=========+=========+
 ----
 
@@ -806,7 +815,7 @@ offset from `-14:00`.
 [source]
 ----
 +=========+=========+=========+=========+=========+
-|YYYY:YYYM|MMMD:DDDD|HHHH:Hmmm|mmmU:ssss|ss--:----|
+|MYYY:YYYY|DDDD:DMMM|mmmH:HHHH|ssss:Ummm|----:--ss|
 +=========+=========+=========+=========+=========+
 ----
 
@@ -814,7 +823,7 @@ offset from `-14:00`.
 [source]
 ----
 +=========+=========+=========+=========+=========+
-|YYYY:YYYM|MMMD:DDDD|HHHH:Hmmm|mmmo:oooo|oo--:----|
+|MYYY:YYYY|DDDD:DMMM|mmmH:HHHH|oooo:ommm|ssss:ssoo|
 +=========+=========+=========+=========+=========+
 ----
 
@@ -822,7 +831,7 @@ offset from `-14:00`.
 [source]
 ----
 +=========+=========+=========+=========+=========+=========+
-|YYYY:YYYM|MMMD:DDDD|HHHH:Hmmm|mmmk:ssss|ssff:ffff|ffff:----|
+|MYYY:YYYY|DDDD:DMMM|mmmH:HHHH|ssss:Ummm|ffff:ffss|----:ffff|
 +=========+=========+=========+=========+=========+=========+
 ----
 
@@ -830,7 +839,7 @@ offset from `-14:00`.
 [source]
 ----
 +=========+=========+=========+=========+=========+=========+=========+
-|YYYY:YYYM|MMMD:DDDD|HHHH:Hmmm|mmmo:oooo|ooss:ssss|ffff:ffff|ff--:----|
+|MYYY:YYYY|DDDD:DMMM|mmmH:HHHH|oooo:ommm|ssss:ssoo|ffff:ffff|----:--ff|
 +=========+=========+=========+=========+=========+=========+=========+
 ----
 
@@ -838,7 +847,7 @@ offset from `-14:00`.
 [source]
 ----
 +=========+=========+=========+=========+=========+=========+=========+
-|YYYY:YYYM|MMMD:DDDD|HHHH:Hmmm|mmmk:ssss|ssff:ffff|ffff:ffff|ffff:ff--|
+|MYYY:YYYY|DDDD:DMMM|mmmH:HHHH|ssss:Ummm|ffff:ffss|ffff:ffff|--ff:ffff|
 +=========+=========+=========+=========+=========+=========+=========+
 ----
 
@@ -846,7 +855,7 @@ offset from `-14:00`.
 [source]
 ----
 +=========+=========+=========+=========+=========+=========+=========+=========+
-|YYYY:YYYM|MMMD:DDDD|HHHH:Hmmm|mmmo:oooo|ooss:ssss|ffff:ffff|ffff:ffff|ffff:----|
+|MYYY:YYYY|DDDD:DMMM|mmmH:HHHH|oooo:ommm|ssss:ssoo|ffff:ffff|ffff:ffff|----:ffff|
 +=========+=========+=========+=========+=========+=========+=========+=========+
 ----
 
@@ -854,16 +863,16 @@ offset from `-14:00`.
 [source]
 ----
 +=========+=========+=========+=========+=========+=========+=========+=========+
-|YYYY:YYYM|MMMD:DDDD|HHHH:Hmmm|mmmo:oooo|ooss:ssss|ffff:ffff|ffff:ffff|ffff:----|
+|MYYY:YYYY|DDDD:DMMM|mmmH:HHHH|ssss:Ummm|ffff:ffss|ffff:ffff|ffff:ffff|ffff:ffff|
 +=========+=========+=========+=========+=========+=========+=========+=========+
 ----
 
-==== Opcode `0x7B`: Nanoseconds @ Offset (8 bytes)
+==== Opcode `0x7B`: Nanoseconds @ Offset (9 bytes)
 [source]
 ----
-+=========+=========+=========+=========+=========+=========+=========+=========+
-|YYYY:YYYM|MMMD:DDDD|HHHH:Hmmm|mmmo:oooo|ooss:ssss|ffff:ffff|ffff:ffff|ffff:----|
-+=========+=========+=========+=========+=========+=========+=========+=========+
++=========+=========+=========+=========+=========+=========+=========+=========+=========+
+|MYYY:YYYY|DDDD:DMMM|mmmH:HHHH|oooo:ommm|ssss:ssoo|ffff:ffff|ffff:ffff|ffff:ffff|--ff:ffff|
++=========+=========+=========+=========+=========+=========+=========+=========+=========+
 ----
 
 WARNING: Opcodes `0x7D`, `0x7E`, and `7F` are illegal; they are reserved for future use.
@@ -920,15 +929,15 @@ Unlike the short-form encoding, the long-form encoding reserves:
 
 If the timestamp's length is greater than or equal to `8`, it has fractional seconds that are encoded as a
 `(coefficient, exponent)` pair, similar to a <<decimals, decimal>>. However, it is illegal for the fractional
-seconds value to be greater than or equal to `1.0` or less than `0.0`. For this reason, both the exponent and
-the coefficient are encoded using unsigned types. The included exponent `VarUInt` is implicitly negative, preventing
-the encoding of decimal numbers greater than `1.0`. The coefficient `FixedUInt` is unsigned to prevent the encoding of
-fractional seconds less than `0.0`. Note that validation is still required; namely:
+seconds value to be greater than or equal to `1.0` or less than `0.0`. For this reason, both the coefficient and
+the exponent are encoded using unsigned types. The included coefficient `VarUInt` is unsigned to prevent the encoding of
+fractional seconds less than `0.0`. The exponent `FixedUInt` is implicitly negative, discouraging the encoding of
+decimal numbers greater than `1.0`. Note that validation is still required; namely:
 
 * An exponent value of `0` is illegal, as that would result in a fractional seconds greater than `1.0` (a whole second).
 * If `coefficient * 10^-exponent > 1.0`, that `(coefficient, exponent)` pair is illegal.
 
-If the timestamp's length is `3`, the least significant bit in the final byte (`h`) is a flag
+If the timestamp's length is `3`, the most significant bit in the final byte (`h`) is a flag
 that indicates month (`0`) or day (`1`) precision.
 
 ==== Opcode `0xF7`: Long-form timestamp
@@ -936,7 +945,7 @@ that indicates month (`0`) or day (`1`) precision.
 ----
      1         2         3         4         5         6         7       8         n
 +=========+=========+=========+=========+=========+=========+=========+=======+   +=========+
-|YYYY:YYYY|YYYY:YYMM|MMDD:DDDh|HHHH:mmmm|mmoo:oooo|oooo:ooss|ssss|----|VarUInt|...|FixedUInt|...
+|YYYY:YYYY|MMYY:YYYY|hDDD:DDMM|mmmm:HHHH|oooo:oomm|ssoo:oooo|----|ssss|VarUInt|...|FixedUInt|...
 +=========+=========+=========+=========+=========+=========+=========+=======+   +=========+
 ----
 
--- a/src/binary-encoding.adoc
+++ b/src/binary-encoding.adoc
@@ -73,7 +73,7 @@ as two's complement.
 .Figure {counter:figure}: `_VarInt_` encoding of `_14_`
 [source,%unbreakable]
 ----
-              ┌──── Lowest bit is 1 (`end`), indicating
+              ┌──── Lowest bit is 1 (end), indicating
               │     this is the only byte.
 0 0 0 1 1 1 0 1
 └─────┬─────┘
@@ -83,7 +83,7 @@ as two's complement.
 .Figure {counter:figure}: `_VarInt_` encoding of `_-14_`
 [source,%unbreakable]
 ----
-              ┌──── Lowest bit is 1 (`end`), indicating
+              ┌──── Lowest bit is 1 (end), indicating
               │     this is the only byte.
 1 1 1 0 0 1 0 1
 └─────┬─────┘
@@ -686,7 +686,7 @@ using UTC time. This required applications to perform conversion logic when tran
 *In Ion 1.1, all binary timestamp fields are encoded in local time.*
 
 [[short_form_timestamp]]
-=== Short-form timestamp
+=== Short-form Timestamp
 
 If an opcode has a high nibble of `0x7_`, it represents a short-form timestamp. This encoding focuses on making the
 most common timestamp precisions and ranges the most compact; less common precisions can still be expressed via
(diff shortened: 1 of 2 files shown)


src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
+= Binary Encoding
+:toc:
+
+[[encoding_primitives]]
+== Encoding primitives
```
almann:
```suggestion
== Encoding Primitives
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+the ``VarUInt``'s least significant bits.
+
+The least significant bits in the `VarUInt` indicate the number of bytes that were used to encode the integer. The number
+of bytes can be found by counting the number of trailing zeros in the first byte(s) and adding one to it. All bits that
+are more significant than the first `1` represent the magnitude of the `VarUInt`.
+
+==== Example encoding of VarUInt 14 ====
+[source]
+----
+              ┌──── Lowest bit is 1 (`end`), indicating
```
almann:
```suggestion
              ┌──── Lowest bit is 1 (end), indicating
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+=== `VarUInt`
+
+A variable-length unsigned integer.
+
+The bytes of a ``VarUInt``s are written in
+link:https://en.wikipedia.org/wiki/Endianness:[Little Endian byte order]. This means that the first byte will contain
+the ``VarUInt``'s least significant bits.
+
+The least significant bits in the `VarUInt` indicate the number of bytes that were used to encode the integer. The number
+of bytes can be found by counting the number of trailing zeros in the first byte(s) and adding one to it. All bits that
```
toddjonker: I think "trailing" and endian-ness evoke tension, so how about:

```suggestion
of bytes can be found by counting the number of least-significant zeros in the first byte(s) and adding one to it. All bits that
```

(and ditto below)

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+as two's complement.
+
+==== Example encoding of VarInt 14 ====
+[source]
+----
+              ┌──── Lowest bit is 1 (`end`), indicating
+              │     this is the only byte.
+0 0 0 0 1 1 1 1
+└─────┬─────┘
+ 2's comp. 14
```
toddjonker: I think this is incorrect bits for 14

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
+= Binary Encoding
+:toc:
+
+[[encoding_primitives]]
+== Encoding primitives
+
+[[varuint]]
+=== `VarUInt`
```
toddjonker: I think there will be confusion (and probably bugs) in spec and code if we reuse these terms across Ion versions.

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+└────┬────┘      └──────┬──────┘
+lowest 6 bits    highest 8 bits
+of the 2's       of the 2's
+comp. integer    comp. integer
+----
+
+[[fixeduint]]
+=== `FixedUInt`
+
+A fixed-width, little-endian, unsigned integer whose length is inferred from the context in which it appears.
```
toddjonker, line 117: Can these be padded?

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+
+[[multi_byte_macro_invocations]]
+=== Multibyte macro invocations
+
+While invocations of macro addresses in the range `[0, 63]` can be encoded in a single byte using
+<<single_byte_macro_invocations, single byte macro invocations>>, many applications will benefit from defining more than
+64 macros.
+
+If the high nibble of the opcode is `0x4_`, then the low nibble represents the four least significant bits of the macro
+address. A <<varuint, `VarUInt`>> follows that contains the remaining, more significant bits.
```
toddjonker: This feels backwards in endianness, when compared with the way bits carry over in VarUInt.

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+   │       └── 4 least significant    significant bits of the macro
+opcode high    bits of the macro      address
+nibble 4       address
+
+Magnitude bits: 01000101_101001_0000
+Decoded value : 71,312
+Biased value  : 71,376
+----
+
+NOTE: From this point on in the document, example encodings are given in hexidecimal notation.
```
toddjonker:
```suggestion
NOTE: From this point on in the document, example encodings are given in hexadecimal notation.
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+| `*`
+|Single-byte macro invocations
+
+|`0x4_`
+| `*`
+|Multibyte macro invocations
+
+.4+|`0x5_`
+| `0`-`8`
+|Integers
```
toddjonker:
```suggestion
|Fixed-width integers up to 8 bytes wide
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+   └─┬─┘
+FixedInt -944
+----
+
+==== Example variable-length encoding of `-944`
+[source]
+----
+┌──── Opcode F5 indicates a variable-length integer, VarUInt length follows
+│   ┌─── VarUInt 2; a 2-byte FixedInt follows
+│   │    that two bytes follow.
```
toddjonker:
```suggestion
│   ┌─── VarUInt 2; a 2-byte FixedInt follows.
│   │
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+
+[[decimals]]
+=== Decimals
+
+If an opcode has a high nibble of `0x6_`, it represents a decimal. Low nibble values `0x_E` and below indicate
+the number of trailing bytes used to encode the decimal.
+
+The body of the decimal is encoded as a <<varint, `VarInt`>> representing its exponent, followed by a `FixedInt`
+representing its coefficient. The width of the coefficient is the total length of the decimal encoding minus the length
+of the exponent. It is possible for the coefficient to have a width of zero, indicating a coefficient of `0`.
```
toddjonker: Wow, this is silly!  It would make more sense for the _exponent_ to be optional, to shave a byte from decimal-encoded integers like `17.` aka `17d0`.  Eliding the coefficient shaves bits for the far less useful set of values `0.000...`.

(Or has my brain fully decaffeinated?)

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
++=========+
+|YYYY:YYY-|
++=========+
+----
+
+==== Opcode `0x71`: Month (2 bytes)
+[source]
+----
++=========+=========+
+|YYYY:YYYM|MMM-:----|
```
toddjonker: I observe this is big-endian layout...

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
++=========+=========+=========+=========+=========+
+|YYYY:YYYM|MMMD:DDDD|HHHH:Hmmm|mmmo:oooo|oo--:----|
++=========+=========+=========+=========+=========+
+----
+
+==== Opcode `0x77`: Milliseconds @ UTC or Unknown (6 bytes)
+[source]
+----
++=========+=========+=========+=========+=========+=========+
+|YYYY:YYYM|MMMD:DDDD|HHHH:Hmmm|mmmk:ssss|ssff:ffff|ffff:----|
```
toddjonker:
```suggestion
|YYYY:YYYM|MMMD:DDDD|HHHH:Hmmm|mmmU:ssss|ssff:ffff|ffff:----|
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
++=========+=========+=========+=========+=========+
+|YYYY:YYYM|MMMD:DDDD|HHHH:Hmmm|mmmU:ssss|ss--:----|
++=========+=========+=========+=========+=========+
+----
+
+==== Opcode `0x76`: Seconds @ Offset (5 bytes)
+[source]
+----
++=========+=========+=========+=========+=========+
+|YYYY:YYYM|MMMD:DDDD|HHHH:Hmmm|mmmo:oooo|oo--:----|
```
toddjonker: Missing the seconds subfield.

```suggestion
|YYYY:YYYM|MMMD:DDDD|HHHH:Hmmm|mmmo:oooo|ooss:ssss|
```
I think that's right?

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
++=========+=========+=========+=========+=========+=========+=========+
+|YYYY:YYYM|MMMD:DDDD|HHHH:Hmmm|mmmo:oooo|ooss:ssss|ffff:ffff|ff--:----|
++=========+=========+=========+=========+=========+=========+=========+
+----
+
+==== Opcode `0x79`: Microseconds @ UTC or Unknown (7 bytes)
+[source]
+----
++=========+=========+=========+=========+=========+=========+=========+
+|YYYY:YYYM|MMMD:DDDD|HHHH:Hmmm|mmmk:ssss|ssff:ffff|ffff:ffff|ffff:ff--|
```
toddjonker:
```suggestion
|YYYY:YYYM|MMMD:DDDD|HHHH:Hmmm|mmmU:ssss|ssff:ffff|ffff:ffff|ffff:ff--|
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+|YYYY:YYYM|MMMD:DDDD|HHHH:Hmmm|mmmo:oooo|ooss:ssss|ffff:ffff|ffff:ffff|ffff:----|
++=========+=========+=========+=========+=========+=========+=========+=========+
+----
+
+==== Opcode `0x7B`: Nanoseconds @ UTC or Unknown (8 bytes)
+[source]
+----
++=========+=========+=========+=========+=========+=========+=========+=========+
+|YYYY:YYYM|MMMD:DDDD|HHHH:Hmmm|mmmo:oooo|ooss:ssss|ffff:ffff|ffff:ffff|ffff:----|
++=========+=========+=========+=========+=========+=========+=========+=========+
```
toddjonker: I think this one needs `o`s replaced with `U` and more `f`s?

It's the same `f`s as microseconds, and same in the next entry, it's late and I can't math to figure out what's wrong.

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+----
+
+==== Opcode `0x7B`: Nanoseconds @ Offset (8 bytes)
+[source]
+----
++=========+=========+=========+=========+=========+=========+=========+=========+
+|YYYY:YYYM|MMMD:DDDD|HHHH:Hmmm|mmmo:oooo|ooss:ssss|ffff:ffff|ffff:ffff|ffff:----|
```
toddjonker: Also suspicious.

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+
+If the timestamp's length is `3`, the least significant bit in the final byte (`h`) is a flag
+that indicates month (`0`) or day (`1`) precision.
+
+==== Opcode `0xF7`: Long-form timestamp
+[source]
+----
+     1         2         3         4         5         6         7       8         n
++=========+=========+=========+=========+=========+=========+=========+=======+   +=========+
+|YYYY:YYYY|YYYY:YYMM|MMDD:DDDh|HHHH:mmmm|mmoo:oooo|oooo:ooss|ssss|----|VarUInt|...|FixedUInt|...
```
toddjonker:
```suggestion
|YYYY:YYYY|YYYY:YYMM|MMDD:DDDH|HHHH:mmmm|mmoo:oooo|oooo:ooss|ssss|----|VarUInt|...|FixedUInt|...
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+│  ┌─── Length: VarUInt 24
+│  │   v  a  r  i  a  b  l  e     l  e  n  g  t  h     e  n  c  o  d  i  n  g
+F8 31 76 61 72 69 61 62 6c 65 20 6c 65 6e 67 74 68 20 65 6e 63 6f 64 69 6e 67
+      └────────────────────────────────┬────────────────────────────────────┘
+                                  UTF-8 bytes
+----
+
+[[symbols_with_inline_text]]
+=== Symbols with inline text
+
```
toddjonker: Missing some preface text as with strings.

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+|0 to 255
+|0
+
+|`0xE2`
+|256 to 16,640
+|256
+
+|`0xE3`
+|16,641 to infinity
+|16,641
```
toddjonker: Doesn't two bytes give us 64k addresses, not 16k?

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+
+Each of the container types (list, s-expression, and struct) has both a length-prefixed encoding and a delimited
+encoding.
+
+The length-prefixed encoding places more burden on the writer, but simplifies reading and enables skipping
+over uninteresting values in the data stream. In contrast, the delimited encoding is simpler and faster for
+writers, but requires the reader to visit each child value in turn to skip over the container.
+
+[[lists]]
+=== Lists
```
toddjonker, line 1323: About here I noticed that nulls weren't always being covered in the section for each type.

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+field names>>, but allows writers to choose between representing each field name as a symbol address
+(for example: `$10`) or as inline UTF-8 bytes (for example: `"foo"`). This encoding is potentially less
+dense, but offers writers significant flexibility over whether and when field names are added to the
+symbol table.
+
+An opcode with a high nibble of `0xD_` indicates a struct with <<varsym, `VarSym`>> field names. The lower
+nibble of the opcode indicates how many bytes were used to encode all of its nested `(field name, value)`
+pairs.
+
+WARNING: Empty structs MUST be written using `0xC0`. `0xD0` is a reserved opcode.
```
toddjonker: Surprising, since `[]` and `()` can be written delimited.  Why?

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+Structs have 3 available encodings:
+
+. <<structs_with_symbol_address_field_names, Structs with symbol address field names>>
+. <<structs_with_varsym_field_names, Structs with `VarSym` field names>>
+. <<delimited_structs, Delimited structs with `VarSym` field names>>
+
+[[structs_with_symbol_address_field_names]]
+==== Structs with symbol address field names
+
+An opcode with a high nibble of `0xC_` indicates a struct with symbol address field names. The lower
```
toddjonker: Are we doing anything with the `0xC1` opcode, which would be impossible length?

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+│┌── Length: 7
+││ ┌─ VarSym -3      ┌─ VarSym: 11 ($11)
+││ │   f  o  o       │
+D9 FD 66 6F 6F 51 01 17 91 02
+      └──┬───┘ └─┬─┘    └─┬─┘
+      3 UTF-8    1        2
+       bytes
+----
+
+// TODO: Demonstrate splicing macro values into the struct via VarSym escape code `0x00`.
```
toddjonker: I was going to comment on this gap, maybe uncomment?

```suggestion
TODO: Demonstrate splicing macro values into the struct via VarSym escape code `0x00`.
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+
+[[delimited_structs]]
+==== Delimited structs
+
+Opcode `0xF3` indicates the beginning of a delimited struct with `VarSym` field names,
+while opcode `0xF0` closes the most recently opened delimited container that has not yet been closed.
+
+NOTE: There is no delimited encoding for structs with symbol address field names.
```
toddjonker:
```suggestion
NOTE: There is no delimited encoding for structs that exclusively use symbol address field names.
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+It is illegal for an annotations sequence to appear before any of the following:
+
+* The end of the stream
+* A <<nops, `NOP`>>
+// TODO: Links
+* An E-expression (that is: a macro invocation). To add annotations to a macro invocation, see the
+`annotate` macro.
+
+If an annotations sequence appears before one or more additional annotations sequences, the sequences
+are concatenated.
```
toddjonker: Repeated paragraph.

```suggestion
```

toddjonker (commented):
Overall, very good!   My main concern with the content is that at this point our little-endian-ness feels halfhearted.  I've called out a few spots where it was most glaring.

src/binary-encoding.adoc
zslayton: I'm open to renaming them. What would you suggest?

src/binary-encoding.adoc
zslayton: As written, yes. I'm open to forbidding padded representations if you want to reserve leading zero bytes for something in the future.

src/binary-encoding.adoc
zslayton: I'm not sure I follow; could you describe the arrangement you expected?

src/binary-encoding.adoc
zslayton: > Wow, this is silly! It would make more sense for the exponent to be optional, to shave a byte from decimal-encoded integers like 17. aka 17d0. Eliding the coefficient shaves bits for the far less useful set of values 0.000....

Wow, that's a great point! I ported over the 1.0 representation without too much thought, but your approach sounds much more beneficial. I'll take a pass at writing this up.

src/binary-encoding.adoc
zslayton: This is an interesting observation. Our timestamp encoding was following IEEE-754's precendent of "there are N bytes, and these bit ranges within those bytes hold significance X, Y, and Z." However, now that I think about it I would need to have the least significant time fields in the most significant bits to ensure that the field offsets for more significant time fields are always the same. I'll rework this section.

src/binary-encoding.adoc
zslayton: You ever stare at `VarUInt`s so long that all bytes become 7 bits? 😵‍💫 Good catch.

src/binary-encoding.adoc
zslayton: I considered having an entry for `null.TYPE` in each section, but figured its content would be redundant with the dedicated _Typed nulls_ section. Now that you've read the spec all the way through, would you rather have both?

src/binary-encoding.adoc
zslayton: The opcode overview table at the top lists it as `Reserved` for now.

src/binary-encoding.adoc
zslayton: Oops, I worded this poorly. I meant to convey that you can't use `0xD0` for an empty length-prefixed struct, because we already have `0xC0`. Since there are no fields, we can use the same opcode for both the symbol address and `VarSym` field name struct encodings. As phrased though, I accidentally forbade empty delimited struct. I'll tweak that.

src/binary-encoding.adoc
zslayton: I've been using the phrase "structs with symbol address field names" to refer to the `VarUInt`-centric struct encoding. I'll try to clarify this with an extra clause.

src/binary-encoding.adoc
zslayton: I changed this to:
> Integers up to 8 bytes wide

because "fixed width integer" is potentially overloaded with `int8`, `int16`, etc.

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+
+The bytes of a ``VarUInt``s are written in
+link:https://en.wikipedia.org/wiki/Endianness:[Little Endian byte order]. This means that the first byte will contain
+the ``VarUInt``'s least significant bits.
+
+The least significant bits in the `VarUInt` indicate the number of bytes that were used to encode the integer. The number
+of bytes can be found by counting the number of trailing zeros in the first byte(s) and adding one to it. All bits that
+are more significant than the first `1` represent the magnitude of the `VarUInt`.
+
+==== Example encoding of VarUInt 14 ====
```
almann:
```suggestion
==== Example encoding of `VarUInt` for `14` ====
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+[source]
+----
+              ┌──── Lowest bit is 1 (`end`), indicating
+              │     this is the only byte.
+0 0 0 1 1 1 0 1
+└─────┬─────┘
+unsigned int 14
+----
+
+==== Example encoding of VarUInt 729 ====
```
almann:
```suggestion
==== Example encoding of `VarUInt` for `729` ====
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+             │     integer is two bytes wide.
+            ┌┴┐
+0 1 1 0 0 1 1 0  0 0 0 0 1 0 1 1
+└────┬────┘      └──────┬──────┘
+lowest 6 bits    highest 8 bits
+of the unsigned  of the unsigned
+integer          integer
+----
+
+==== Example encoding of VarUInt 21,043 ====
```
almann:
```suggestion
==== Example encoding of `VarUInt` for `21,043` ====
```

src/binary-encoding.adoc
almann: I don't have great suggestions, but let me suggest a few:

* `FlexInt`/`FlexUInt` - distinct spelling, only one character more.  My favorite of the bunch
* `ElasticInt`/`ElasticUInt` - a bit long in my opinion.
* `VaryInt`/`VaryUInt` - a bit close to `Var` for my tastes but _technically_ not the same if `Var` stands for *variable*.

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+From an encoding perspective, ``VarInt``s are structurally similar to a `VarUInt` (<<varuint, described above>>). Both
+encode their bytes using Little Endian byte order, and both use the count of trailing zero bits to indicate how many
+bytes were used to encode the integer. They differ in the _interpretation_ of their bits; while a ``VarUInt``'s bits
+are unsigned, a ``VarInt``'s bits are encoded using
+link:https://en.wikipedia.org/wiki/Two%27s_complement[two's complement notation].
+
+TIP: An implementation could choose to read a `VarInt` by instead reading a `VarUInt` and then reinterpreting its bits
+as two's complement.
+
+==== Example encoding of VarInt 14 ====
```
almann:
```suggestion
==== Example encoding of `VarInt` for `14` ====
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+[source]
+----
+              ┌──── Lowest bit is 1 (`end`), indicating
+              │     this is the only byte.
+0 0 0 0 1 1 1 1
+└─────┬─────┘
+ 2's comp. 14
+----
+
+==== Example encoding of VarInt -14 ====
```
almann:
```suggestion
==== Example encoding of `VarInt` for `-14` ====
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+[source]
+----
+              ┌──── Lowest bit is 1 (`end`), indicating
+              │     this is the only byte.
+1 1 1 0 0 1 0 1
+└─────┬─────┘
+ 2's comp. -14
+----
+
+==== Example encoding of VarInt 729 ====
```
almann:
```suggestion
==== Example encoding of `VarInt` for `729` ====
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+             │     integer is two bytes wide.
+            ┌┴┐
+0 1 1 0 0 1 1 0  0 0 0 0 1 0 1 1
+└────┬────┘      └──────┬──────┘
+lowest 6 bits    highest 8 bits
+of the 2's       of the 2's
+comp. integer    comp. integer
+----
+
+==== Example encoding of VarInt -729 ====
```
almann:
```suggestion
==== Example encoding of `VarInt` for `-729` ====
```

src/binary-encoding.adoc
almann: There are no restrictions for any of our integers in Ion 1.0 and we have implementations that exploit this in varying capacity, I would not restrict it in 1.1 for any of our primitive integer encodings.

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+of the 2's       of the 2's
+comp. integer    comp. integer
+----
+
+[[fixeduint]]
+=== `FixedUInt`
+
+A fixed-width, little-endian, unsigned integer whose length is inferred from the context in which it appears.
+
+==== Example encoding of FixedUInt 3,954,261 ====
```
almann:
```suggestion
==== Example encoding of `FixedUInt` for `3,954,261` ====
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+integer          integer          integer
+----
+
+[[fixedint]]
+=== `FixedInt`
+
+A fixed-width, little-endian, signed integer whose length is known from the context in which it appears. Its bytes
+are interpreted as two's complement.
+
+==== Example encoding of FixedInt -3,954,261 ====
```
almann:
```suggestion
==== Example encoding of `FixedInt` for `-3,954,261` ====
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+└──────┬──────┘  └──────┬──────┘  └──────┬──────┘
+lowest 8 bits    next 8 bits of   highest 8 bits
+of the 2's       the 2's comp.    of the 2's comp.
+comp. integer    integer          integer
+----
+
+[[varsym]]
+=== `VarSym`
+
+A variable-length symbol whose UTF-8 bytes can be inline, found in the symbol table, or derived from a macro
```
almann: Should we be using *symbol token* to describe these things?

```suggestion
A variable-length symbol token whose UTF-8 bytes can be inline, found in the symbol table, or derived from a macro
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+
+The meanings of each opcode are organized loosely by their high and low nibbles.
+
+[cols="^.^1a,^.^1a,3a"]
+|===
+|High nibble | Low nibble | Meaning
+
+|`0x0_` to `0x3_`
+| `*`
+|Single-byte macro invocations
```
almann: single-byte and multi-byte is a little strange here.  Maybe something like:

> Macro invocation where the macro ID is inline with the opcode.

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+
+|`0x4_`
+| `*`
+|Multibyte macro invocations
```
almann: Similarly maybe something like:

> Macro invocation where the macro ID follows as a `VarUInt`.

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+=== Overview table
+
+The meanings of each opcode are organized loosely by their high and low nibbles.
+
+[cols="^.^1a,^.^1a,3a"]
+|===
+|High nibble | Low nibble | Meaning
+
+|`0x0_` to `0x3_`
+| `*`
```
almann: Does `*` just mean all of these are used for the encoding of that type in some way, or does it have more meaning than that?

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+
+.9+|`0xE_`
+|`0`
+|Ion version marker
+
+|`1`-`3`
+<|Symbols with symbol table text
+
+|`4`-`6`
+<|Annotations with symbol table text
```
almann: We refer to this elsewhere as "*symbol address* text".  I think we should be consistent here and I tend to like the address text versus "*symbol table* text".

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+|`0x4_`
+| `*`
+|Multibyte macro invocations
+
+.4+|`0x5_`
+| `0`-`8`
+|Integers
+
+|`9`
+<|Reserved
```
almann: I might make these things emphasized.

```suggestion
<|_Reserved_
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+|`0xB_`
+|`*`
+|S-expressions
+
+.3+|`0xC_`
+|`0`
+|Empty struct
+
+|`1`
+<|Reserved
```
almann:
```suggestion
<|_Reserved_
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+
+|`2`-`F`
+<|Structs with symbol address field names
+
+.2+|`0xD_`
+|`0`-`1`
+|Reserved
```
almann:
```suggestion
|_Reserved_
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+<|`null.null`
+
+|`B`
+<|Typed nulls
+
+|`C`-`D`
+<|NOP
+
+|`E`
+<|Reserved
```
almann:
```suggestion
<|_Reserved_
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+If the value of the opcode is less than `64` (`0x40`), it represents an invocation of the macro at the corresponding
+__address__—an offset within the local macro table.
+
+==== Example encoding of a single-byte invocation of the macro at address `7`
+[source]
+----
+0 0 0 0 0 1 1 1
+└──────┬──────┘
+  FixedUInt 7
+----
```
almann: Style note--I would model these as figures, not as sub-sections.  E.g.,

```asciidoc
.Figure {counter:figure-number}. Top-level E-expressions
[source,plain,%unbreakable]
----
a (:values 1 2 3) b (:void) c
a 1 2 3 b c
----
```

src/binary-encoding.adoc
almann: My interpretation is that the least four significant bits are in the op-code, followed by a `VarUInt` that encodes the remaining bits in LE order (each byte that follows modulo the encoding hold bits from lower to higher).  So in principle the bit order is consistent, so I don't quite follow the "backwards" point as well.

Now one thing I think one could expect (though I don't know if it is worth the complexity) is that the encoding of the `VarUInt` actually starts with the low nibble of the op-code, but I think that means some special purpose machinery for this case so I am good with the definition as it stands.

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+
+NOTE: From this point on in the document, example encodings are given in hexidecimal notation.
+
+[[booleans]]
+== Booleans
+
+`0x5E` represents boolean `true`, while `0x5F` represents boolean `false`.
+
+// XXX: Structurally, these examples should be at depth 3. However, all other example sections in the doc
+//      are at heading depth 4, which keeps them out of the TOC. These are also at depth 4 for consistency.
```
almann: More reason why to use the figure style 😃

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+==== Example encoding of `true`
+[source]
+----
+5e
+----
+
+==== Example encoding of `false`
+[source]
+----
+5f
```
almann: Super minor, I think we should standardize to upper-case for hexadecimal, we use upper elsewhere in the doc.

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+[[integers]]
+=== Integers
+
+Opcodes in the range `0x50` to `0x58` represent an integer. The opcode is followed by a <<fixedint, `FixedInt`>> that
+represents the integer value. The low nibble of the opcode (`0x_0` to `0x_8`) indicates the size of the `FixedInt`.
+Opcode `0x50` represents integer `0`; no more bytes follow.
+
+Integers that require more than 8 bytes are encoded using the variable-length integer opcode `0xF5`,
+followed by a
+<<varuint, VarUInt>> indicating how many bytes of representation data follow.
```
almann: Should there be an example for that below?

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1600 @@
...
+
+==== Example encoding of `3.14`
+[source]
+----
+┌──── Opcode in range 5A-5D indicates a float
+│┌─── Low nibble B indicates a 2-byte float
+││
+5B 42 47
+   └─┬─┘
+half-precision 3.14
```
almann: We have to be a little careful here with the examples, if `3.14` (I didn't check) is irrational with respect to base-2, the actual rational value has to be indicated for the smaller formats.

See: https://amazon-ion.github.io/ion-docs/docs/float.html

> In text, binary float is represented using familiar base-10 digits. While this is convenient for human representation, there is no explicit notation for expressing a particular floating point value as binary32 or binary64. Furthermore, many base-10 real numbers are irrational with respect to base-2 and cannot be expressed exactly in either binary floating point encoding (e.g. 1.1e0).
>
> Because of this asymmetry, the rules for Ion text float notation when round-tripping to Ion binary MUST be observed:
>
> * Any text notation that can be exactly represented as binary32 MAY be encoded as either binary32 or binary64 in Ion binary.
> * Any text notation that can only be exactly represented as binary64 MUST be encoded as binary64 in Ion binary.
> * Any text notation that has no exact representation (i.e. irrational in base-2 or more precision than the binary64 mantissa), MUST be encoded as binary64. This is to ensure that irrational numbers or truncated values are represented in the highest fidelity of the float data type.

almann (commented):
Overall, this is looking good!  I have reviewed up to *Timestamps* L685 and will break up my review into parts.

Some minor things so far:

* Let's use title case for all of our section headers.
* Examples should use some kind of "figure" syntax, sections are overweight for this.  I have an example that I've been using elsewhere that we could standardize on.

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1618 @@
...
+   │  └────── Exponent: FixedInt 3
+   └───────── Decimal length: VarUInt 1
+----
+
+[[timestamps]]
+== Timestamps
+
+NOTE: In Ion 1.0, text timestamp fields were encoded using the local time while binary timestamp fields were encoded
+using UTC time. This required applications to perform conversion logic when transcribing from one format to the other.
+*In Ion 1.1, all binary timestamp fields are encoded in local time.*
```
almann, line 732: I would add some "as per ISO 8601" or whatever.

src/binary-encoding.adoc
zslayton: There's an example of variable-length encoding, it's just being used on a value that's only a few bytes long.

```
==== Example variable-length encoding of `-944`
[source]
----
┌──── Opcode F5 indicates a variable-length integer, VarUInt length follows
│   ┌─── VarUInt 2; a 2-byte FixedInt follows
│   │
F5 05 50 fc
      └─┬─┘
   FixedInt -944
----

File: src/binary-encoding.adoc [hunk 51]
```diff
@@ -0,0 +1,1618 @@
...
+| Unused
+|===
+
+NOTE: Timestamps are encoded in Little Endian byte order. In the diagrams below, the bytes are read from left to right
+(least significant to most significant), while the bits within each byte should be read from right to left. (Again
+least significant to most significant.) +
+While this encoding may complicate the diagrams somewhat, it guarantees that the timestamp's subfields (`year`, `month`,
+etc.) occupy the same bit indexes regardless of how many bytes there are overall. (The last subfield,
+`fractional_seconds`, always begins at the same bit index when present, but can vary in length according to the
+precision.)
```
almann: I like this, though it is interesting that the least significant bits in our representation correspond to the most significant *field*.  So are timestamps have effective big-endian with respect to its fields.  I think this effectively makes timestamps mixed-endian.

src/binary-encoding.adoc
zslayton: I had confirmed that `3.14` could be roundtripped losslessly through `f32` and `f64` using the [Float Toy](https://evanw.github.io/float-toy/) website.

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1618 @@
...
+| Year
+
+| 3
+| Month or Day (see below)
+
+| 4
+| Illegal. The hour cannot be specified without also specifying minutes.
+
+| 5
+| Illegal
```
almann: Suggestion to make case consistent.

```suggestion
| 0
| _Illegal_.

| 1
| _Illegal_.

| 2
| Year

| 3
| Month or Day (see below)

| 4
| _Illegal_. The hour cannot be specified without also specifying minutes.

| 5
| _Illegal_.
```

src/binary-encoding.adoc
zslayton: > My interpretation is that the least four significant bits are in the op-code, followed by a VarUInt that encodes the remaining bits in LE order (each byte that follows modulo the encoding hold bits from lower to higher). So in principle the bit order is consistent, so I don't quite follow the "backwards" point as well.

This is what I was going for. 👍 I didn't move the continuation flag bits into the opcode's low nibble because it would require special machinery.

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1618 @@
...
+[[symbols_with_symbol_table_addresses]]
+=== Symbols with text in the symbol table
+
+Symbol values whose text can be found in the local symbol table are encoded using opcodes `0xE1` through `0xE3`:
+
+* `0xE1` represents a symbol whose address in the symbol table (aka its symbol ID) is a 1-byte
+<<fixeduint, `FixedUInt`>> that follows the opcode.
+* `0xE2` represents a symbol whose address in the symbol table is a 2-byte <<fixeduint, `FixedUInt`>> that follows
+the opcode.
+* `0xE3` represents a symbol whose address in the symbol table is a <<varuint,`VarUInt`>> that follows the opcode.
```
almann: Any particular reason why we special cased only up to 2-byte integers?  If we had the space  in the opcodes, 3-4 byte fixed symbols would probably be enough for all practical use cases (definitely 3), and save the variable length ones for ridiculously sized symbols (i.e., >4-5 bytes).

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1618 @@
...
+FE 31 49 20 61 70 70 6c 61 75 64 20 79 6f 75 72 20 63 75 72 69 6f 73 69 74 79
+      └────────────────────────────────┬────────────────────────────────────┘
+                            24 bytes of binary data
+----
+
+
+[[clobs]]
+=== Clobs
+
+Opcode `FF` indicates a clob--character data of an unspecified encoding.
```
almann:
```suggestion
Opcode `FF` indicates a clob--binary character data of an unspecified encoding.
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1618 @@
...
+Structs have 3 available encodings:
+
+. <<structs_with_symbol_address_field_names, Structs with symbol address field names>>
+. <<structs_with_varsym_field_names, Structs with `VarSym` field names>>
+. <<delimited_structs, Delimited structs with `VarSym` field names>>
+
+[[structs_with_symbol_address_field_names]]
+==== Structs with symbol address field names
+
+An opcode with a high nibble of `0xC_` indicates a struct with symbol address field names. The lower
```
almann:
```suggestion
An opcode with a high nibble of `0xC_` indicates a struct with symbol address field names (which is similar to the only available encoding of structs in Ion 1.0). The lower
```

src/binary-encoding.adoc
almann: Thanks! This was my major concern here.

src/binary-encoding.adoc
rmarrowstone: What differentiates this from the 1.0 `Var(U)Int` is that this may be called "definite" (you know the byte count in the beginning) while 1.0 could be called "indefinite" (you only know you're done when you're done).

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1618 @@
...
+[[encoding_primitives]]
+== Encoding primitives
+
+[[varuint]]
+=== `VarUInt`
+
+A variable-length unsigned integer.
+
+The bytes of a ``VarUInt``s are written in
+link:https://en.wikipedia.org/wiki/Endianness:[Little Endian byte order]. This means that the first byte will contain
```
rmarrowstone: > the first byte

do you mean "first byte_s_" or can we only represent 7-byte ints (an 8 byte VarInt would use 1 byte to describe length)?

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1618 @@
...
+the ``VarUInt``'s least significant bits.
+
+The least significant bits in the `VarUInt` indicate the number of bytes that were used to encode the integer.
```
rmarrowstone: Sorry for jumping in late and missing the convo here...

I assume we considered specifying the length prefix as the "first" bits, as opposed to the least significant in the first byte(s). That strikes me as cleaner/clearer _a priori_ but I'm guessing this was chosen after some _a posteriori_ observations that it is simpler/faster to implement as specified.

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1618 @@
...
+└───┬───┘        └──────┬──────┘  └──────┬──────┘
+lowest 6 bits    next 8 bits of   highest 8 bits
+of the unsigned  the unsigned     of the unsigned
+integer          integer          integer
+----
+
+NOTE: Ion 1.0 uses link:https://en.wikipedia.org/wiki/Endianness[Big Endian byte order] and indicates the width of the
+integer by having
+link:https://amazon-ion.github.io/ion-docs/docs/binary.html#varuint-and-varint-fields[an `end` flag bit in the highest
+ bit of each byte].
```
tgregg: Is it worth pointing out that the size characteristics are exactly the same between the two, or is that better left for a separate document?

src/binary-encoding.adoc
tgregg: Style note: the Wikipedia article spells this "little-endian" (hyphen, lowercase). That may be a good spelling to standardize on.

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1621 @@
...
+
+If the high nibble of the opcode is `0x4_`, then the low nibble represents the four least significant bits of the macro
+address. A <<varuint, `VarUInt`>> follows that contains the remaining, more significant bits.
+
+Because the first 64 macro addresses can already be encoded using high nibbles `0` to `3`, the decoded value is biased
+by 64. (That is: the reader must add 64 to the decoded value. If the decoded value is `0`, the macro address that it
+represents is `64`.)
+
+Because the address is encoded using a `VarUInt`, there is no (theoretical) limit to the number of addresses that can
+be invoked. However, larger addresses require more bytes to encode. This following table shows the number of bytes
```
```suggestion
be invoked. However, larger addresses require more bytes to encode. The following table shows the number of bytes
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1621 @@
...
+* 0 bits (0 bytes), representing the value 0e0 and indicated by opcode `0x5A`
+* 16 bits (2 bytes, link:https://en.wikipedia.org/wiki/Half-precision_floating-point_format[half precision]),
+indicated by opcode `0x5B`
+* 32 bits (4 bytes, link:https://en.wikipedia.org/wiki/Single-precision_floating-point_format[single precision]),
+indicated by opcode `0x5C`
+* 64 bits (8 bytes, link:https://en.wikipedia.org/wiki/Double-precision_floating-point_format[double precision]),
+indicated by opcode `0x5D`
+
+Note that in the Ion data model, float values are always 64 bits. However, if a value can be losslessly serialized
+in fewer than 64 bits, applications may choose to do so.
```
tgregg: Do we mean the following?

```suggestion
Note that in the Ion data model, float values are always 64 bits. However, if a value can be losslessly serialized
in fewer than 64 bits, Ion implementations may choose to do so.
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1621 @@
...
+┌──── Opcode in range 5A-5D indicates a float
+│┌─── Low nibble D indicates an 8-byte,
+││    double-precision value.
+5D 40 09 21 FB 54 44 2D 18
+   └──────────┬──────────┘
+double-precision 3.141592653589793
+----
+
+[[decimals]]
+=== Decimals
```
tgregg, line 652: Let's add an example of a non-zero int-like decimal value (exponent length 0)

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1621 @@
...
+seconds value to be greater than or equal to `1.0` or less than `0.0`. For this reason, both the coefficient and
+the exponent are encoded using unsigned types. The included coefficient `VarUInt` is unsigned to prevent the encoding of
+fractional seconds less than `0.0`. The exponent `FixedUInt` is implicitly negative, discouraging the encoding of
+decimal numbers greater than `1.0`. Note that validation is still required; namely:
+
+* An exponent value of `0` is illegal, as that would result in a fractional seconds greater than `1.0` (a whole second).
+* If `coefficient * 10^-exponent > 1.0`, that `(coefficient, exponent)` pair is illegal.
+
+If the timestamp's length is `3`, the most significant bit in the final byte (`h`) is a flag
+that indicates month (`0`) or day (`1`) precision.
```
```suggestion
If the timestamp's length is `3`, the most significant bit in the final byte (`h`) is a flag
that indicates month (`0`) or day (`1`) precision. If the timestamp's length is greater than `3`, the (`h`) bit is treated as the least-significant bit of the hour (`H`) bits.
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1621 @@
...
+│        │        │        ┌─── Opcode 0xF0 closes the most recently opened (and still open)
+│        │        │        │    delimited container: the outer list.
+│        │        │        │
+F1 51 01 F1 51 02 F0 51 03 F0
+   └─┬─┘    └─┬─┘    └─┬─┘
+     1        2        3
+----
+
+[[s_expressions]]
+=== S-Expressions
```
tgregg, line 1416: Can/should we just collapse this whole section by saying, basically, that the encodings are the same as for list, but for length-prefixed `0xB_` is used and for delimited `0xF2` is used?

src/binary-encoding.adoc
toddjonker: Those proposals all emphasize the variable-width aspect, which is fine in this context (and I kinda like "flex").  However that aspect is not distinguishing WRT fully fledged integers (or symbols), so the terms don't work well as type names in the macro-signature DSL.   There, I'm currently considering `plain_int` _etc._

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1621 @@
...
+<<varuint, `VarUInt`>> that indicates the byte length.
+
+Each field in the struct is encoded as a  <<varsym, `VarSym`>> field name, followed by an
+opcode-prefixed value.
+
+.Figure {counter:figure}: Length-prefixed encoding of `_{"foo": 1, $11: 2}_`
+[source,%unbreakable]
+----
+┌─── Opcode with high nibble `D` indicates a struct with VarSym field names
+│┌── Length: 7
```
```suggestion
│┌── Length: 9
```

src/binary-encoding.adoc
toddjonker: I think we should allow padding in both fixed and flex; we should be explicit here.

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1621 @@
...
+while opcode `0xF0` closes the most recently opened delimited container that has not yet been closed.
+
+NOTE: While length-prefixed structs can choose between <<structs_with_symbol_address_field_names, structs with
+symbol address field names>> and <<structs_with_varsym_field_names, structs with `VarSym` field names>>,
+delimited structs always use `VarSym`-encoded field names.
+
+.Figure {counter:figure}: Delimited encoding of the empty struct (`_{}_`)
+[source,%unbreakable]
+----
+┌─── Opcode 0xF3 indicates the beginning of a delimited struct with `VarSym` field names.│
```
```suggestion
┌─── Opcode 0xF3 indicates the beginning of a delimited struct with `VarSym` field names.
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1621 @@
...
+
+[[annotations]]
+=== Annotations
+
+Annotations can be encoded either <<annotations_with_symbol_addresses, as symbol addresses>>
+or <<annotations_with_varsym_text, as ``VarSym``s>>. In both encodings, the annotations sequence appears
+just before the value that it decorates.
+
+If an annotations sequence appears before one or more additional annotations sequences, the sequences
+are concatenated.
```
tgregg: I don't think I like this from an implementation perspective... In the reader, I'd love to be able to scan for values and record one start/end position for its annotation sequence that I can return to later if the user requests annotations. It's less simple (and quite possibly slower) if we're supporting N annotation sequences between values. Is there a killer use case I'm missing?

src/binary-encoding.adoc
```diff
@@ -633,43 +628,44 @@ double-precision 3.141592653589793
...
-of the exponent. It is possible for the coefficient to have a width of zero, indicating a coefficient of `0`.
+The body of the decimal is encoded as a <<varint, `VarInt`>> representing its coefficient, followed by a `FixedInt`
+representing its exponent. The width of the exponent is the total length of the decimal encoding minus the length
+of the coefficient. It is possible for the exponent to have a width of zero, indicating an exponent of `0`.
 
 Decimal values that require more than 14 bytes can be encoded using the variable-length decimal opcode: `0xF6`.
 
-A decimal with a coefficient of `-0` (which cannot be encoded in a `VarInt`) is encoded using opcode `6F`.
+A decimal with a coefficient of `-0` (which cannot be encoded as a `VarInt`) is encoded using opcode `6F`.
+The opcode is followed by a `VarUInt` representing the byte length and a `FixedInt` representing the exponent.
```
toddjonker: I'm guessing this needs to change as well

tgregg (commented):
Looking good.

toddjonker:
Personally, yeah I'd rather have it repeated, so that the section for each type is complete.

src/binary-encoding.adoc
toddjonker: Oh actually I guess it was wrong before (should've read "coefficient ... FixedInt").

But now I suggest this just be a single VarInt, no need for a length prefix.  It'll be more compact and faster!

src/binary-encoding.adoc
```diff
@@ -758,27 +754,37 @@ consume only enough bits to represent the fractional precision supported by the
...
+|-YYY:YYYY|
 +=========+
 ----
 
 ==== Opcode `0x71`: Month (2 bytes)
 [source]
 ----
 +=========+=========+
-|YYYY:YYYM|MMM-:----|
+|MYYY:YYYY|----:-MMM|
```
toddjonker: Yeah this is what I was talking about elsewhere, this feels more consistent endian-ness.

src/binary-encoding.adoc
```diff
@@ -948,76 +955,79 @@ that indicates month (`0`) or day (`1`) precision.
...
 [[symbols_with_inline_text]]
-=== Symbols with inline text
+=== Symbols With Inline Text
 
-==== Example encoding of a symbol with empty text (`''`)
-[source]
+If the high nibble of the opcode is `0x9_`, it represents a symbol whose text follows the opcode. The low nibble of the
+opcode indicates how many UTF-8 bytes follow. Opcode `0x90` represents a symbol with empty text (`''`).
+
+.Figure {counter:figure}: Encoding of a symbol with empty text (`_''_`)
```
toddjonker, line 1183: We've got seemingly inconsistent typesetting conventions; I don't understand why the `''` here, and other sample data elsewhere, is italicized.

src/binary-encoding.adoc
```diff
@@ -1327,7 +1337,8 @@ An opcode with a high nibble of `0xD_` indicates a struct with <<varsym, `VarSym
 nibble of the opcode indicates how many bytes were used to encode all of its nested `(field name, value)`
 pairs.
 
-WARNING: Empty structs MUST be written using `0xC0`. `0xD0` is a reserved opcode.
+WARNING: Empty length-prefixed structs MUST be written using `0xC0`. `0xD0` is a reserved opcode. Empty
+<<delimited_structs, delimited structs>> have their own encoding.
```
toddjonker:
```suggestion
WARNING: This form cannot be used to encode an empty struct; `0xD0` is a reserved opcode.  Empty structs can be written using either the length-prefixed form `0xC0` or the <<delimited_structs, delimited form>> `0xF3 0xF0`.
```

src/binary-encoding.adoc
toddjonker: Yeah, I would, so each section has full coverage of the type.

src/binary-encoding.adoc
toddjonker: I think just "symbol address" is good; "symbol address text" feels self-contradictory to me.

src/binary-encoding.adoc
zslayton: I meant it to be the Kleene star. I've replaced `*` with `A`-`F` for clarity.

src/binary-encoding.adoc
zslayton: Switched to "symbol address."

src/binary-encoding.adoc
zslayton: > But now I suggest this just be a single VarInt, no need for a length prefix. It'll be more compact and faster!

Sold!

src/binary-encoding.adoc
zslayton: Initial versions of this doc included lots of comparisons with the 1.0 encodings, but I found it to be pretty distracting from the core content. I think I'll remove this note, especially since we're going to rename `VarUInt`/`VarInt` in this version.

It would be good to have a different document that explained our experiences with the 1.0 encodings and our rationale for changing them.

src/binary-encoding.adoc
zslayton: Done. I think I saw the abbreviation "LE" so many times I internalized the capitalization. 😅

src/binary-encoding.adoc
zslayton: Good idea, done.

src/binary-encoding.adoc
zslayton: I eliminated all of the exposition as suggested, but I've kept the example encodings. Let me know if you think I should remove the examples as well.

src/binary-encoding.adoc
zslayton: Allowing 'chaining' means you can write annotation sequences of arbitrary length without buffering. However, in practice sequences are so rarely longer than 2 annotations (or ~4 bytes) that this probably isn't very useful. I'm ok with removing it.

src/binary-encoding.adoc
zslayton: The example data is being italicized in the context of figure labels. It's subjective, but I found the un-italicized data to be visually jarring:

image

while the italicized version felt more integrated:

image

src/binary-encoding.adoc
zslayton: Good question! Consider this example from the doc:
```
             ┌──── There's 1 zero in the least significant bits, so this
             │     integer is two bytes wide.
            ┌┴┐
0 1 1 0 0 1 1 0  0 0 0 0 1 0 1 1
└────┬────┘      └──────┬──────┘
lowest 6 bits    highest 8 bits
of the unsigned  of the unsigned
integer          integer
```

When the two bytes are read into an integer, the least significant bits of the first byte will be at the "end."

```
0 0 0 0 1 0 1 1  0 1 1 0 0 1 1 0
└──────┬──────┘  └────┬────┘ └┬┘
    highest 8      lowest 6   └── Continuation flag bits
      bits           bits
```

This enables a reader to examine the first byte in the serialized form and use a "count trailing zeros" operation (on some platforms this is "reverse bits" + "count leading zeros") to get a count of the bytes and read them into an integer. Then the entire integer can be shifted right by `num_bytes` (the number of flag bits) to get the value itself.

If instead we were to store the continuation flags in the "first" (highest) bits of the first byte(s) like this:
```
 ┌──── Continuation flag bits
┌┴┐
1 0 0 1 1 0 0 1  0 0 0 0 1 0 1 1
    └────┬────┘  └──────┬──────┘
lowest 6 bits    highest 8 bits
of the unsigned  of the unsigned
integer          integer
```
that would mean that the continuation bits would appear midway through the magnitude bits like this:
```
                  ┌──── Continuation flag bits
                 ┌┴┐
0 0 0 0 1 0 1 1  0 1 1 0 0 1 1 0
└──────┬──────┘      └────┬────┘ 
    highest 8          lowest 6   
      bits               bits
```

src/binary-encoding.adoc
zslayton: Added an `s`; there's no spec-defined size limitation on `VarInt`.

src/binary-encoding.adoc
rmarrowstone: Makes sense. Thanks!

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1618 @@
...
+       bytes
+----
+
+TODO: Demonstrate splicing macro values into the struct via VarSym escape code `0x00`.
+
+[[delimited_structs]]
+==== Delimited structs
+
+Opcode `0xF3` indicates the beginning of a delimited struct with <<varsym, `VarSym`>> field names,
+while opcode `0xF0` closes the most recently opened delimited container that has not yet been closed.
```
almann: I think the generic closing text needs to be changed to illustrate the `VarSym` requirements.

```suggestion
Opcode `0xF3` indicates the beginning of a delimited struct with <<varsym, `VarSym`>> field names,
while the `VarSym` control sequence `0x00 0xF0` closes the current struct.
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1618 @@
...
+----
+
+[[nulls]]
+=== Nulls
+
+The opcode `0xEA` indicates an untyped null (that is: `null`, or its alias `null.null`).
+
+The opcode `0xEB` indicates a typed null; a byte follows whose value indicates the type. The byte-to-type
+mapping is as follows:
+
```
almann: I wonder if we should call this the "constant" opcode and only happen to have nulls here.  Also I would indicate that "all octets that are not in this table are illegal and reserved for future use."

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1618 @@
...
+┌──── The opcode `0xEA` represents a null (null.null)
+EA
+----
+
+==== Example encoding of `null.string`
+[source]
+----
+┌──── The opcode `0xEB` indicates a typed null; a byte indicating the type follows
+│  ┌──── Byte 0x05 indicates the type `string`
+EB 0x5
```
almann:
```suggestion
┌──── The opcode `0xEB` indicates a typed null; a byte indicating the type follows
│  ┌──── Byte `0x05` indicates the type `string`
EB 05
```

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1618 @@
...
+
+If an annotations sequence appears before one or more additional annotations sequences, the sequences
+are concatenated.
+
+It is illegal for an annotations sequence to appear before any of the following:
+
+* The end of the stream
+* A <<nops, `NOP`>>
+// TODO: Links
+* An E-expression (that is: a macro invocation). To add annotations to a macro invocation, see the
```
almann: I just realized that in the table *and* the macro invocation sections we don't call out these things as E-Expressions, I think we should be clearer about that there.

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1618 @@
...
+`annotate` macro.
+
+[[annotations_with_symbol_addresses]]
+==== Annotations with symbol addresses
+Opcodes `0xE4` through `0xE6` indicate one or more annotations encoded as symbol addresses. If the opcode is:
+
+* `0xE4`, a single <<varuint, `VarUInt`>>-encoded symbol address follows.
+* `0xE5`, two <<varuint, `VarUInt`>>-encoded symbol addresses follow.
+* `0xE6`, a <<varuint, `VarUInt`>> follows that represents the number of bytes needed to encode
+the annotations sequence, which can be made up of any number of `VarUInt` symbol addresses.
```
almann: Do we have the 1.0 style fully length prefixed variant somewhere?

src/binary-encoding.adoc
```diff
@@ -0,0 +1,1618 @@
...
+            └──┬───┘    └── The annotated value: `false`
+            3 UTF-8
+             bytes
+----
+
+[[nops]]
+=== ``NOP``s
+
+A `NOP` (short for "no-operation") is the binary equivalent of whitespace. `NOP` bytes have no meaning,
+but can be used as padding to achieve a desired alignment.
```
almann, line 1854: I know the restrictions follow as 1.0, but it might be worth enumerating where `NOP` is allowed and how this works in the new flavors of struct.  I already saw the one error case of `NOP` after annotations--I kind of forgot if that was a 1.0 error too, so re-enumerating the validity is probably a good thing.

almann (commented):
Finished my first pass through annotations--looks good overall, minor comments below.

src/binary-encoding.adoc
zslayton: > Any particular reason why we special cased only up to 2-byte integers? If we had the space in the opcodes, 3-4 byte fixed symbols would probably be enough for all practical use cases (definitely 3)

2 bytes gets us 65k+ addresses, which covers most use cases. I agree that 3 bytes would cover all practical use cases. There's one more `Reserved` slot in the `0xE_` nibble if we'd like to use it for that.

src/binary-encoding.adoc
```diff
@@ -688,203 +733,329 @@ NOTE: In Ion 1.0, text timestamp fields were encoded using the local time while
...
+The timestamp's fractional second precision (if present) is either 3 digits (milliseconds), 6 digits (microseconds),
+or 9 digits (nanoseconds).
+
+==== Opcodes by precision and offset
+
+Each opcode with a high nibble of `0x7_` indicates a different precision and offset encoding pair.
+
+[cols="^1a,^1a,^1a,.^4a"]
+|===
+|Opcode | Precision | Serialized size in bytes{asterisk} | Offset encoding
```
almann, line 777: This is a great table!

src/binary-encoding.adoc
zslayton: Good catch!

src/binary-encoding.adoc
zslayton: We don't at the moment.

src/binary-encoding.adoc
zslayton: One other thought: because the symbol addresses are biased according to their representation, the `FlexUInt`  encoding starts at 65k+. The 3-byte `FlexUInt` encoding:
```
E3 FE FF
```
would represent 16383, which when biased produces symbol address `$82175`.

A 4-byte `FlexUInt` encoding (the same size as an opcode + 3 bytes) can store up to:
```
E3 FC FF FF
```
which comes out to 2097151 before biasing and 2,162,943 after.

A dedicated 3-byte representation would hold a lot more, but I wanted to spell out the current encoding's capacity before we decide to use another opcode.

src/binary-encoding.adoc
zslayton: Added.

src/binary-encoding.adoc
zslayton: Good call, done.

src/binary-encoding.adoc
```diff
@@ -930,23 +1101,24 @@ Unlike the short-form encoding, the long-form encoding reserves:
 If the timestamp's length is greater than or equal to `8`, it has fractional seconds that are encoded as a
 `(coefficient, exponent)` pair, similar to a <<decimals, decimal>>. However, it is illegal for the fractional
 seconds value to be greater than or equal to `1.0` or less than `0.0`. For this reason, both the coefficient and
-the exponent are encoded using unsigned types. The included coefficient `VarUInt` is unsigned to prevent the encoding of
+the exponent are encoded using unsigned types. The included coefficient `FlexUInt` is unsigned to prevent the encoding of
 fractional seconds less than `0.0`. The exponent `FixedUInt` is implicitly negative, discouraging the encoding of
 decimal numbers greater than `1.0`. Note that validation is still required; namely:
```
almann: What is interesting, is what we're calling an implicitly negative exponent, is what other people would call *scale* and not only that, but our restrictions on the *coefficient*/*scale* are what, in other contexts, would be called a *mantissa*.  I am not sure that we need to introduce mantissa (fractional seconds is probably good enough), but we might consider calling this negative exponent *scale*.

almann (commented):
These revisions look good to me (primarily around re-naming `Var...` to `Flex...`.  Minor comments  below.

src/binary-encoding.adoc
almann: We should add a TODO somewhere to remind us about this.

src/binary-encoding.adoc
almann: That's fair--I am not really eager to specify this if we think we don't need it.

src/binary-encoding.adoc
almann: > What differentiates this from the 1.0 Var(U)Int is that this may be called "definite" (you know the byte count in the beginning) while 1.0 could be called "indefinite" (you only know you're done when you're done).

My issue with this, is that all of the 1.1 encodings are definite.  You know the byte count for `Fixed...` integers as well.  Definite conveys something very specific about a difference of 1.0/1.1 variable length integers that I don't think stands on its own.

src/binary-encoding.adoc
zslayton: I'm ok with renaming those, but I'd like to leave that for a future change as I'll need to read up on the terminology.

src/binary-encoding.adoc
zslayton: Added.

src/binary-encoding.adoc
```diff
@@ -646,15 +674,25 @@ The opcode is followed by a `VarUInt` representing the byte length and a `FixedI
...
 ----
 
-.Figure {counter:figure}: Encoding of decimal `_1.27_`
+.Figure {counter:figure}: Encoding of decimal `_7d0_`
+[source,%unbreakable]
+----
+┌──── Opcode in range 60-6F indicates a decimal
+│┌─── Low nibble 1 indicates a 1-byte decimal
+││
+60 0F
```
```suggestion
61 0F
```

tgregg (approved):
All my comments have been addressed.

almann (approved):
🚢

src/binary-encoding.adoc
almann: Added #220.

src/binary-encoding.adoc
almann: Added #221.

PR #216, opened 2023-05-12
Update Macros by Example chapter based on recent decisions.

toddjonker:
* Parameter declarations are now name-first to align with Lisp best-practices and for consistency with `for` and future binding forms.
* "Argument cardinality" is now expressed as "grouping", declared by `[]`.
* Added `...+` to mirror `[]+` the same way that `...` mirrors `[]*`.
* "Stream cardinality" is now just "cardinality", and a cardinality modifier is required when a type is declared.
* When no type+cardinality is declared, the default is `any*` rather than `any!`.
* Replaced the `tagless` modifier with a smaller set of primitive types.

See #217 for related Glossary changes.

----

### DIFF
--- a/src/macros-by-example.adoc
+++ b/src/macros-by-example.adoc
@@ -301,8 +301,7 @@ The intended input domain is now clear and the Ion parser can emit an error soon
 In this context the types include all the normal “concrete” Ion types, abstract
 supertypes like `*number*`, `*text*`, and `*lob*`, and the unconstrained “top type” `*any*`.
 The latter is the default type, and the signature `[foo]` is equivalent to `[(foo *any* *{asterisk}*)]`
-meaning
-that the parameter `foo` accepts zero or more values of any type.
+meaning that the parameter `foo` accepts zero or more values of any type.
 
 TIP: These types also serve a second purpose: they can allow the binary encoding to be more compact by
 avoiding type tags or using fixed-width values.
@@ -978,7 +977,7 @@ The binary encoding of macro-shaped parameters are similarly tagless, eliding an
 mentioning `point` and just writing its arguments with minimal delimiting.
 
 Macro types can be grouped and/or combined with any cardinality, following the same rules as
-before.  Note that grouped macro types require callers to use two layers of delimiting
+before.  Note that grouped macro shapes require callers to use two layers of delimiting
 containers: and outer list for the group, and an inner S-expression for
 each macro instance:
 
--- a/src/macros-by-example.adoc
+++ b/src/macros-by-example.adoc
@@ -69,9 +69,9 @@ term “encoding expression”.
 
 NOTE: Any Ion 1.1 document with macros can be fully-expanded into an equivalent Ion 1.0 document.
 
-We’ll streamline future examples with a couple conventions.  First, assume that any E-expression
+We can streamline future examples with a couple conventions.  First, assume that any E-expression
 is occurring within an Ion 1.1 document;
-second, we’ll use the relation notation, `⇒`, to mean “expands to”.  So we can say:
+second, we use the relation notation, `⇒`, to mean “expands to”.  So we can say:
 
 ----
 (:pi) ⇒ 3.141592653589793
@@ -106,27 +106,35 @@ The struct
 in this macro is our first non-trivial _template_, an expression in Ion’s new domain-specific language
 for defining macro functions.  This expression language treats Ion scalar values (except for
 symbols) as literals, giving the decimal in ``pi``’s template its intended meaning.  Expressions
-that are structs are interpreted as _quasi-literals_: the field names are literal, but the field
+that are structs are interpreted _almost_ literally: the field names are literal, but the field
 “values” are arbitrary expressions.  This is why the `amount` and `currency` field names show up
-as-is in the expansion.  The sub-expressions in this template demonstrate that the expression
-language treats symbols as variable references.  Here, `a` and `c` in the template refer to the
-parameters of the macro, and during expansion they are “filled in” with the values supplied by
-the invocation of the macro.
+as-is in the expansion.  We call these almost-literal forms _quasi-literals_.
 
 The template language also treats lists quasi-literally, and every element inside the list is an
 expression.  Here’s a silly macro to illustrate:
 
 [{mrk}]
 ----
-(*macro* reverse [a, b] [b, a])
+(*macro* reverse [a, c] [c, a])
 ----
 ----
-(:reverse first {amount:a, currency:c}) ⇒ [{amount:a, currency:c}, first]
+(:reverse first 1990) ⇒ [1990, first]
 ----
 
-Note that the `a` in the E-expression is _not_ part of the
+The sub-expressions in these templates demonstrate that the expression language treats symbols as
+variable references.  Each `a` and `c` in the templates above refer to the parameters of the
+surrounding macro, and during expansion they are “filled in” with the values supplied by
+the invocation of the macro.
+
+These names are part of the macro language that have no relation to data encoded using the macro:
+
+----
+(:reverse c {amount:a, currency:c}) ⇒ [{amount:a, currency:c}, c]
+----
+
+The `a` and ``c``s in this E-expression is _not_ part of the
 expression language and not a reference to the macro’s first parameter or any other named entity.
-From the point of view of ``reverse``’s template, the input struct is literal data.
+From the point of view of ``reverse``’s template, the inputs are literal data.
 
 E-expressions can nest, so we could also encode the same data using `price`:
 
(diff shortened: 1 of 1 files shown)


src/macros-by-example.adoc
```diff
@@ -106,27 +106,35 @@ The struct
...
-the invocation of the macro.
+as-is in the expansion.  We call these almost-literal forms _quasi-literals_.
 
 The template language also treats lists quasi-literally, and every element inside the list is an
 expression.  Here’s a silly macro to illustrate:
 
 [{mrk}]
 ----
-(*macro* reverse [a, b] [b, a])
+(*macro* reverse [a, c] [c, a])
```
zslayton: Curious about this change--do `a` and `c` relate to `amount` and `currency` in the example below? Now that we're using non-consecutive letters, my brain is trying to ascribe them meaning.

src/macros-by-example.adoc
```diff
@@ -384,7 +397,8 @@ returns too-few or too-many values.
 
 [{mrk}]
 ----
-(*macro* reverse [a, b] [b, a])
+(*macro* reverse [(a *any!*), (b *any!*)]
+  [b, a])
```
zslayton: Since `!` is something that we've mentioned only briefly, maybe worth a comment reminding the reader what that means alongside the definition?

```suggestion
(*macro* reverse [(a *any!*), (b *any!*)] // Recall that `!` is "exactly one" cardinality
  [b, a])
```

(I notice you've kept the names `a` and `b` here.)

src/macros-by-example.adoc
```diff
@@ -541,46 +559,41 @@ clarity of intent and terminology.
...
 
 
 ==== Exactly-One
 
-Most parameters require exactly one value and thus have _exactly-one cardinality_.  This is the
-default when a signature has no modifier, but the `*!*` modifier can be used for clarity.
-
-TIP: The signatures `[_param_]`, `[(*any* _param_)]`, and `[(*any!* _param_)]` are all
-equivalent.
+Most parameters require exactly one value and thus have _exactly-one cardinality_.
```
zslayton: > Most parameters require exactly one value

I think this is true for these examples, but not necessarily in general.

src/macros-by-example.adoc
```diff
@@ -631,24 +643,94 @@ just last place.
...
+
+The non-rest versions of multi-value parameters can be annoying to invoke, since they require the
+use of `values` or some other template to produce the stream of values.  To streamline invocation,
+a macro can opt-in to special syntax that uses a list as delimiting syntax to group the
+applicable sub-expressions.  This is denoted by wrapping the parameter's type in `[]`:
+
+[{mrk}]
+----
+(*macro* prices
+  [(amount *[number]{asterisk}*), (currency *symbol!*)]
```
zslayton, line 717: I was under the impression that `*` was the default for grouped expressions and the only explicit cardinality for those was `+`.

src/macros-by-example.adoc
```diff
@@ -764,50 +835,63 @@ the type tags are unnecessary and may add up to a non-trivial amount of wasted s
...
 
+The following primitive types are available:
+
+[cols="1,1"]
+|===
+|*Primitive Type*                               |*Description*
+| `*var_symbol*`                                | Tagless symbol (SID or text)
+| `*var_string*`                                | Tagless string
+| `*var_int*`                                   | Tagless, variable-width signed int
+| `*uint*`                                      | Tagless, variable-width unsigned int
```
zslayton: For symmetry with the signed int primitives, I think I'd spell `uint` as `var_uint`.

zslayton (approved):
Looks good! A few non-blocking comments below.

src/macros-by-example.adoc
zslayton: Reading through the glossary changes in #217, I realize that I've been thinking this because `[]` is a grouping mode like `...`, and we don't require `...*`, only `...+`.

src/macros-by-example.adoc
toddjonker: So I was trying to set up a more glaring set of conflicts for the prose below discussing scoping and expansion-time versus compile-time symbols.   It's perhaps more confusing than helpful, so I'll revisit.

src/macros-by-example.adoc
toddjonker: How about "many" instead of "most"?

src/macros-by-example.adoc
toddjonker: I didn't remember what we resolved during the onsite, so I opted for consistency in always requiring that marker.  It did feel redundant, so I'd be happy to remove it.   Your point about `...` is a good supporting argument.

@almann Do you agree with making the `*` optional and default for `[]`?

src/macros-by-example.adoc
toddjonker: Agree, with the caveat that we gotta re-spell `var_`.

zslayton: `elastic_int` 😛

src/macros-by-example.adoc
toddjonker: Changed to "many" in [738bce8](https://github.com/amazon-ion/ion-docs/pull/216/commits/738bce85ac87a1f306351b4f54a9396c05b70d2a)

src/macros-by-example.adoc
```diff
@@ -115,14 +115,14 @@ expression.  Here’s a silly macro to illustrate:
...
-(*macro* reverse [a, c] [c, a])
+(*macro* reverse [a, b] [b, a])
 ----
 ----
 (:reverse first 1990) ⇒ [1990, first]
 ----
 
 The sub-expressions in these templates demonstrate that the expression language treats symbols as
-variable references.  Each `a` and `c` in the templates above refer to the parameters of the
+variable references.  Each `a`, `b`, and `c` in the templates above refer to the parameters of the
```
zslayton, line 125: Is the `c` here referring to the one declared on line 88?

```
(*macro* price
  [a, c]                          // signature
  { amount: a, currency: c })     // template
```

toddjonker: Yeah, that was the intent, but it's awkward for sure.  Ideas to clarify?

src/macros-by-example.adoc
```diff
@@ -275,27 +283,29 @@ To detect problems close to their source, macro signatures can declare type cons
...
-
-This example reveals additional syntax for parameter declarations.  So far, a parameter was
-declared by a symbol denoting its name, now we have an S-expression pairing a type and a name.
 In this context the types include all the normal “concrete” Ion types, abstract
 supertypes like `*number*`, `*text*`, and `*lob*`, and the unconstrained “top type” `*any*`.
-The latter is the default type, so the signature `[foo]` is equivalent to `[(*any* foo)]` meaning
-that the parameter `foo` accepts any value.
+The latter is the default type, and the signature `[foo]` is equivalent to `[(foo *any* *{asterisk}*)]`
+meaning
+that the parameter `foo` accepts zero or more values of any type.
```
almann: I think the wrapping is just a bit off here:

```suggestion
The latter is the default type, and the signature `[foo]` is equivalent to `[(foo *any* *{asterisk}*)]`
meaning that the parameter `foo` accepts zero or more values of any type.
```

src/macros-by-example.adoc
almann: I am good with this.

src/macros-by-example.adoc
```diff
@@ -888,19 +977,19 @@ parameter produces a stream of structs.
 The binary encoding of macro-shaped parameters are similarly tagless, eliding any opcodes
 mentioning `point` and just writing its arguments with minimal delimiting.
 
-Macro types can be combined with any cardinality, following the same rules as before.  Note that
-when combined with `***` or `*+*` this requires callers to use two layers of delimiting
-containers, and outer delimiter for the parameter-stream as a whole, and an inner delimiter for
+Macro types can be grouped and/or combined with any cardinality, following the same rules as
+before.  Note that grouped macro types require callers to use two layers of delimiting
```
almann:
```suggestion
before.  Note that grouped macro shapes require callers to use two layers of delimiting
```

almann (approved):
🚢 LGTM, minor suggestions below.

src/macros-by-example.adoc
toddjonker: OK, I'll make that change in a separate CR.

toddjonker: Updated in #218

src/macros-by-example.adoc
zslayton: Maybe a rephrasing parenthetical?

> The symbols in the templates above (`a` and `c` in the `price` macro, `a` and `b` in the `reverse` macro) refer to the parameters of their respective surrounding macros [....]

toddjonker: Rephrased in #226 thanks for the great suggestion!

PR #219, opened 2023-05-16
Update Modules by Example chapter based on recent decisions.

toddjonker:
* Parameter declarations are now name-first to align with Lisp best-practices and for consistency with `for` and future binding forms.
* Encoding directives are now S-expressions.
* Shared modules have a new S-expression syntax.
* Encoding directives and modules use similar load, use, symbol_table, and macro_table clauses.

There's more that should be added here, so consider TODO:
* Bringing back coverage of "tunneled modules" embedded in `$ion_shared_symbol_table`.
* Examples of module names inside `symbol_table` and `macro_table`.
* I'm increasingly dissatisfied with the long comparison with symbol table directives. Would be better to start fresh here and cover legacy concerns in a separate chapter.

### Description of changes:

Sorry for another lengthy PR, but these were fairly global changes.

----

### DIFF
--- a/src/modules-by-example.adoc
+++ b/src/modules-by-example.adoc
@@ -86,11 +86,11 @@ Upon encountering the `$ion_1_1` IVM, the environment is reset to the default st
   * The macro table is empty.
   * The symbol table is the Ion 1.1 system symbol table.
 
-To customize this environment, we use an _encoding directive_: a
-top-level struct annotated with `$ion_encoding`. Like `$ion_symbol_table`, this directive defines a
-new encoding environment that goes into effect immediately after the struct closes.
+To customize this environment, we use an _encoding directive_: a top-level S-expression annotated
+with `$ion_encoding`. Like `$ion_symbol_table`, this directive defines a new encoding environment
+that goes into effect immediately after the directive closes.
 
-NOTE: We use the term "encoding directive" to refer to the `$ion_encoding` struct, and "local
+NOTE: We use the term "encoding directive" to refer to the `$ion_encoding` S-expression, and "local
 symbol table directive" to refer to the `$ion_symbol_table` struct.  Both forms are valid in
 Ion 1.1.
 
@@ -111,6 +111,9 @@ $ion_encoding::(
 Each syntactic form affects one of the main components of the environment.
 The `*symbol_table*` and `*macro_table*` clauses specify the layout of those tables, while the
 preceding clauses enumerate the available modules that may be installed into them.
+
+NOTE: Using an S-expression instead of a struct constrains the order in which
+clauses are encountered, making it both more code-like and easier to parse.
 
 Let’s look at some examples illustrating the relation between `$ion_symbol_table` and
 `$ion_encoding`.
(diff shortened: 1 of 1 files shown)


src/modules-by-example.adoc
```diff
@@ -716,44 +742,43 @@ exported implicitly via `*macro*` or explicitly via `*export*`.
...
-**$ion_encoding**::{
-  modules: [(*module* local
-              (*symbols* ["s1", "s2"])
-              (*macro* m1 ...))],
-  install_symbols: [local]
-  install_macros:  [local]
-}
+**$ion_encoding**::(
+  (*module* mod1
+    (*symbol_tables* ["s1", "s2"])
```
zslayton:
```suggestion
    (*symbol_table* ["s1", "s2"])
```

zslayton (approved):
Nice! +1 to the idea of making a new home for the legacy comparison stuff.

src/modules-by-example.adoc
```diff
@@ -98,16 +98,19 @@ The general syntax of an encoding directive is as follows:
...
-  install_macros:  [ /* names declared in 'modules' field */ ]
-}
+$ion_encoding::(
+  (*retain* ...)        // Reuse selected modules from the current segment
+  (*load* ...)          // Get a shared module from the catalog
+  (*module* ...)        // Define a new module inline
+  ...
+  (*symbol_table* ...)  // Install modules into the symbol table
+  (*macro_table* ...)   // Install modules into the macro table
+)
```
almann, line 108: Is it worth noting somewhere why we prefer the S-Expression for this?

Something like:

> In Ion 1.1, we do not use a struct to encode the _encoding directive_ because the order makes it easier to parse.

src/modules-by-example.adoc
```diff
@@ -407,77 +403,85 @@ a generalization of Ion 1.0's shared symbol tables. As discussed in
...
-  )
-}
+    (*macro* poly  [(first *point!*), (second *point!*), (rest *point\...+*)]
+      [first, second, rest]))
+)
 ----
 
-The `module` field here is very similar to the `*module*` S-expression inside `$ion_encoding`.
-Here, no symbolic name is declared, since one will be assigned when the module is ``*import*``ed.
+This S-expression is very similar to the `*module*` S-expression inside `$ion_encoding`.
```
almann, line 432: We probably have to search replace "S-expression" to "S-Expression".  I think that is the preferred spelling, yes?

almann (approved):
🚢 Looks great!  Minor comments/questions below.

src/modules-by-example.adoc
toddjonker: Traditional convention is "S-expression" with upper S, lower e.

* https://amazon-ion.github.io/ion-docs/docs/spec.html#sexp
* https://en.wikipedia.org/wiki/S-expression

src/modules-by-example.adoc
toddjonker: Good idea, I also caught peripheral text that still said `struct`.

Issue #222, opened 2023-05-17
Add Long-Form Timestamp Examples


almann:
There is enough complexity (e.g., month/day precision) in the different lengths of binary timestamp to provide examples up-to and including fractional seconds.  We have only one large, just the body example currently.

PR #223, opened 2023-05-17
Adds new diagrams for timestamps

### LINKED ISSUE #220: Rework 1.1 Timestamps
This is currently fairly confusing, we should rework this section.

> I like this, though it is interesting that the least significant bits in our representation correspond to the most significant *field*.  So are timestamps have effective big-endian with respect to its fields.  I think this effectively makes timestamps mixed-endian.

_Originally posted by @almann in https://github.com/amazon-ion/ion-docs/pull/215#discussion_r1195614957_

### LINKED ISSUE #221: Ion 1.1 Timestamp Scale/Mantissa
> What is interesting, is what we're calling an implicitly negative exponent, is what other people would call *scale* and not only that, but our restrictions on the *coefficient*/*scale* are what, in other contexts, would be called a *mantissa*.  I am not sure that we need to introduce mantissa (fractional seconds is probably good enough), but we might consider calling this negative exponent *scale*.

_Originally posted by @almann in https://github.com/amazon-ion/ion-docs/pull/215#discussion_r1196850547_

almann:
Makes the diagrams vertical to make the endian aspects easier to read.
Also changes the text to emphasize that the bit-packed parts of
timestamp are interpreted within a `FixedUInt`.  Updates the text around
long-form timestamp to use *scale* instead of *exponent* to be more
consistent with nomenclature in the wild.

Decided against the use of *mantissa* to describe fractional seconds,
because while the former is concise, it doesn't really reduce text and
may be harder for readers to understand.

Fixes #220, Fixes #221

----

### DIFF
--- a/src/binary-encoding.adoc
+++ b/src/binary-encoding.adoc
@@ -841,8 +841,8 @@ though it is less compact than opcodes `0x73`-`0x77` above.
 |===
 _{asterisk} Serialized size in bytes does not include the opcode._
 
-The body of short-form timestamps are encoded as a `FixedUInt` of the specified size of the op-code.  This integer is
-then partitioned in bit-fields for the timestamp's subfields.  Note that endian does not apply here because the
+The body of short-form timestamps are encoded as a `FixedUInt` of the size specified by the opcode.  This integer is
+then partitioned into bit-fields representing the timestamp's subfields.  Note that endianness does not apply here because the
 bit-fields are defined over the body interpreted as an _integer_.
 
 The following letters to are used to denote bits in each subfield in diagrams that follow. Subfields occur in the same
@@ -896,7 +896,7 @@ consume only enough bits to represent the fractional precision supported by the 
 | Unused
 |===
 
-We will denote the timestamp encoding as follows with each byte ordered from vertically from top to bottom.  The
+We will denote the timestamp encoding as follows with each byte ordered vertically from top to bottom.  The
 respective bits are denoted using the letter codes defined in the table above.
 
 [%unbreakable,source]
@@ -914,7 +914,7 @@ byte 0   |  0xNN   | <-- hex notation for constants like opcodes
          +=========+
 ----
 
-The bytes are read from top to down (least significant to most significant), while the bits within each byte should be
+The bytes are read from top to bottom (least significant to most significant), while the bits within each byte should be
 read from right to left (also least significant to most significant.)
 
 NOTE: While this encoding may complicate human reading, it guarantees that the timestamp's subfields (`year`, `month`,
--- a/src/binary-encoding.adoc
+++ b/src/binary-encoding.adoc
@@ -841,7 +841,7 @@ though it is less compact than opcodes `0x73`-`0x77` above.
 |===
 _{asterisk} Serialized size in bytes does not include the opcode._
 
-The body of short-form timestamps are encoded as a `FixedUInt` of the size specified by the opcode.  This integer is
+The body of a short-form timestamp is encoded as a `FixedUInt` of the size specified by the opcode.  This integer is
 then partitioned into bit-fields representing the timestamp's subfields.  Note that endianness does not apply here because the
 bit-fields are defined over the body interpreted as an _integer_.
 
@@ -1220,8 +1220,8 @@ Unlike the short-form encoding, the long-form encoding reserves:
 Similar to short-form timestamps, with the exception of representing the fractional seconds, the components of the
 timestamp are encoded as bit-fields on a `FixedUInt` that corresponds to the length that followed the opcode.
 
-If the timestamp's overall length is greater than or equal to `8`, the `FixedUInt` part of the timestamp is `8` bytes
-and the remainder of the length is used to encode fractional seconds. The fractional seconds are encoded as a
+If the timestamp's overall length is greater than or equal to `8`, the `FixedUInt` part of the timestamp is `7` bytes
+and the remaining bytes are used to encode fractional seconds. The fractional seconds are encoded as a
 `(coefficient, scale)` pair, which is _similar_ to a <<decimals, decimal>>. The primary difference is that the *scale*
 represents a negative *exponent* because it is illegal for the fractional seconds value to be greater than or equal to
 `1.0` or less than `0.0`. The coefficient is encoded as a `FlexUInt` (instead of `FlexInt`) to prevent the encoding of
--- a/src/binary-encoding.adoc
+++ b/src/binary-encoding.adoc
@@ -841,6 +841,10 @@ though it is less compact than opcodes `0x73`-`0x77` above.
 |===
 _{asterisk} Serialized size in bytes does not include the opcode._
 
+The body of short-form timestamps are encoded as a `FixedUInt` of the specified size of the op-code.  This integer is
+then partitioned in bit-fields for the timestamp's subfields.  Note that endian does not apply here because the
+bit-fields are defined over the body interpreted as an _integer_.
+
 The following letters to are used to denote bits in each subfield in diagrams that follow. Subfields occur in the same
 order in all encoding variants, and consume the same number of bits, with the exception of the fractional bits, which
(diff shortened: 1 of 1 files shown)


src/binary-encoding.adoc
```diff
@@ -841,6 +841,10 @@ though it is less compact than opcodes `0x73`-`0x77` above.
 |===
 _{asterisk} Serialized size in bytes does not include the opcode._
 
+The body of short-form timestamps are encoded as a `FixedUInt` of the specified size of the op-code.  This integer is
+then partitioned in bit-fields for the timestamp's subfields.  Note that endian does not apply here because the
+bit-fields are defined over the body interpreted as an _integer_.
+
```
zslayton:
```suggestion
The body of short-form timestamps are encoded as a `FixedUInt` of the size specified by the opcode.  This integer is
then partitioned into bit-fields representing the timestamp's subfields.  Note that endianness does not apply here because the
bit-fields are defined over the body interpreted as an _integer_.

```

I think the "here" in "does not apply here" could be spelled out a bit more.

src/binary-encoding.adoc
```diff
@@ -887,24 +891,30 @@ consume only enough bits to represent the fractional precision supported by the
...
-| `-`
+| `.`
 | n/a
 | Unused
 |===
 
-Timestamps are encoded in little-endian byte order. In the diagrams below, the first byte shows the leading opcode
-in hexadecimal; the trailing bytes show the meaning and layout of their respective bits using the letter codes defined
-in the table above.
+We will denote the timestamp encoding as follows with each byte ordered from vertically from top to bottom.  The
```
zslayton:
```suggestion
We will denote the timestamp encoding as follows with each byte ordered vertically from top to bottom.  The
```

src/binary-encoding.adoc
```diff
@@ -887,24 +891,30 @@ consume only enough bits to represent the fractional precision supported by the
...
 ----
-  byte 1           byte 2                         byte N
-+========+========================+=====+========================+
-| OPCODE | high nibble:low nibble | ... | high nibble:low nibble |
-+========+========================+=====+========================+
+          7       0 <--- bit position
+          |       |
+         +=========+
+byte 0   |  0xNN   | <-- hex notation for constants like opcodes
+         +=========+ <-- boundary between some encoding primitive (e.g., opcode/`FlexUInt`)
```
zslayton: > between some encoding primitive (...) 

The word "between" made me expect a second clause. ("Between some encoding primitive and ________").

src/binary-encoding.adoc
```diff
@@ -887,24 +891,30 @@ consume only enough bits to represent the fractional precision supported by the
...
+     1   |nnnn:nnnn| <-- bits denoted with a `:` as a delimeter to aid in reading
+         +---------+ <-- octet boundary within an encoding primitive
+         ...
+         +---------+
+     N   |nnnn:nnnn|
+         +=========+
 ----
 
-The bytes are read from left to right (least significant to most significant), while the bits within each byte should be
+The bytes are read from top to down (least significant to most significant), while the bits within each byte should be
```
zslayton:
```suggestion
The bytes are read from top to bottom (least significant to most significant), while the bits within each byte should be
```
or
```suggestion
The bytes are read from the top down (least significant to most significant), while the bits within each byte should be
```

zslayton (approved):
The diagrams look good! I left some comments around typos and suggested rephrasing.

src/binary-encoding.adoc
almann: I will edit this as:

> boundary between encoding primitives (e.g., opcode/`FlexUInt`)

src/binary-encoding.adoc
```diff
@@ -841,6 +841,10 @@ though it is less compact than opcodes `0x73`-`0x77` above.
 |===
 _{asterisk} Serialized size in bytes does not include the opcode._
 
+The body of short-form timestamps are encoded as a `FixedUInt` of the size specified by the opcode.  This integer is
```
```suggestion
The body of a short-form timestamp is encoded as a `FixedUInt` of the size specified by the opcode.  This integer is
```

src/binary-encoding.adoc
```diff
@@ -1096,27 +1217,49 @@ Unlike the short-form encoding, the long-form encoding reserves:
...
-`(coefficient, exponent)` pair, similar to a <<decimals, decimal>>. However, it is illegal for the fractional
-seconds value to be greater than or equal to `1.0` or less than `0.0`. For this reason, both the coefficient and
-the exponent are encoded using unsigned types. The included coefficient `FlexUInt` is unsigned to prevent the encoding of
-fractional seconds less than `0.0`. The exponent `FixedUInt` is implicitly negative, discouraging the encoding of
-decimal numbers greater than `1.0`. Note that validation is still required; namely:
+Similar to short-form timestamps, with the exception of representing the fractional seconds, the components of the
+timestamp are encoded as bit-fields on a `FixedUInt` that corresponds to the length that followed the opcode.
+
+If the timestamp's overall length is greater than or equal to `8`, the `FixedUInt` part of the timestamp is `8` bytes
+and the remainder of the length is used to encode fractional seconds. The fractional seconds are encoded as a
```
```suggestion
and the remaining bytes are used to encode fractional seconds. The fractional seconds are encoded as a
```

src/binary-encoding.adoc
```diff
@@ -1096,27 +1217,49 @@ Unlike the short-form encoding, the long-form encoding reserves:
...
-If the timestamp's length is greater than or equal to `8`, it has fractional seconds that are encoded as a
-`(coefficient, exponent)` pair, similar to a <<decimals, decimal>>. However, it is illegal for the fractional
-seconds value to be greater than or equal to `1.0` or less than `0.0`. For this reason, both the coefficient and
-the exponent are encoded using unsigned types. The included coefficient `FlexUInt` is unsigned to prevent the encoding of
-fractional seconds less than `0.0`. The exponent `FixedUInt` is implicitly negative, discouraging the encoding of
-decimal numbers greater than `1.0`. Note that validation is still required; namely:
+Similar to short-form timestamps, with the exception of representing the fractional seconds, the components of the
+timestamp are encoded as bit-fields on a `FixedUInt` that corresponds to the length that followed the opcode.
+
+If the timestamp's overall length is greater than or equal to `8`, the `FixedUInt` part of the timestamp is `8` bytes
```
```suggestion
If the timestamp's overall length is greater than or equal to `8`, the `FixedUInt` part of the timestamp is `7` bytes
```

I think this is what you're trying to say? It takes 7 bytes to encode a long-form timestamp through seconds, then the FlexUInt + FixedUInt.

src/binary-encoding.adoc
```diff
@@ -1096,27 +1217,49 @@ Unlike the short-form encoding, the long-form encoding reserves:
...
-----
-     1         2         3         4         5         6         7         8           n
-+=========+=========+=========+=========+=========+=========+=========+========+   +=========+
-|YYYY:YYYY|MMYY:YYYY|hDDD:DDMM|mmmm:HHHH|oooo:oomm|ssoo:oooo|----|ssss|FlexUInt|...|FixedUInt|...
-+=========+=========+=========+=========+=========+=========+=========+========+   +=========+
+.Figure {counter:figure}: Encoding of the _body_ of a long-form timestamp
+[%unbreakable,source]
+----
+         +=========+
+byte 0   |YYYY:YYYY|
```
tgregg, line 1242: In the other diagrams, byte 0 began with the opcode. Should we do that here too, for consistency?

almann: In this one, it was specifically only the body.  I had added #222 before this PR to cover this because when I went through this section I realized that we should have a bunch of examples w/opcodes.

src/binary-encoding.adoc
almann: Great catch!

tgregg (approved):

toddjonker (approved):
I _really_ like this layout, less strain on my brain!

Also @zslayton I love the new opcode layout table, super easy to understand.

PR #224, opened 2023-05-17
Add the grammar for directives, modules, and macros.

toddjonker:
### Note

GitHub preview ignores the table formatting; it shouldn't have any borders or shading.

----

### DIFF
--- a/src/grammar.adoc
+++ b/src/grammar.adoc
@@ -126,7 +126,7 @@ _module-body_ `)`
 |base-shape         |::=| _tagged-type_  \|  _primitive-type_  \|  _macro-ref_
 |simple-cardinality |::=|  `*!*`  \|  `*+*`  \|  `*'?'*`  \|  `*'{asterisk}'*`
 
-|grouped-param-spec  |::=|  `(` _param-name_ `[` _base-shape_ `]` _grouped-cardinality_ `)`
+|grouped-param-spec  |::=|  `(` _param-name_ `[` _base-shape_ `]` _grouped-cardinality_? `)`
 |grouped-cardinality |::=|  `*+*`  \|  `*'{asterisk}'*`
 
 |rest-spec        |::=|  `(` _param-name_ _base-shape_ _rest-cardinality_  `)`
--- a/src/grammar.adoc
+++ b/src/grammar.adoc
@@ -1,0 +1,208 @@
+[[sec:grammar]]
+== Domain Grammar
+
+:nrm: subs="+normal"
+:bnf: cols="10,^2,40",frame=none,grid=none,stripes=none,options="noheader"
+:sp2: {nbsp}{nbsp}{nbsp}{nbsp}
+
+This chapter presents Ion 1.1's _domain grammar_, by which we mean the grammar of the domain
+of values that drive Ion's encoding features.
+
+We use a BNF-like notation for describing various syntactic parts of a document,
+including Ion data structures. In such cases, the BNF should be interpreted loosely to
+accommodate Ion-isms like commas and unconstrained ordering of struct fields.
+
+All `()[]{}` below are literal tokens of Ion syntax.  Single-quoted `*'?'*` and `*'{asterisk}'*`
+denote literal Ion symbols, while unquoted |, ?, and {asterisk} are BNF notation.
+
+
+=== Documents
+
+TODO this section needs much work.
+
+[{bnf}]
+|===
+|document  |::=| _segment_*
+|segment   |::=| _ivm_?  _value_* _directive_?
+|directive |::=| _symtab-directive_ \| _encoding-directive_
+|===
+
+
+=== Encoding Directives
+
+[{bnf}]
+|===
+|encoding-directive |::=| `$ion_encoding::(` _retention_? _module-decl_*
+                                             _symtab_? _top-mactab_? `)`
+
+|retention |::=|  `(` `*retain*` _retainees_ `)`
+|retainees |::=|  `*'{asterisk}'*`  \|  _module-name_*
+
+|module-decl |::=| _dependency_  \|  _inline-module-def_
+|dependency  |::=| _load-decl_  \|  _use-decl_  \|  _import-decl_
+
+|use-decl    |::=|  `(` `*use*` _use-item_* `)`
+|use-item    |::=| _module-name_  \|  _load-decl_
+
+|symtab      |::=|  `(` `*symbol_table*` _symtab-item_* `)`
+|symtab-item |::=|  _module-name_  \|  `[` _text_* `]`
+
+|top-mactab  |::=|  `(` `*macro_table*` _module-name_* )
+
+|===
+
+
+==== Catalog Access
+
+[{bnf}]
+|===
+|load-decl       |::=|  `(` `*load*` _load-body_ `)`
+|import-decl     |::=|  `(` `*import*` _load-body_ `)`
+|load-body       |::=| _module-name_ _catalog-name_ _catalog-version_ _symbol-maxid_?
+
+|catalog-name    |::=|  _unannotated-string_
+|catalog-version |::=|  _unannotated-uint_
+|symbol-maxid    |::=|  _unannotated-uint_
+|===
+
+
+=== Macro References
+
+[{bnf}]
+|===
+|macro-ref     |::=|  _macro-name_  \|  _local-ref_  \|  _qualified-ref_
+
+|local-ref     |::=|  < symbol of the form ``':``_name-or-address_``'`` >
+|qualified-ref |::=|  < symbol of the form ``':``_module-name_``:``_name-or-address_``'`` >
+
+|module-name     |::=|  _unannotated-identifier-symbol_
+|macro-name      |::=|  _unannotated-identifier-symbol_
+|macro-address   |::=|  _unannotated-uint_
+|name-or-address |::=|  _macro-name_  \|  _macro-address_
+|===
+
+
+=== Module Definitions
+
+[{bnf}]
+|===
+|inline-module-def   |::=| `(` `*module*` _module-name_ _module-body_ `)`
+|shared-module-def   |::=| ``$ion_shared_module::``__ion-version-marker__``::(`` _catalog-key_
+_module-body_ `)`
+|catalog-key         |::=| `(` `*catalog_key*` _catalog-name_ _catalog-version_ `)`
+|===
+
+
+==== Module Bodies
+
+[{bnf}]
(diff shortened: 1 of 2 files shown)


src/grammar.adoc
```diff
@@ -0,0 +1,208 @@
...
+
+|special-form |::=|  `(` `*literal*` _datum_ `)`                             +
+                 \|  `(` `*if_void*`   _template_ _template_ _template_ `)`  +
+                 \|  `(` `*if_single*` _template_ _template_ _template_ `)`  +
+                 \|  `(` `*if_many*`   _template_ _template_ _template_ `)`  +
+                 \|  `(` `*when*`   _param-name_ _template_ `)`              +
+                 \|  `(` `*unless*` _param-name_ _template_ `)`              +
+                 \|  `(` `*for*` `[` _for-clause_* `]` _template_ `)`
+
+|for-clause       |::=| `(` _identifier_ _template_ `)`
```
zslayton, line 169: At one time, I think we had suggested that the second expr (here: `_template_`) _had_ to be a variable name. I like that this loosens that rule, because the result is very flexible/powerful.

src/grammar.adoc
```diff
@@ -0,0 +1,208 @@
...
+
+|quasi-literal |::=|  `[` _template_* `]`  \|  `{` _quasi-field_* `}`
+|quasi-field   |::=| _text_ `:` _template_
+
+|special-form |::=|  `(` `*literal*` _datum_ `)`                             +
+                 \|  `(` `*if_void*`   _template_ _template_ _template_ `)`  +
+                 \|  `(` `*if_single*` _template_ _template_ _template_ `)`  +
+                 \|  `(` `*if_many*`   _template_ _template_ _template_ `)`  +
+                 \|  `(` `*when*`   _param-name_ _template_ `)`              +
+                 \|  `(` `*unless*` _param-name_ _template_ `)`              +
```
zslayton: Just my 2c: the names `when` and `unless` are unfamiliar to me. I know they're used for control flow in some languages (Perl springs to mind), but they aren't second nature to me the way `if` is.

src/grammar.adoc
```diff
@@ -0,0 +1,208 @@
...
+
+|simple-param-spec  |::=|  `(` _param-name_  _base-shape_ _simple-cardinality_ `)`
+|base-shape         |::=| _tagged-type_  \|  _primitive-type_  \|  _macro-ref_
+|simple-cardinality |::=|  `*!*`  \|  `*+*`  \|  `*'?'*`  \|  `*'{asterisk}'*`
+
+|grouped-param-spec  |::=|  `(` _param-name_ `[` _base-shape_ `]` _grouped-cardinality_? `)`
+|grouped-cardinality |::=|  `*+*`  \|  `*'{asterisk}'*`
+
+|rest-spec        |::=|  `(` _param-name_ _base-shape_ _rest-cardinality_  `)`
+|rest-cardinality |::=|   `*\...*`  \|  `*\...+*`
```
zslayton, line 132: Checking my understanding: `x [any]*` is allowed, but `x any...*` is not? I would expect `*` to be either optional or forbidden in both grouping modes. (I think I'd prefer forbidden?)

zslayton (approved):

src/grammar.adoc
toddjonker: I've been inches from removing these entirely, they've just been hanging on.  I wouldn't mind removing them until we see whether we have convincing scenarios for them.

src/grammar.adoc
```diff
@@ -0,0 +1,208 @@
...
+
+|===
+
+
+==== Catalog Access
+
+[{bnf}]
+|===
+|load-decl       |::=|  `(` `*load*` _load-body_ `)`
+|import-decl     |::=|  `(` `*import*` _load-body_ `)`
```
almann, line 60: To check my current understanding here, *import* is *load* + *use*, yes?

src/grammar.adoc
```diff
@@ -0,0 +1,208 @@
...
+|load-body       |::=| _module-name_ _catalog-name_ _catalog-version_ _symbol-maxid_?
+
+|catalog-name    |::=|  _unannotated-string_
+|catalog-version |::=|  _unannotated-uint_
```
almann, line 64: At some point we should bike shed over the name of this spelling, quite honestly I think these are just "external name" and "external version."  Not a blocker for this PR, just a term I dislike a lot.

src/grammar.adoc
```diff
@@ -0,0 +1,208 @@
...
+
+|macro-alias |::=|  `(` `*alias*` _macro-name_ _macro-ref_ `)`
+
+|module-mactab   |::=|  `(` `*macro_table*` _macro-or-export_* `)`
+|macro-or-export |::=|  _macro-defn_  \|  _export_
+
+|export      |::=|  `(` `*export*` _export-item_* `)`
+|export-item |::=|  _macro-ref_                                         +
+                \|  `(` `*from*` _module-name_ _name-or-address_* `)`   +
+                \|  `(` `*all_from*` _module-name_ `)`
```
almann: I had thought we had aligned on just having *module-name* as an export item.  Is there any reason to require non-orthogonal syntax for `(all_from module-name)` for this case?

src/grammar.adoc
```diff
@@ -0,0 +1,208 @@
...
+
+|grouped-param-spec  |::=|  `(` _param-name_ `[` _base-shape_ `]` _grouped-cardinality_? `)`
+|grouped-cardinality |::=|  `*+*`  \|  `*'{asterisk}'*`
+
+|rest-spec        |::=|  `(` _param-name_ _base-shape_ _rest-cardinality_  `)`
+|rest-cardinality |::=|   `*\...*`  \|  `*\...+*`
+
+|tagged-type   |::=| _abstract-type_ \| _concrete-type_
+|abstract-type |::=| `*any*` \| `*number*` \| `*exact*` \| `*text*` \| `*lob*` \| `*sequence*`
+|concrete-type |::=|  `*'null'*`  \|  `*bool*`  \|  `*timestamp*`  \|  `*int*`  \|  `*decimal*`
```
almann, line 136: Are *`null`* and *`'null'`* are allowed as syntactic synonyms?  And if *`null`* works, presumably *`null.null`* would work too.

src/grammar.adoc
```diff
@@ -0,0 +1,208 @@
...
+[{bnf}]
+|===
+|template |::=|  _identifier_  \|  _literal_  \|  _quasi-literal_
+             \|  _special-form_  \|  _macro-invocation_
+
+|literal       |::=|  _null_  \|  _bool_  \|  _int_  \|  _float_  \|  _decimal_  \| _timestamp_
+                  \|  _string_  \|  _blob_  \|  _clob_
+
+|quasi-literal |::=|  `[` _template_* `]`  \|  `{` _quasi-field_* `}`
+|quasi-field   |::=| _text_ `:` _template_
```
almann, line 161: From a notation point of view, there is something weird about omitting commas for lists and structs, but not the colon here.  I know it's because we're in a grammar over the data model, but still it is odd to me because we are conflating this in the BNF.

src/grammar.adoc
```diff
@@ -0,0 +1,208 @@
...
+                 \|  `(` `*if_void*`   _template_ _template_ _template_ `)`  +
+                 \|  `(` `*if_single*` _template_ _template_ _template_ `)`  +
+                 \|  `(` `*if_many*`   _template_ _template_ _template_ `)`  +
+                 \|  `(` `*when*`   _param-name_ _template_ `)`              +
+                 \|  `(` `*unless*` _param-name_ _template_ `)`              +
+                 \|  `(` `*for*` `[` _for-clause_* `]` _template_ `)`
+
+|for-clause       |::=| `(` _identifier_ _template_ `)`
+
+|macro-invocation |::=|  `(` _macro-ref_ _macro-arg_* `)`
```
almann, line 171: Technically this conflicts with the special forms in the grammar (since generally BNFs are not ordered, while PEGs are), probably worth noting that *special-form* takes precedence and one needs to use qualification to invoke a macro that shadows a special form within a template.

src/grammar.adoc
```diff
@@ -0,0 +1,208 @@
...
+|name-or-address |::=|  _macro-name_  \|  _macro-address_
+|===
+
+
+=== Module Definitions
+
+[{bnf}]
+|===
+|inline-module-def   |::=| `(` `*module*` _module-name_ _module-body_ `)`
+|shared-module-def   |::=| ``$ion_shared_module::``__ion-version-marker__``::(`` _catalog-key_
```
almann, line 90: I would pull this out into a different section because it is not part of the domain model.  To wit, it is orphaned in the grammar.  At a minimum it should be noted that it is not formally part of the model, but I think it goes in a separate annex of the grammar alone with shared symbol table.

almann (commented):
This is generally looking really good, though I have some questions about some of the constructs below.

src/grammar.adoc
toddjonker: Yes.  

Note that I'm trying to avoid commentary in this chapter, more for reference.  (I tried to do a callout but that's not possible from inside a table)

src/grammar.adoc
toddjonker: I think you're talking about "catalog"?

src/grammar.adoc
toddjonker: Prior revisions allowed `null` but everyone seems unhappy with that.  Here, these are all actual symbols (thus the `'`s), so `'null'` is correct but `null` and `null.null` are erroneous.

src/grammar.adoc
toddjonker: Ah, that should be up at the level of `macro_table` (ie sibling of `macro` and `export`), but I forgot.  Will fix.

src/grammar.adoc
toddjonker: I don't see a better way to express that only the field-value is a template.  Have a suggestion?

src/grammar.adoc
toddjonker: Adding an admonition regarding that.

src/grammar.adoc
toddjonker: Your understanding is correct.  I'm a bit torn on your suggestion, so moving this to a tracking issue #230 for discussion.

src/grammar.adoc
toddjonker: #231

almann (approved):
🚀 looks good, though I am stewing on domain grammar as a term and think we should have something more specific.

PR #235, opened 2023-05-26
Argument encoding bitmap, expression groups

zslayton:
Adds definitions for the Argument Encoding Bitmap (formerly "Presence Bitmap"), length-prefixed and delimited expression groups, and examples for both.

----

### DIFF
--- a/src/binary-encoding.adoc
+++ b/src/binary-encoding.adoc
@@ -2424,8 +2424,8 @@ parameter can have 0, 1, or 2 bits.
 |===
 
 The total number of bits in the AEB can be calculated by analyzing the signature of the macro being invoked.
-If the macro has no parameters or all of its parameters have a cardinality of exactly-one, no bits are required; the
-AEB will be omitted altogether.
+If the macro has no parameters or all of its parameters have a cardinality of either exactly-one or one-or-more,
+no bits are required; the AEB will be omitted altogether.
 If the macro has many parameters with a cardinality other than exactly-one, it is possible for the AEB to require more
 than one byte to encode; in such cases, the bytes are written in little-endian order.
 AEB bytes can contain unused bits.
@@ -2588,8 +2588,8 @@ the end of the sequence of pages.
 │  ┌──── AEB: 0b0000_0011; the arguments for grouped parameter `x` have been encoded
 │  │     as a delimited expression group. Count-prefixed pages of `compact_int`
 │  │     expressions follow.
-│  │   ┌──── Count prefix: FlexUInt 2; 3 `compact_int`s follow.
-│  │   │        ┌──── Count prefix: FlexUInt 1; a single `compact_int`s follow.
+│  │   ┌──── Count prefix: FlexUInt 2; 2 `compact_int`s follow.
+│  │   │        ┌──── Count prefix: FlexUInt 1; a single `compact_int` follows.
 │  │   │        │    ┌──── Count prefix: FlexUInt 0; no more pages follow.
 │  │   │        │    │
 00 03 05 03 05 03 07 01
--- a/src/binary-encoding.adoc
+++ b/src/binary-encoding.adoc
@@ -2298,16 +2298,304 @@ considered "shapes" rather than types because while their encoding is always sta
 produced by their expansion is not. A single macro can produce streams of varying length and containing values of
 different Ion types depending on the arguments provided in the invocation.
 
-TODO: Examples
+See the link:macros-by-example.adoc#macro_shapes[Macro Shapes] section of _Macros by Example_ for more information.
 
 === Encoding E-expressions With Multiple Arguments
 
-TODO
+If a macro has more than one parameter, the E-expression arguments corresponding to each parameter are encoded one
+after the other moving from left to right.
 
-=== Grouped Parameter Encodings
+.Figure {counter:figure}: Definition of macro `_foo_` at address `_0_`
+[%unbreakable,source]
+----
+(macro foo             // Macro name
+  [                    // Parameters
+    (a string!),
+    (b compact_symbol!),
+    (c uint16!)
+  ]
+  /* ... */            // Body (elided)
+)
+----
 
-TODO
+.Figure {counter:figure}: Encoding of E-expression for macro with multiple parameters: `_(:0 "hello" baz 512)_`
+[%unbreakable,source]
+----
+┌──── The opcode is less than 0x40, so it is an E-expression invoking the macro at
+│     address 0, `foo`. `foo`'s first parameter is a string, so an opcode follows.
+│
+│  ┌──── Opcode 0x85 indicates a 5-byte string. 5 UTF-8 bytes follow.
+│  │
+│  │                 ┌──── `foo`'s second parameter is a compact_symbol, so a `FlexSym` follows.
+│  │                 │     FlexSym -3: 3 bytes of UTF-8 text follow.
+│  │                 │
+│  │                 │           ┌──── `foo`'s third parameter is a uint16, so a 2-byte
+│  │                 │           │     2-byte `FixedUInt` follows.
+│  │                 │           │     FixedUInt: 512
+│  │  h  e  l  l  o  │   b  a  z │
+00 85 68 65 6C 6C 6F FD 62 61 7A 00 20
+      └──────┬─────┘    └───┬──┘
+        UTF-8 bytes    UTF-8 bytes
+----
 
-=== Voidable Parameter Encodings
+[[argument_encoding_bitmap]]
+=== Argument Encoding Bitmap (AEB)
 
-TODO
+The examples in previous sections have only shown how to encode invocations of macros which have either no parameters
+at all (aka _constants_) or whose parameters all have
+link:macros-by-example.adoc#exactly_one[a cardinality of exactly-one].
+
+If a macro has any parameters with a cardinality of zero-or-one (`?`), zero-or-more (`*`), or one-or-more (`+`),
+then E-expressions invoking that macro will begin with an _argument encoding bitmap_ (AEB).
+An AEB is a series of bits that correspond to a macro parameter and communicate additional
+information about how the arguments corresponding to that parameter have been encoded in the current E-expression.
+In particular, the AEB indicates whether a parameter that accepts `(:void)` has any arguments at all, and how a
+grouped parameter's arguments have been delimited.
+
+The number of bits allotted to each parameter is determined by its cardinality, as shown in the table below; each
+parameter can have 0, 1, or 2 bits.
+
+[cols="^.^2a,^.^2a,^.^2a,^.^1a,^.^1a,.^6a"]
+|===
+|Grouping Mode|Cardinality|Example parameter signature|Number of bits|Bit(s) value|Encoding
+
+.7+|Ungrouped
+|Exactly-one
+|`(x int!)`
+|0
+|_n/a_
+|One expression
+
+.2+|Zero-or-one
+.2+|`(x int?)`
+.6+|1
+|`0`
+<|No expression; equivalent to `(:void)`
+
+|`1`
+<|One expression
+
+.2+|Zero-or-more
+.2+|`(x int*)`
+|`0`
+<|No expression; equivalent to `(:void)`
+
+|`1`
+<|One expression
+
+.2+|One-or-more
+.2+|`(x int+)`
+|`0`
+<|No expression; equivalent to `(:void)`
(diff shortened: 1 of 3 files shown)


src/binary-encoding.adoc
```diff
@@ -2298,16 +2298,302 @@ considered "shapes" rather than types because while their encoding is always sta
...
+|`10`
+<|<<length_prefixed_expression_groups, Length-prefixed expression group>>
+
+|`11`
+<|<<delimited_expression_groups, Delimited expression group>>
+
+|===
+
+The total number of bits in the AEB can be calculated by analyzing the signature of the macro being invoked.
+If the macro has no parameters or all of its parameters have a cardinality of exactly-one, no bits are required; the
```
tgregg: Or one-or-more?

src/binary-encoding.adoc
```diff
@@ -2298,16 +2298,302 @@ considered "shapes" rather than types because while their encoding is always sta
...
+.Figure {counter:figure}: Delimited encoding of `_(:compact_foo [1, 2, 3])_` using two pages
+[%unbreakable,source]
+----
+┌──── The opcode is less than 0x40, so it is an E-expression invoking the macro at
+│     address 0: `foo`. `foo` takes a group of int expressions as a parameter (`x`),
+│     so an argument encoding bitmap (AEB) follows.
+│  ┌──── AEB: 0b0000_0011; the arguments for grouped parameter `x` have been encoded
+│  │     as a delimited expression group. Count-prefixed pages of `compact_int`
+│  │     expressions follow.
+│  │   ┌──── Count prefix: FlexUInt 2; 3 `compact_int`s follow.
```
```suggestion
│  │   ┌──── Count prefix: FlexUInt 2; 2 `compact_int`s follow.
```

src/binary-encoding.adoc
```diff
@@ -2298,16 +2298,302 @@ considered "shapes" rather than types because while their encoding is always sta
...
+│  │   │        ┌──── Count prefix: FlexUInt 1; a single `compact_int`s follow.
```
```suggestion
│  │   │        ┌──── Count prefix: FlexUInt 1; a single `compact_int` follows.
```

tgregg (approved):

src/binary-encoding.adoc
```diff
@@ -2298,16 +2298,302 @@ considered "shapes" rather than types because while their encoding is always sta
...
 different Ion types depending on the arguments provided in the invocation.
 
-TODO: Examples
+See the link:macros-by-example.adoc#macro_shapes[Macro Shapes] section of _Macros by Example_ for more information.
 
 === Encoding E-expressions With Multiple Arguments
 
-TODO
+If a macro has more than one parameter, the E-expression arguments corresponding to each parameter are encoded one
+after the other moving from left to right.
```
toddjonker: I don't think the if-clause adds any clarity.

```suggestion
E-expression arguments corresponding to each parameter are encoded one
after the other moving from left to right.
```

src/binary-encoding.adoc
```diff
@@ -2298,16 +2298,302 @@ considered "shapes" rather than types because while their encoding is always sta
...
 
-=== Grouped Parameter Encodings
+.Figure {counter:figure}: Definition of macro `_foo_` at address `_0_`
+[%unbreakable,source]
+----
+(macro foo             // Macro name
+  [                    // Parameters
+    (a string!),
+    (b compact_symbol!),
```
toddjonker, line 2313: Noting different terminology here versus grammar's `var_symbol`.  I do think `compact_symbol` is better, though a bit long.  I cut issue #236 to track.

src/binary-encoding.adoc
```diff
@@ -2298,16 +2298,302 @@ considered "shapes" rather than types because while their encoding is always sta
...
+│  │                 │           │     2-byte `FixedUInt` follows.
+│  │                 │           │     FixedUInt: 512
+│  │  h  e  l  l  o  │   b  a  z │
+00 85 68 65 6C 6C 6F FD 62 61 7A 00 20
+      └──────┬─────┘    └───┬──┘
+        UTF-8 bytes    UTF-8 bytes
+----
+
+[[argument_encoding_bitmap]]
+=== Argument Encoding Bitmap (AEB)
```
toddjonker, line 2341: [nitpick] This no longer feels like a "bitmap", but I don't have a better term to offer ATM.

src/binary-encoding.adoc
```diff
@@ -2298,16 +2298,302 @@ considered "shapes" rather than types because while their encoding is always sta
...
+|One-or-more
+|`(x int+)`
+|0
+|_n/a_
+<|One expression
+
+.8+|Grouped
+.4+|Zero-or-more
+.4+|`(x [int])` +
+`(x int...)`
```
toddjonker:
```suggestion
`(x int\...)`
```

Keeps Adoc from turning three periods into single ellipses character.

src/binary-encoding.adoc
```diff
@@ -2298,16 +2298,302 @@ considered "shapes" rather than types because while their encoding is always sta
...
+
+|`10`
+<|Length-prefixed expression group
+
+|`11`
+<|Delimited expression group
+
+.4+|One-or-more
+.4+|`(x [int]+)` +
+`(x int...+)`
```
toddjonker:
```suggestion
`(x int\...+)`
```

src/binary-encoding.adoc
```diff
@@ -2298,16 +2298,302 @@ considered "shapes" rather than types because while their encoding is always sta
...
+|`1`
+<|One expression
+
+.2+|Zero-or-more
+.2+|`(x int*)`
+|`0`
+<|No expression; equivalent to `(:void)`
+
+|`1`
+<|One expression
```
toddjonker, line 2383: I think we've (or I've) forgotten either a grammar constraint or an encoding solution, b/c I don't know how to encode a non-unary tagless parameter like `(x uint8+)`, since their "one expression" can't be anything but one value.

src/binary-encoding.adoc
```diff
@@ -2298,16 +2298,302 @@ considered "shapes" rather than types because while their encoding is always sta
...
+| `\((a int*) (b [string]) (c [float]))`
+| `---ccbba`
+| 5
+
+| `\((a [int]) (b [string]) (c [float]+))`
+| `--ccbbaa`
+| 6
+
+| `\((a int*) (b [string]) (c [float]) (d [bool]) (e blob...))`
+| `eddccbba` +
```
toddjonker, line 2472: At one point we had discussed intermingling these bits with the argument expressions, to avoid an expensive full-expression prefix.   I am still worried about the cost of having to compute all of this before rendering anything at all.

src/binary-encoding.adoc
```diff
@@ -2298,16 +2298,302 @@ considered "shapes" rather than types because while their encoding is always sta
...
+| `\((a int!) (b string?) (c float!))`
+| `-------b`
+| 1
+
+| `\((a int!) (b string*) (c float?))`
+| `------cb`
+| 2
+
+| `\((a int*) (b string!) (c [float]))`
+| `-----cca`
```
toddjonker, line 2460: i think there's still room for confusion here on the order of the `c` bits.

toddjonker:
Now that we have a lot more prose committed, does anyone still have concerns about the terms "tagged" and "tagless"?  It's still working well for me, but I know others had concerns about them.

src/binary-encoding.adoc
```diff
@@ -2298,16 +2298,302 @@ considered "shapes" rather than types because while their encoding is always sta
...
+----
+
+===== Delimited Tagless Expression Groups
+
+Tagless type encodings do not have an opcode, and so cannot use the closing delimiter opcode--`0xF0` is a valid first
+byte for many tagless encodings.
+
+Instead, tagless expressions are grouped into 'pages', each of which is prefixed by a <<flexuint, `FlexUInt`>>
+representing a count (not a byte-length) of the expressions that follow. If a prefix has a count of zero, that marks
+the end of the sequence of pages.
```
toddjonker, line 2552: I'm really quite happy with this approach!  It's super flexible.   The only strong critique I have is that it has high overhead for small sequences of narrow values like `int8`.   But I suppose in those case length-prefixing would be preferred.

toddjonker (commented):
Nice work!  Well-organized and easy to read.

src/binary-encoding.adoc
zslayton: It was intended to contrast with previous sections that discussed the encoding of E-expressions with a single parameter. I agree that the latter clause stands alone, though. Updated.

src/binary-encoding.adoc
zslayton: Earlier versions of this tried "configuration," "specification" ("spec"), "header," and "prefix." I'll leave it as-is for the moment and we can change it when we think of something better.

src/binary-encoding.adoc
zslayton: Yeah, at the moment `uint8+` is not forbidden, but it's identical to `uint8` in both encoding and "expansion." We should probably forbid it to avoid confusion. Assuming there's consensus on that, I'll make the change in a future revision.

zslayton: Opened #238 to track this.

src/binary-encoding.adoc
zslayton: The text currently says:
> Bits are assigned from least significant to most significant (commonly: right-to-left).

and the table shows the meaning of the bits in the same order --v

image

What do you think would help to clarify it further?

src/binary-encoding.adoc
zslayton: One of our most recent discussions with Chris led us to the conclusion that macros with enough non-unary parameters to encounter that issue would be so rare in practice that it wasn't worth the complexity of splitting up the prefix.

src/binary-encoding.adoc
zslayton: >  The only strong critique I have is that it has high overhead for small sequences of narrow values like int8. But I suppose in those case length-prefixing would be preferred.

Yes, I think that's true. Writers also have the option of doing a delimited sequence comprised of very few pages (1?) with a large count. The data size difference is likely negligible, though.

src/binary-encoding.adoc
zslayton: Expanding on this a bit:
* The width of the AEB is statically known, so bytes can be reserved in the encoding buffer without needing to adjust the size later.
* 8 bytes of AEB can be stored in an integer on 64-bit architectures, which means that doing all of the bit-twiddling needed to configure the AEB is fast and easy unless you have more than 32 grouped parameters (or an even greater number of less-expensive parameters).
* Keeping the AEB bytes together means you don't have to worry about splitting two bits across bytes that are potentially very far apart in memory.

zslayton:
@toddjonker said:
> Now that we have a lot more prose committed, does anyone still have concerns about the terms "tagged" and "tagless"? It's still working well for me, but I know others had concerns about them.

They work well for me too. We need to talk about the dichotomy of has-opcode and no-opcode so frequently in the binary encoding space that if we didn't use those we'd have to pick another very similar name pair. Part of me wonders if the best way to get consistency is to just rename 'opcode' to 'tag', though 'opcode' is certainly more descriptive.

src/binary-encoding.adoc
toddjonker: To clarify, the "cost" I'm concerned about is buffering, not bit-twiddling.  This has similar downside to length-prefixing, in that the entire content of the E-expression (recursively!) needs to be analyzed in order to determine what to write, and we can't write any bytes until that's done.  Put another way, a lot of detailed decisions need to be made before we can write that, and it's only bounded by the width of the entire eexpr.   Given the number of encoding options, I expect that in practice implementations well end up buffering the entire thing.  🤷 

I think that "rare in practice" was predicated on one bit per parameter, but here we can spill-over to a second prefix-byte in as few as 5 grouped parameters.

Just something we should keep an eye on as we move to implementation and real-world uses.

src/binary-encoding.adoc
toddjonker: This is perhaps mostly my brain struggling with the little-endianness.  I think we need to specify up-front and be consistent that where we express sequences of 0|1 bits they are big-endian, and where we write sequences of (individually big-endian) octets they stack vertically.   Or something like that.

toddjonker: Perhaps a `0b` prefix to indicate the big-endian bit sequences

toddjonker (approved):

Issue #237, opened 2023-05-26
Allow ungrouped, non-unary, tagless parameters?


toddjonker:
> I think we've (or I've) forgotten either a grammar constraint or an encoding solution, b/c I don't know how to encode a non-unary tagless parameter like `(x uint8+)`, since their "one expression" can't be anything but one value.

_Originally posted by @toddjonker in https://github.com/amazon-ion/ion-docs/pull/235#discussion_r1207484609_

toddjonker:
To be more clear:
* I think `+` and `*` with ungrouped tagless types, including macros, are problematic since we don't have a way to actually encode an expression returning more than one value.
* I think `?` is okay, since we _can_ represent the absence of value: we have a presence bit in binary, and text can use `(:)` without ambiguity.

toddjonker:
This has been resolved across various changes, most recently with argument groups in #302 and the use of both presence and grouping bits in binary E-expressions.

Issue #230, opened 2023-05-19
Discuss syntax for `[]*` and `...*`


toddjonker:
Checking my understanding: `x [any]*` is allowed, but `x any...*` is not? I would expect `*` to be either optional or forbidden in both grouping modes. (I think I'd prefer forbidden?)

_Originally posted by @zslayton in https://github.com/amazon-ion/ion-docs/pull/224#discussion_r1197894068_

popematt:
@toddjonker I think that #308 and #302 have obviated the need for this issue. Would you agree?

toddjonker:
Agreed; closing.

PR #240, opened 2023-06-01
Add more defaulting to macro parameter declarations.

toddjonker:
* The type is optional for both simple and grouped params (defaults to `any`).
* The cardinality is optional for simple params (defaults to `!`).
* Simple tagless params (including macros) can only use `?`, since there's no way to encode more than one value.
* Grouped params can only be modified by `+`.

Changes to the "by Example" chapters are consistent with the grammar, and generally drop use of `!`.  At this point there's no examples using optional-any.

Addresses #230 and #237

### Notes:

This is a draft response to recent suggestions. It doesn't go as far as @zslayton has proposed but leaves room to consider that separately.

I'm becoming inclined to ditch the `!` modifier altogether.  It's only used in conjuction with simple tagged params, it'll probably never be used, and its presence seems to highlight the different default for `[]`.   (I added it for narrative reasons that are now less impactful in the face of grouped parameters.)

----

### DIFF
--- a/src/grammar.adoc
+++ b/src/grammar.adoc
@@ -118,20 +118,24 @@ _module-body_ `)`
 
 |signature   |::=|  _param-specs_ _result-spec_?
 |param-specs |::=|  `(` _param-spec_* _rest-spec_? `)`  \|  `[` _param-spec_* _rest-spec_? `]`
-|param-spec  |::=|  _param-name_  \|  _simple-param-spec_  \|  _grouped-param-spec_
+|param-spec  |::=|  _param-name_  \|  `(` _param-name_ _param-shape_ `)`
 |param-name  |::=|  _unannotated-identifier-symbol_
 
-|simple-param-spec  |::=|  `(` _param-name_  _base-shape_ _simple-cardinality_ `)`
-|base-shape         |::=| _tagged-type_  \|  _primitive-type_  \|  _macro-ref_
-|simple-cardinality |::=|  `*!*`  \|  `*+*`  \|  `*'?'*`  \|  `*'{asterisk}'*`
+|param-shape        |::=|  _simple-shape_  \|  _grouped-shape_
 
-|grouped-param-spec  |::=|  `(` _param-name_ `[` _base-shape_ `]` _grouped-cardinality_? `)`
-|grouped-cardinality |::=|  `*+*`  \|  `*'{asterisk}'*`
+|simple-shape       |::=|  _tagless-type_ `*?*`?
+                       \|  _tagged-type_? _any-cardinality_?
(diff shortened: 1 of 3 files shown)


zslayton (commented):
> Simple tagless params (including macros) can only use ?, since there's no way to encode more than one value.

I may be misunderstanding this, but I think we need to allow `*` and `+` for macro shapes:

```
(macro foo ()
  (values 1 2 3))

(macro bar [(f foo)]
  // ... body elided
)

(:bar ()) // Error: `f` expected 1 value, but received 3
```

> I'm becoming inclined to ditch the ! modifier altogether. It's only used in conjuction with simple tagged params, it'll probably never be used, and its presence seems to highlight the different default for []. (I added it for narrative reasons that are now less impactful in the face of grouped parameters.)

👍

toddjonker:
> I may be misunderstanding this, but I think we need to allow `*` and `+` for macro shapes:
> 
> ```
> (macro foo ()
>   (values 1 2 3))
> 
> (macro bar [(f foo)]
>   // ... body elided
> )
> 
> (:bar ()) // Error: `f` expected 1 value, but received 3
> ```

Macro types are significantly different from the other types. Assuming we add return-type specifiers (it's been accepted in concept but not written up), the modifier here is effectively double-bookkeeping that information (AFAICT). So I fret that the modifier will be frequently forgotten (the author knows that `foo` returns multiple values), to later surprise. I fret that `foo*` would be misinterpreted by non-authors to mean multiple `foo`s are accepted.

zslayton:
> Macro types are significantly different from the other types. Assuming we add return-type specifiers (it's been accepted in concept but not written up), the modifier here is effectively double-bookkeeping that information (AFAICT). So I fret that the modifier will be frequently forgotten (the author knows that foo returns multiple values), to later surprise. I fret that foo* would be misinterpreted by non-authors to mean multiple foos are accepted.

I'm ok with this in principle, but I think we need to be more explicit about that special behavior. Specifically, I think we need a section describing (or a TODO) how macro shapes with non-unary expansions work. Otherwise this:

> Simple tagless params (including macros) can only use ?, since there's no way to encode more than one value.

is misleading as a macro _can_ produce more than one value. The situation with `?` is potentially confusing too; a macro can expand to `-> int?`--do I still put `?` on the param signature?

toddjonker:
I'll cut an #250 to track your feedback.  It's a good point, I know we've circled this issue countless times, but it we've not quite nailed it.

> > Simple tagless params (including macros) can only use ?, since there's no way to encode more than one value.
> 
> is misleading as a macro _can_ produce more than one value.

Agreed, that's a bug in the commit-log; it should read "more than one argument", perhaps.

toddjonker:
I took WIP out of the title, I think this is merge-able (though I'd like to hear @almann thoughts).

zslayton (approved):

almann (approved):
LGTM

PR #248, opened 2023-06-02
Add glossary entry for "quasi-literal".

toddjonker:
Also fixed some dumb-quotes.
----

### DIFF
--- a/src/glossary.adoc
+++ b/src/glossary.adoc
@@ -116,6 +116,7 @@ A template, denoted as a list or struct, that is _partly_ (“quasi-”) literal
 List-shaped templates treat the elements as nested templates.
 Struct-shaped templates treat the field names as literal, but the corresponding values as
 templates.
+S-expressions denote operator invocations and are not treated quasi-literally.
 
 rest parameter::
 A macro parameter—always the final parameter—declared with the `*\...*` or `*\...+*` modifier,
--- a/src/glossary.adoc
+++ b/src/glossary.adoc
@@ -10,7 +10,7 @@ A macro can be _fixed arity_ or _variable arity_.
 
 argument::
 A single sub-expression within a macro invocation, corresponding to one of the macro’s
-parameters.  This is by default a one-to-one relation, but a parameter's grouping mode
+parameters.  This is by default a one-to-one relation, but a parameter’s grouping mode
 can change this to a many-to-one relation.
 
 argument group::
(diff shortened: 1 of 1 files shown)


src/glossary.adoc
```diff
@@ -111,6 +111,12 @@ A macro reference that consists of a module name and either a macro name exporte
 or a numeric address within the range of the module’s exported macro table. In text, these look
 like `:__module-name__:__name-or-address__`.
 
+quasi-literal::
+A template, denoted as a list or struct, that is _partly_ (“quasi-”) literal.
+List-shaped templates treat the elements as nested templates.
+Struct-shaped templates treat the field names as literal, but the corresponding values as
+templates.
```
zslayton, line 118: I suggest adding a sentence mentioning why S-expressions are _not_ quasi-literals; listing two out of the three container types naturally raises the question.

src/glossary.adoc
```diff
@@ -77,7 +77,7 @@ See also _grouping mode_.
 grouping mode::
 One of three ways that a macro parameter is given arguments. A _simple parameter_ accepts one
 argument, a _grouped parameter_ accepts a list of arguments, and a _rest parameter_ accepts
-"all the rest" of the trailing arguments without a grouping list.
+“all the rest” of the trailing arguments without a grouping list.
```
zslayton, line 80: RE: dumb quotes, I believe my text editor has been using them by default throughout. Is there a reason you prefer smart quotes? Does your editor have a setting to enable them? I'm ok with either so long as we're consistent, though I'm not looking forward to the find-and-replace I'll need to do to update things.

This is not a blocking concern.

toddjonker: I think straight-quotes look amateurish in the context of the PDF document.

AsciiDoc has rather awkward syntax for curly quotes: "\`word`"
I did these manually, since I can't figure out how to get IDEA to do them automatically. 

We have several typographic inconsistencies to remedy, but I've been inclined to deal with it later.  I'm opinionated, so perhaps I should propose a basic style guide?

src/glossary.adoc
toddjonker: Added.

src/glossary.adoc
zslayton: > We have several typographic inconsistencies to remedy, but I've been inclined to deal with it later. 

Yeah, I think that's wise. I'd rather do it in as few passes as possible once we have content we're happy with. (O, to have an autoformatter!)

> I'm opinionated, so perhaps I should propose a basic style guide?

Sure, that'd be good.

zslayton (approved):

PR #249, opened 2023-06-07
Add chapter on the system module and its macros.

toddjonker:
The bottom chunk of this is rough, but want to get the shape in place for further iteration.

----

### DIFF
--- a/src/system-module.adoc
+++ b/src/system-module.adoc
@@ -6,8 +6,7 @@ include::styles.adoc[]
 The symbols and macros of the system module `$ion` are available everywhere within an Ion document,
 with the version of that module being determined by the spec-version of each segment.
 
-The specific system symbols are largely uninteresting to users since they don't affect the text
-content that they interact with.
+The specific system symbols are largely uninteresting to users; while the binary encoding heavily leverages the system symbol table, the text encoding that users typically interact with does not.
 The system macros are more visible, especially to authors of macros.
 
 This chapter catalogs the system-provided macros.
--- a/src/system-module.adoc
+++ b/src/system-module.adoc
@@ -1,0 +1,257 @@
+[[sec:sysmod]]
+== The System Module
+
+include::styles.adoc[]
+
+The symbols and macros of the system module `$ion` is available everywhere within an Ion document,
+with the version of that module being determined by the spec-version of each segment.
+
+The specific system symbols are largely uninteresting to users since they don't affect the text
+content that they interact with.
+The system macros are more visible, especially to authors of macros.
+
+This chapter catalogs the system-provided macros.
+The examples below use unqualified names, which works assuming no other module exports the same
+name, but the unambiguous form `:$ion:__macro-name__` is always correct.
+
+
+IMPORTANT: This list is not complete. We expect it to grow and evolve as we gain experience
+writing macros.
+
+
+=== Primitive Operators
+
+This section describes operators that cannot be defined as macros.
+
+
+==== Stream Constructors
+
+===== `void`
+
+[{nrm}]
+----
+(void) \-> any?
+----
+
+Produces an empty stream.
+The most common use of this operator is to supply “no value” to a voidable parameter.
+To make such use more readable, the special-case E-expression `(:)` is synonymous to `(:void)`.
+
+
+===== `values`
+
+[{nrm}]
+----
+(values (v any\...)) \-> any*
+----
+
+Produces a stream from any number of arguments, concatenating the streams produced by the nested
+expressions.
+Used to aggregate multiple values or sub-streams to pass to a single argument, or to return
+multiple results. Generally only useful with more than one subexpression.
+
+
+==== Value Constructors
+
+===== `make_string`
+
+[{nrm}]
+----
+(make_string (content text\...)) \-> string
+----
+
+Produces a non-null, unannotated string containing the concatenated content produced by the
+arguments. Nulls and annotations are discarded.
+
+TODO Probably useful to allow some other Ion scalars (at least) to allow type conversion.
+I think this would be most useful for ints, since the binary representation is more compact than
+as characters. Lobs wouldn’t work well, though.
+
+
+===== `make_symbol`
+
+[{nrm}]
+----
+(make_symbol (content text\...)) \-> symbol
+----
+
+Like `make_string` but produces a symbol.
+
+
+===== `make_list`
+
+[{nrm}]
+----
+(make_list (vals any\...)) \-> list
+----
+
+Produces a non-null, unannotated list from any number of inputs.
+Template expressions of the form `[E~1~, …, E~n~]` are equivalent to `(make_list E~1~ … E~n~)`.
+
+
+===== `make_sexp`
+
+[{nrm}]
+----
+(make_sexp (vals any\...)) \-> sexp
+----
+
+Like make_list but produces a sexp.
+
+[{nrm}]
+----
+(:make_sexp)      ⇒ ()
+(:make_sexp null) ⇒ (null)
+----
+
+
+===== `make_struct`
+
+[{nrm}]
+----
+(make_struct (kv any\...)) \-> struct
+----
+
+Produces a non-null, unannotated struct from any number of elements.
+The ``kv``s are processed in order, incrementally adding fields to an initially-empty struct.
+Various forms of ``kv``s are allowed:
+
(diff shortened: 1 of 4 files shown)


src/system-module.adoc
```diff
@@ -0,0 +1,257 @@
+[[sec:sysmod]]
+== The System Module
+
+include::styles.adoc[]
+
+The symbols and macros of the system module `$ion` is available everywhere within an Ion document,
```
zslayton:
```suggestion
The symbols and macros of the system module `$ion` are available everywhere within an Ion document,
```

src/system-module.adoc
```diff
@@ -0,0 +1,257 @@
...
+with the version of that module being determined by the spec-version of each segment.
+
+The specific system symbols are largely uninteresting to users since they don't affect the text
+content that they interact with.
```
zslayton: Just as an aside, I've seen `name` (`$4`) in the wild quite a bit.

src/system-module.adoc
```diff
@@ -0,0 +1,257 @@
...
+----
+(make_string (content text\...)) \-> string
+----
+
+Produces a non-null, unannotated string containing the concatenated content produced by the
+arguments. Nulls and annotations are discarded.
+
+TODO Probably useful to allow some other Ion scalars (at least) to allow type conversion.
+I think this would be most useful for ints, since the binary representation is more compact than
+as characters. Lobs wouldn’t work well, though.
```
zslayton: > I think this would be most useful for ints, since the binary representation is more compact than
as characters.

I'm not sure I followed this. `(make_string 500)` would be going in the wrong direction, wouldn't it?

src/system-module.adoc
```diff
@@ -0,0 +1,257 @@
...
+
+
+===== `make_sexp`
+
+[{nrm}]
+----
+(make_sexp (vals any\...)) \-> sexp
+----
+
+Like make_list but produces a sexp.
```
zslayton: Possibly worth mentioning that this (and `literal`) are how one makes a `sexp` in the TDL?

src/system-module.adoc
```diff
@@ -0,0 +1,257 @@
...
+----
+(make_struct (kv any\...)) \-> struct
+----
+
+Produces a non-null, unannotated struct from any number of elements.
+The ``kv``s are processed in order, incrementally adding fields to an initially-empty struct.
+Various forms of ``kv``s are allowed:
+
+* A (non-null) string or symbol is treated as a field name, and MUST be followed by another value
+to comprise a key-value pair in the result. Annotations on the field name are discarded.
```
zslayton, line 122: Just checking, do we ignore annotations in other places too (vs error)? e.g. in the arguments for `make_string` and `annotate`?

src/system-module.adoc
```diff
@@ -0,0 +1,257 @@
...
+* A (non-null) struct is merged into the result as-is, after discarding annotations.
+* Any other type of value evokes an expansion error.
+
+Template expressions of the form `{T~1~:E~1~, …, T~n~:E~n~}` are equivalent to
+`(make_struct (*literal* T~1~) E~1~ … (*literal* T~n~) E~n~)`, assuming that no expression E
+produces more than one value.
+In that case, the `make_struct` variant would misbehave: the second value produced by E would be
+treated as the next key.
```
zslayton, line 130: An example (or TODO for one) would probably help here.

src/system-module.adoc
```diff
@@ -0,0 +1,257 @@
...
+
+
+===== `make_decimal`
+
+[{nrm}]
+----
+(make_decimal (coefficient int) (exponent int)) \-> decimal
+----
+
+Since decimal is already compact, this is perhaps most useful in conjunction with packed arrays.
```
zslayton: Also when the exponent is often the same and can be baked into a macro.

src/system-module.adoc
```diff
@@ -0,0 +1,257 @@
...
+===== `make_float` ❓
+
+[{nrm}]
+----
+(make_float ieee) \-> float
+----
+
+Included for completeness, but of unclear utility.
+
+TODO coerce an int or decimal to float?  Perhaps useful to use fixed-width ints to encode various float widths? This may not be useable to convert “IEEE bits” to float, since they would be converted to int before arriving here.
```
zslayton: I've always thought it would be for converting `int` to a `float`, especially small ones. (Maybe `decimal`? That one feels less clear.)

src/system-module.adoc
```diff
@@ -0,0 +1,257 @@
...
+
+
+===== `make_timestamp`
+
+[{nrm}]
+----
+(make_timestamp
+  (year int) (month int) (day int?)
```
zslayton: Month is `?` too, right?

src/system-module.adoc
```diff
@@ -0,0 +1,257 @@
...
+
+===== `annotate`
+
+[{nrm}]
+----
+(annotate (ann [text]*) value) \-> any
+----
+
+Produces the `value` prefixed with the annotations ``ann``s.
+Each `ann` must be a non-null, unannotated string or symbol.
```
zslayton, line 213: Here we require 'unannotated' while we ignore annotations elsewhere.

src/system-module.adoc
```diff
@@ -0,0 +1,257 @@
...
+
+
+===== Local Symtab Appending
+
+[{nrm}]
+----
+(*macro* lst_append
+  \((symbols string\...))
+  (*if_void* symbols
+    (void)                  // Produce nothing if no symbols provided.
```
zslayton, line 261: Nice.

src/system-module.adoc
toddjonker: I've cut #252 to track this (and will link from the doc)

src/system-module.adoc
toddjonker: yep 😣

src/system-module.adoc
toddjonker: Good call-out.  Now that you mention it, I feel icky about ignoring annotations.  That seems likely to lead to users of the macro being surprised by data-loss.

We could use a written-down design principle on this sort of thing, as well as related topics like handling of nulls (typed, untyped, and mis-typed).  #254

src/system-module.adoc
toddjonker: It's a bit inconsistent. #254

src/system-module.adoc
toddjonker: Let's see... in binary the macro variant takes 2 bytes for the system-macro invocation, 3 bytes for the integer.  Versus 4(?) bytes for the 3-character string.  So yeah, not a win for small numbers, but larger numbers should have an increasing advantage, no?

That said, what I had in mind was more along the lines of inserting a variable number into a fixed string, along the lines of the traditional ASIN to URL macro.

#255

src/system-module.adoc
toddjonker: Do you mean users have written `$4` rather than `name`?

Would you like more or different text here?  I admit its not the most clear sentence around...

src/system-module.adoc
zslayton: > Let's see... in binary the macro variant takes 2 bytes for the system-macro invocation, 3 bytes for the integer. Versus 4(?) bytes for the 3-character string. So yeah, not a win for small numbers, but larger numbers should have an increasing advantage, no?

I see, so anywhere you would have written `"7000000"` you could instead invoke a macro that was equivalent to `(make_string 7000000)` to save space. Yep, that checks out.

> That said, what I had in mind was more along the lines of inserting a variable number into a fixed string, along the lines of the traditional ASIN to URL macro.

This is a compelling use case!

src/system-module.adoc
zslayton: > Do you mean users have written $4 rather than name?

No, I just mean that `name` appears as a field name frequently and so gets encoded as `$4`. It doesn't change the thrust of what you're saying, it's just the most common way for users to interact with the system symbol table apart from LSTs.

> Would you like more or different text here? I admit its not the most clear sentence around...

Maybe:
```suggestion
The specific system symbols are largely uninteresting to users; while the binary encoding heavily leverages the system symbol table, the text encoding that users typically interact with does not.
```
?

zslayton (approved):

src/system-module.adoc
toddjonker: That's great! Thanks!

PR #258, opened 2023-06-08
Add Encoding Directives chapter.

toddjonker:
### Description of changes:

This is again a fairly-direct import of our internal drafts. While it should be consistent with current consensus, it still has gaps to be filled.

----

### DIFF
--- a/src/directives.adoc
+++ b/src/directives.adoc
@@ -1,0 +1,308 @@
+[[sec:directives]]
+== Encoding Directives
+
+include::styles.adoc[]
+
+[NOTE]
+.TODO
+====
+This intro section is probably misplaced in the context of the larger book.
+Move or integrate elsewhere.
+====
+
+Ion 1.0 uses symbol tables to capture and compress repeated symbol text.
+At all points in an Ion document, there exists an _encoding environment_ that contains the _current
+symbol table_ mapping symbol IDs to text.
+The encoding environment of a document is controlled by directives embedded in the document.
+These directives are encoding artifacts and not part of the application data model.
+Ion 1.0 has two of directive forms:
+
+* An Ion Version Marker (IVM) resets the environment to the default provided by that version of
+Ion.
+* An `$ion_symbol_table` struct defines a new environment that takes effect immediately after the
+struct closes.
+
+The latter form includes a feature that allows the new environment to be specified in terms of
+the current one in a limited fashion: the current symbol table can be imported as if it were
+shared, so that new symbols can be appended to it.
+
+To increase compression across many documents that have similar content (for example, they use the
+same schema), Ion 1.0 has _shared symbol tables_ that capture a portion of an encoding context—a
+list of symbols—that can be imported into many local symbol tables.
+
+Ion 1.1 generalizes and refactors these features:
+
+* _Macros_ are a generalization of symbols in the sense that they are a feature to enable increased
+density. The Ion parser expands integer symbol IDs to symbol text, it now also expands macro
+expressions into data of arbitrary type and cardinality.
+* The encoding context is extended to contain a _local macro table_ alongside the local symbol
+table. In much the same way that the local symbol table defines an address-space for identifying
+symbols, the local macro table defines an address-space for macros.
+* Symbols and macros are defined and collected inside _encoding modules_.
+Modules subsume shared symbol tables while remaining backwards-compatible with their current data
+model and catalog semantics.
+Any existing Ion shared symbol table is a valid encoding module, albeit one that only declares
+symbols and not macros.
+* The `$ion_symbol_table` struct and its behavior are subsumed by a new `$ion_encoding` top-level
+S-expression that imports and defines modules, then separately assembles the local symbol and
+macro tables.
+
+This chapter focuses on the new components of the top-level context, the `$ion_encoding`
+S-expression that controls it, and the new parts of shared symbol tables that effect shared
+modules.
+// TODO The inner workings of modules are covered by TODO and macros by TODO
+
+=== Document Structure
+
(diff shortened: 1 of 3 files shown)


src/directives.adoc
```diff
@@ -0,0 +1,308 @@
...
+This intro section is probably misplaced in the context of the larger book.
+Move or integrate elsewhere.
+====
+
+Ion 1.0 uses symbol tables to capture and compress repeated symbol text.
+At all points in an Ion document, there exists an _encoding environment_ that contains the _current
+symbol table_ mapping symbol IDs to text.
+The encoding environment of a document is controlled by directives embedded in the document.
+These directives are encoding artifacts and not part of the application data model.
+Ion 1.0 has two of directive forms:
```
zslayton:
```suggestion
Ion 1.0 has two directive forms:
```

src/directives.adoc
```diff
@@ -0,0 +1,308 @@
...
+
+To increase compression across many documents that have similar content (for example, they use the
+same schema), Ion 1.0 has _shared symbol tables_ that capture a portion of an encoding context—a
+list of symbols—that can be imported into many local symbol tables.
+
+Ion 1.1 generalizes and refactors these features:
+
+* _Macros_ are a generalization of symbols in the sense that they are a feature to enable increased
+density. The Ion parser expands integer symbol IDs to symbol text, it now also expands macro
+expressions into data of arbitrary type and cardinality.
```
zslayton:
```suggestion
density. The Ion parser expands integer symbol IDs to symbol text; it now also expands macro
expressions into data of arbitrary type and cardinality.
```

src/directives.adoc
```diff
@@ -0,0 +1,308 @@
...
+Modules subsume shared symbol tables while remaining backwards-compatible with their current data
+model and catalog semantics.
+Any existing Ion shared symbol table is a valid encoding module, albeit one that only declares
+symbols and not macros.
+* The `$ion_symbol_table` struct and its behavior are subsumed by a new `$ion_encoding` top-level
+S-expression that imports and defines modules, then separately assembles the local symbol and
+macro tables.
+
+This chapter focuses on the new components of the top-level context, the `$ion_encoding`
+S-expression that controls it, and the new parts of shared symbol tables that effect shared
```
zslayton: At this stage in my reading, this clause caught my eye:

> the new parts of shared symbol tables

I assume it's referring to the optional syntax needed to declare macros in an SST.

src/directives.adoc
```diff
@@ -0,0 +1,308 @@
...
+The system module and its macros are in fact available everywhere in the document, and cannot be
+removed or redefined by `$ion_encoding`: to a large degree, it's as if the `*retain*`,
+`*symbol_table*`, and `*macro_table*` clauses all have `$ion` as their implicit first element.
+As a result, system macros can always be invoked by `(:$ion:__name__ …)`.
+
+System macros have one additional bit of special handling: they are binary-encoded using a
+dedicated opcode, using a dedicated address space that’s independent of the explicitly-enumerated
+modules in `*macro_table*`.
+This means that the initial range of unqualified numeric macro references like `(:3 …)` don’t
+inherently refer to system macros.
```
zslayton, line 104: This aligns with my understanding. However, I was confused by the prior statement that:

> to a large degree, it's as if .... `macro_table` clauses all have `$ion` as their implicit first element

as that would load those into the one-byte-invocations space. (Maybe that nuance is what "to a large degree" was referring to?)

src/directives.adoc
```diff
@@ -0,0 +1,308 @@
...
+the same algorithm for inexact match on the version.
+
+TIP: A primary design tenet of Ion 1.1 is to remain compatible with existing catalog APIs and
+services that vend shared symbol tables.
+Existing shared symbol tables are shared modules that export no macros.
+
+As suggested by its name, the _symbol-maxid_ argument only affects symbol allocation, not macros.
+Use of macros from a shared module requires exact match of the _shared-version_, and a module that
+was imported inexactly will trigger an error if its name appears within a local module or
+`*macro_table*`.
```
zslayton, line 193: This is in support of 1.0-style SSTs that have tunneled macro definitions, right? Allowing 1.0 consumers to keep using the same SST when they don't need the macros? If so, that makes sense. However, it does feel pretty complex.

src/directives.adoc
```diff
@@ -0,0 +1,308 @@
...
+|===
+include::grammar.adoc[tag=import]
+|===
+
+That is, `(*import* _module_ ...)` is equivalent to `(*load* _module_ ...)(*use* _module_)`.
+
+
+==== Assembling the Symbol Table
+
+Modules must be installed into the symbol table affect encoding of symbols.
```
zslayton:
```suggestion
Modules must be installed into the symbol table to affect the encoding of symbols.
```

src/directives.adoc
toddjonker: Yeah you got the nuance.  I def need to find a better way to explain how system macros are handled.  It's tricky because macro address allocation works differently than symbols in a couple ways now.   Cut #264

src/directives.adoc
toddjonker: Removed the last clause, it's old and obsolete.

src/directives.adoc
toddjonker: It primarily falls out from having module loading for symbols and for macros use the same syntax and "namespace" of available modules.  At the time of `load` we don't know how the module will be used, so as long as we support maxId behavior for symbols then the inexact match is possible.

I didn't think this was complex, just a constraint that's easily detected and messaged.  Can you explain what feels complicated here?

zslayton (approved):

PR #265, opened 2023-06-29
Add chapter on encoding modules.

toddjonker:
### Description of changes:

----

### DIFF
--- a/src/modules.adoc
+++ b/src/modules.adoc
@@ -22,7 +22,7 @@ by the `symbols` field of a shared symbol table.
 
 The exported macro array is denoted by the module’s `*macro_table*` clause, with addresses
 allocated to bindings in the order they are declared.
-One address is allocated per `*macro*` definition, while `*export*` clause allocates one address
+One address is allocated per `*macro*` definition, while the `*export*` clause allocates one address
 for each listed macro.
 
 
@@ -34,11 +34,11 @@ This environment is constructed incrementally by each clause in the definition a
 * the _visible modules_, a map from identifier to module
 * the _imported macros_, a map from identifier to macro (or to an ambiguity sentinel)
 * the _local macros_, a map from identifier to a macro and optional exported address.
-* the _exported macros_, an array containing name/module pairs
+* the _exported macros_, an array containing name/macro pairs
 
 Before any clauses of the module definition are examined, the initial environment is as follows:
 
-* The visible modules maps `$ion` to the system module for the appropriate spec version.
+* The visible modules map `$ion` to the system module for the appropriate spec version.
 An inline module also gets the modules previously made available by the enclosing encoding
 directive (via `*retain*`, `*load*`, or `*import*`).
 * The imported macros contain the exported macros from that system module.
@@ -46,7 +46,7 @@ An inline module also gets the exported macros from modules previously ``*use*``
 ``*import*``ed by the enclosing encoding directive.
 * The exported macros and local macros are empty.
 
-The first section of a module definition consists dependency declarations in the form of `*use*`,
+The first section of a module definition consists of dependency declarations in the form of `*use*`,
 `*load*`, and `*import*` clauses.
 This section affects the environment as follows:
 
@@ -239,7 +239,7 @@ include::grammar.adoc[tag=dependency]
 
 ==== The Symbol Table
 
-A module can define a list of exported symbols by copy symbols from other modules and/or
+A module can define a list of exported symbols by copying symbols from other modules and/or
 declaring new symbols.
 
 [{bnf}]
--- a/src/modules.adoc
+++ b/src/modules.adoc
@@ -1,4 +1,4 @@
-[[sec:directives]]
+[[sec:modules]]
 == Encoding Modules
 
 include::styles.adoc[]
@@ -39,10 +39,10 @@ This environment is constructed incrementally by each clause in the definition a
 Before any clauses of the module definition are examined, the initial environment is as follows:
 
 * The visible modules map `$ion` to the system module for the appropriate spec version.
-An inline module also gets the modules previously made available by the enclosing encoding
-directive (via `*retain*`, `*load*`, or `*import*`).
+For an inline module, it also includes the modules previously made available by the enclosing
+encoding directive (via `*retain*`, `*load*`, or `*import*`).
 * The imported macros contain the exported macros from that system module.
-An inline module also gets the exported macros from modules previously ``*use*``d or
+For an inline module, it also contains the exported macros from modules previously ``*use*``d or
 ``*import*``ed by the enclosing encoding directive.
 * The exported macros and local macros are empty.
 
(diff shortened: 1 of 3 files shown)


src/modules.adoc
```diff
@@ -0,0 +1,361 @@
...
+* its _spec version_, denoting the Ion version used to define the module
+* its _exported symbols_, an array of strings denoting symbol content
+* its _exported macros_, an array of <name, macro> pairs, where all names are unique identifiers
+(or null).
+
+The spec version for an inline module is implicitly derived from the Ion version of its
+containing segment. The spec version for a shared module is denoted via a required annotation.
+
+The exported symbol array is denoted by the `*symbol_table*` clause of a module definition, and
+by the `symbols` field of a shared symbol table.
```
zslayton, line 21: Might be worth a `Note` or parenthetical reminding the reader that an SST is "just" a module without macros.

src/modules.adoc
```diff
@@ -0,0 +1,361 @@
...
+
+The exported macro array is denoted by the module’s `*macro_table*` clause, with addresses
+allocated to bindings in the order they are declared.
+One address is allocated per `*macro*` definition, while `*export*` clause allocates one address
+for each listed macro.
```
zslayton:
```suggestion
One address is allocated per `*macro*` definition, while the `*export*` clause allocates one address
for each listed macro.
```

Non-blocking: a diagram or example would probably help quite a bit here.

src/modules.adoc
```diff
@@ -0,0 +1,361 @@
...
+
+==== Internal Environment
+
+The body of a module tracks an internal environment by which macro references are resolved.
+This environment is constructed incrementally by each clause in the definition and consists of:
+
+* the _visible modules_, a map from identifier to module
+* the _imported macros_, a map from identifier to macro (or to an ambiguity sentinel)
+* the _local macros_, a map from identifier to a macro and optional exported address.
+* the _exported macros_, an array containing name/module pairs
```
zslayton:
```suggestion
* the _exported macros_, an array containing name/macro pairs
```
?

src/modules.adoc
```diff
@@ -0,0 +1,361 @@
...
+
+Before any clauses of the module definition are examined, the initial environment is as follows:
+
+* The visible modules maps `$ion` to the system module for the appropriate spec version.
```
zslayton: Nit: there's a subject/verb plurality disagreement in "modules maps."
```suggestion
* The visible modules map `$ion` to the system module for the appropriate spec version.
```

src/modules.adoc
```diff
@@ -0,0 +1,361 @@
...
+An inline module also gets the modules previously made available by the enclosing encoding
```
zslayton: I suggest spelling out what "gets" means a bit more here, maybe something like:

> When a new module is defined inline, its visible modules include all modules previously made available by [....]

src/modules.adoc
```diff
@@ -0,0 +1,361 @@
...
+directive (via `*retain*`, `*load*`, or `*import*`).
+* The imported macros contain the exported macros from that system module.
+An inline module also gets the exported macros from modules previously ``*use*``d or
```
zslayton: Maybe "has access to" is a good (somewhat more explicit?) rephrasing of "gets"?

src/modules.adoc
```diff
@@ -0,0 +1,361 @@
...
+``*import*``ed by the enclosing encoding directive.
+* The exported macros and local macros are empty.
+
+The first section of a module definition consists dependency declarations in the form of `*use*`,
+`*load*`, and `*import*` clauses.
```
zslayton:
```suggestion
The first section of a module definition consists of dependency declarations in the form of `*use*`,
`*load*`, and `*import*` clauses.
```

src/modules.adoc
```diff
@@ -0,0 +1,361 @@
...
+
+include::grammar.adoc[tag=macro-ref]
+
+Macro references are resolved to a specific macro as follows:
+
+* An unqualified _macro-name_ is looked up within the local macros, and if not found then the
+imported macros.
+If it maps to a macro, that’s the resolution of the reference.
+Otherwise, if the name maps to the ambiguity sentinel, a syntax error is signaled due to an
+ambiguous reference. Otherwise, a syntax error is signaled due to an unbound reference.
```
zslayton: This may be splitting a hair, but to me this is not a _syntax_ error as it requires evaluation beyond lexing/parsing to recognize. To know if a given reference is invalid, I need to be tracking all of the state of prior imports and evaluating additional rules on top of the syntax.

src/modules.adoc
```diff
@@ -0,0 +1,361 @@
...
+module is as defined by that version of the Ion specification.
+
+The spec version of a shared or tunneled module must be declared explicitly using an annotation
+of the form `$ion_1___N__`.
+This allows the module to be serialized using any version of Ion, and its meaning will not change.
+
+[{nrm}]
+----
+$ion_shared_module::$ion_1_1::(
+  (*catalog_key* "com.example.symtab" 3)
```
zslayton, line 131: Bikeshed: I think `id` or `identifier` would be more intuitive than `catalog_key`. (I know the group has discussed this a bit in the past, just re-registering that thought.)

src/modules.adoc
```diff
@@ -0,0 +1,361 @@
...
+Shared modules exist independently of the documents that use them.
+They are identified by a _catalog key_ consisting of a string name and an integer version.
+When consumed by a document or another module, they are given a local identifier.
+
+[{bnf}]
+|===
+include::grammar.adoc[tag=shared-module]
+|===
+
+Tunneled modules are shared modules that are defined within a shared symbol table definition.
```
zslayton, line 198: This probably isn't the place for it, but somewhere we'll need a disclaimer saying this is mostly for backwards compatibility and new modules should be encoded with the modern syntax.

src/modules.adoc
```diff
@@ -0,0 +1,361 @@
...
+
+[{bnf}]
+|===
+include::grammar.adoc[tag=dependency]
+|===
+
+
+==== The Symbol Table
+
+A module can define a list of exported symbols by copy symbols from other modules and/or
```
zslayton:
```suggestion
A module can define a list of exported symbols by copying symbols from other modules and/or
```

src/modules.adoc
toddjonker: I think it's reasonable to consider an unbound variable reference a "syntax error".  I suspect you are saying that only Ion problems count as syntax errors, and I disagree with that.

That said, I can strike "syntax" here with no loss of meaning, so I will.

src/modules.adoc
toddjonker: The term "identifier" already has an important meaning throughout these docs.

Rationale for this term: I was intentionally trying to establish the notion of a "catalog" as the source of modules/symtabs, and this denotes the key used to lookup things in the catalog.

zslayton (approved):

PR #283, opened 2023-10-02
Rework PDF rendering pipeline to use dblatex

toddjonker:
### Description of changes:

This uses the `xmlto` and `dblatex` packages to transform AsciiDoctor-generated DocBook XML into LaTeX and then to PDF.  This enables a more robust and configurable rendering solution than AsciiDoctor's direct PDF generation, particularly with respect to math notation.  This is effectively a precondition for the denotational semantics document, which is naturally and unavoidably math-heavy.

There's a fair amount of customization and workarounds in here, about half of which were forced by limitations and outright defects in `dblatex`.  I've tried to document everything internally with the "why"s and links to relevant documentation, and that's what's most worth your attention, IMO.

I do not suggest reading too deeply into this PR unless you're particularly curious about the ugly details.

----


toddjonker:
Here are copies of the resulting Ion Spec PDF:
* [IonSpec-before.pdf](https://github.com/amazon-ion/ion-docs/files/12786110/IonSpec-before.pdf)
* [IonSpec-after.pdf](https://github.com/amazon-ion/ion-docs/files/12786111/IonSpec-after.pdf)

Dockerfile
```diff
@@ -1,6 +1,6 @@
 # Inspired by https://github.com/simdjson/simdjson/blob/master/Dockerfile
 
-FROM ubuntu:23.10
+FROM ubuntu:23.04
```
toddjonker, line 3: This didn't build today using 23.10, which is apparently a dev release (in the prior PR I just used the largest Ubuntu I saw on Docker Hub).  Looks like 23.04 is the `rolling` stable release, while the earlier 22.04 is the `latest` LTS based on the [Ubuntu image docs](https://hub.docker.com/_/ubuntu?uuid=D2F7152D-A7D7-47DD-BE46-1136E3260C79).

zslayton (approved):

PR #286, opened 2023-11-14
Swaps the order of the exponent and coefficient components of decimals

tgregg:
### Description of changes:

- Swap the order of the coefficient and exponent components of the decimal encoding in binary, making exponent (as a `FlexInt`) come first, and coefficient (as a `FixedInt`) come last. Exponents are likely to remain small (and in languages like Java, are unsupported outside the range of a 32-bit signed integer), while coefficients may be arbitrarily large. Using a `Fixed` subfield for the coefficient reduces the amount of work that needs to be done to get to and from the big-endian representation of the value (as is required in some languages; see Java's BigInteger) from reverse-and-shift to just reverse. Note: this also retains the same component order as the Ion 1.0 encoding.
- Reclaims `0x6F`, no longer using it as a special case for zero (or negative zero) coefficients. Instead, proposes that an implicit coefficient component of `0` (represented by coefficient component with width 0) represents the coefficient `0`, while an explicit coefficient of `0` (represented by the `FixedInt` `0`) represents a coefficient of `-0`. This allows decimals of length 15 to be encoded using the opcode `0x6F`, while retaining a compact encoding for coefficients of 0.

Alternative considered:

One downside of the new order of decimal components is that integer-value decimals (other than `0d0`) require one more byte to encode because an exponent of 0 can no longer be encoded implicitly. It is possible, instead of reclaiming `0x6F` for length-15 decimals as proposed here, to use `0x6F` to encode decimals with implicit exponent `0`. However, this has the drawback of increased complexity, as the coefficient would have to be a `FlexInt`, requiring the reverse-and-shift logic for arbitrarily large values that the other changes in this PR propose to avoid. Discussion on this topic is welcomed.

----

### DIFF
--- a/src/binary-encoding.adoc
+++ b/src/binary-encoding.adoc
@@ -653,17 +653,15 @@ EB 02
 [[decimals]]
 ==== Decimals
 
-If an opcode has a high nibble of `0x6_`, it represents a decimal. Low nibble values `0x_E` and below indicate
+If an opcode has a high nibble of `0x6_`, it represents a decimal. Low nibble values indicate
 the number of trailing bytes used to encode the decimal.
 
-The body of the decimal is encoded as a <<flexint, `FlexInt`>> representing its coefficient, followed by a `FixedInt`
-representing its exponent. The width of the exponent is the total length of the decimal encoding minus the length
-of the coefficient. It is possible for the exponent to have a width of zero, indicating an exponent of `0`.
+The body of the decimal is encoded as a <<flexint, `FlexInt`>> representing its exponent, followed by a `FixedInt`
+representing its coefficient. The width of the coefficient is the total length of the decimal encoding minus the length
+of the exponent. It is possible for the coefficient to have a width of zero, indicating a coefficient of `0`. When
+the coefficient is present but has a value of `0`, the coefficient is `-0`.
 
-Decimal values that require more than 14 bytes can be encoded using the variable-length decimal opcode: `0xF6`.
-
-A decimal with a coefficient of `-0` (which cannot be encoded as a `FlexInt`) is encoded using opcode `6F`.
-The opcode is followed by a `FlexInt` representing the exponent.
+Decimal values that require more than 15 bytes can be encoded using the variable-length decimal opcode: `0xF6`.
 
 `0xEB 0x03` represents `null.decimal`.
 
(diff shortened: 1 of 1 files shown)


popematt (commented):
> One downside of the new order of decimal components is that integer-value decimals (other than 0d0) require one more byte to encode because an exponent of 0 can no longer be encoded implicitly. It is possible, instead of reclaiming 0x6F for length-15 decimals as proposed here, to use 0x6F to encode decimals with implicit exponent 0. [...] Discussion on this topic is welcomed.

I agree that we should avoid having two different ways of encoding the coefficient. It should be a FlexInt or a FixedInt, but not both. Here are the options I think we have at our disposal:

1. Accept this PR. Take an opinionated stance and say that we're biasing decimals for non-integer values and biasing integers for integer values. Users can use a one-byte macro id for e.g. `1.` if it's important to them.
1. Use `0x6F` to indicate that there is an implicit zero and a 1 byte _FixedInt_ coefficient. (This helps to optimize small numbers (-128 to +127) which are disproportionally impacted by the change you are proposing.)
1. Generalize the previous option a little more and carve out a space for some zero length scales with a few lengths of _FixedInt_ coefficients. E.g.:


| OpCode | Length | Exponent | Coefficient 
|----|----|----|----
| `0x60` | 0 bytes | 0 | 0
| `0x61` | 1 bytes | 0 | 1 byte  FixedInt
| `0x62` | 2 bytes | 0 | 2 byte FixedInt
| `0x63` | 3 bytes | 0 | 3 byte FixedInt
| `0x64` |  1 bytes | FlexInt | 0
| `0x65` | 2 bytes | FlexInt | 0-1 byte FixedInt
| `0x66` | 3 bytes | FlexInt | 0-2 byte FixedInt
| `0x67` | 4 bytes | FlexInt | 0-3 byte FixedInt
| ... | ... | ... | ...
| `0x6F` | 12 bytes | FlexInt | 0-11 byte FixedInt

It's not quite as "pure" as what you are proposing, but the coefficient is always a FixedInt, and an integer-valued decimal up to about ±8,000,000 can be encoded without having an exponent.

zslayton:
Another alternative: the set of built-in macros is not finalized, but we've often discussed having a `make_decimal` macro that took two variable-width integers: the coefficient and the exponent. If we follow through with that, it becomes trivial to define a macro like this:
```ion_1_1
(macro int_to_dec (x)
    (make_decimal x 0))
```
This macro can cost a single byte to invoke and as little as a single byte to write out the int value, making it as efficient as having a purpose-built opcode to achieve the same.

popematt:
That's a good point, Zack. I think that fully mitigates any concern about integer values being enlarged by the proposed change.

popematt (approved):

zslayton (approved):

Issue #287, opened 2023-11-20
Ion 1.1 spec needs to be updated to reflect changes in macro signatures and grouped parameters


popematt:
* In a macro signature, macro parameters are contained in a s-expression rather than a list
* ~~Cardinality of parameters is no longer supported~~
* Parameter types are no longer grouped in a s-expression (e.g.: `(foo int)`)—the type is annotated on the parameter name (e.g. `int::foo`)
* `(:)`, `(:void)`, and `(:values)` macros are superseded by the stream syntax `[:` ... `]`
* A grouped parameter is indicated by putting `[]` around the variable name (e.g. `[foo]`). When the variable has a type annotation, the type annotation stays with the variable name (e.g. `[int::foo]`).
* The binary encoding uses presence bits for encoding macro arguments
* Presence bits are required in the binary encoding if and only if there are 3 or more parameters
* For a non-grouped parameter, a `1` indicates the parameter is present, and `0` indicates argument is not present.
* For a grouped parameter, `1` indicates a present expression group, and `0` indicates no expression group for that parameter.
* The text encoding no longer requires delimiting containers around the arguments for a grouped parameter. Instead, the arguments should be enclosed in the stream syntax.

toddjonker:
I'm wary of several of these proposals.  It's going to be hard to discuss with all of this under one issue.

toddjonker:
* We agreed to delay implementation of parameter cardinality, not eliminate it.  (As I work on the denotational semantic model, I'm becoming more convinced that we need _more_ strong typing, not less.)
* Removing the S-expression from parameter type declarations eliminates a critical extension point.  As an annotation, there's no good way to do anything but a type-name.
* I'm not immediately in favor of the stream syntax and would like to understand its rationale.
* Even given that, we still need void/values as system-defined macros for use in the template language.
* We should still have both `void` and `values`.  I want to allow the former to accept and ignore `any*` and thus act as an easy, structure-aware mechanism to discard a chuck of data. I find myself needing this almost daily when working with Ion-based DSLs.

popematt:
> We agreed to delay implementation of parameter cardinality, not eliminate it. (As I work on the denotational semantic model, I'm becoming more convinced that we need more strong typing, not less.)

That's my mistake.

I've created the following issues to handle these discussions independently from each other.
 * #289 
 * #290 
 * #291

PR #288, opened 2023-11-20
Use updated macro signature and stream syntax

popematt:
### Issue #, if available:

#287 

### Description of changes:

* Changes syntax of a macro parameter to use annotations—e.g. `(name type)` to `type::name`
* Changes macro signature container from being a list to a sexp—e.g. `[ parameter1, parameter2 ]` to `(parameter1 parameter2)`
* Replaces all `(:)`, `(:void)`, and `(:values)` with the updated stream syntax
* Removes the note about deciding "whether we want an Ion 1.0-style double-length-prefixed sequence"—we don't.

This does not update any documentation about presence bits, nor does it remove any of the documentation about cardinality.

----

### DIFF
--- a/src/binary-encoding.adoc
+++ b/src/binary-encoding.adoc
@@ -1930,9 +1930,6 @@ EB 05
 [[annotations]]
 === Annotations
 
-[sidebar]
-TODO: Decide whether we want an Ion 1.0-style double-length-prefixed sequence.
-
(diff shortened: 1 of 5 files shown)


toddjonker:
This is notably missing the corresponding changes to the domain grammar, which are normative while the example chapters are not.

Issue #291, opened 2023-11-21
Consider adding first-class text syntax for streams


popematt:
Sometimes it's difficult to talk about streams without getting confused with macros because right now, the only way to construct a stream is by using a macro (except for the top-level stream).

For the text encoding, I propose creating a new syntax to represent streams. One possibility is to create a "robot smile" (`[:`) that parallels the "smile expressions" (`(:`). Other possibilities include `{` followed by almost any punctuation or symbol character, or using `(:` followed by another punctuation or symbol character, for example `~` to form `(:~`. (For the rest of this writeup, I will use `[:`, but it does not necessarily have to be that specific syntax.)

For the binary encoding, streams can still be represented the same way as a system macro.

#### Implications for grouped parameters
In addition to making it easier to talk/write about streams, having a text syntax for streams also has the benefit of no longer requiring a delimiting list for the arguments of a grouped parameter in the text encoding. For example:

```
(macro ouch [(stuff [list])] …)
```

_Without stream syntax_—To avoid ambiguity, the delimiter is required even for singleton values. Without this rule, the E-expression `(:ouch [])` would be ambiguous whether the parameter was intended to be void or a singleton empty-list value.

_With stream syntax_—The E-expression `(:ouch [])` has a singleton empty-list value as its argument. The E-expression `(:ouch [:])` has an empty stream (or _void_) as its argument.

#### Implications for `(:values)`
Since the `(:values)` macro is essentially the identity function of a stream, if we add stream syntax, we can eliminate the `(:values)` macro.

Defining a stream in a macro template definition can be done the same way as macro invocations. If the stream syntax is `[:`, a stream can be defined in a template using list syntax and lists can be escaped the same way that s-expressions are currently escaped. On the other hand, using something like `(:~` for the stream syntax would mean that we could define a stream in a template using `(~`.

popematt:
Another way of putting this is that `values` is such a fundamental piece in the macro system, that it deserves to have its own syntactical sugar.

popematt:
A syntax for streams would only add confusion unless expression groups, streams, and the `values` macro can be unified. There is no inherent issue preventing these concepts from being merged, but the spec has continued to keep `values`, streams, and expression groups as separate concepts with subtle distinctions.

PR #302, opened 2024-04-01
Replace grouped-parameter lists with new syntax for argument groups.

toddjonker:
The new syntax `(: ...)` denotes an "argument group" and can only occur in an E-expression.  It replaces the use of lists to pass multiple expressions to a single parameter.

"Grouped parameters" are removed from signatures, leaving only cardinality. To use multiple expressions with a parameter, the caller uses an argument group.  A single expression (of appropriate shape) is always acceptable, regardless of cardinality.

In the template language, macro invocations can use the syntax `(; ...)` to denote argument groups.

As you can see from the text changes, this is more simple than the prior syntax and semantics.  It turns grouping into a syntactic concern at the macro invocation site.

### Issue #, if available:

This change is related to #291 but has at least one key difference. As proposed here, argument groups are not general-purpose streams, and can only appear in macro-argument positions.  This is because they have a critical difference from the `values` macro: their contents are shape-constrained by the relevant macro parameter's type, most obviously for macro-shaped parameters.  Personally, I'm in favor of keeping argument groups and `values` invocations syntactically distinct, to reduce confusion.

### Other notes:

* Content explaining when groups are allowed needs to in sync with the binary encoding: text E-expressions should only allow groups where the binary form allows them.
* In editing this I discovered that these docs currently don't have description of text E-expressions; I'll have to circle back and get that content added.

----

### DIFF
--- a/src/glossary.adoc
+++ b/src/glossary.adoc
@@ -10,13 +10,15 @@ A macro can be _fixed arity_ or _variable arity_.
 
 argument::
 A single sub-expression within a macro invocation, corresponding to one of the macro’s
-parameters.  This is by default a one-to-one relation, but a parameter’s grouping mode
-can change this to a many-to-one relation.
+parameters.  This is by default a one-to-one relation, but a rest parameter accepts
+zero-or-more or one-or-more arguments.
 
 argument group::
-The concrete syntax for a _grouped parameter_.
-In a macro invocation, a list containing multiple arguments, delimited explicitly with `[...]`
-notation in Ion text. Used to express parameters that accept more than one argument.
+The concrete syntax for passing a stream of values to a macro parameter.
+In a text E-expression, a group starts with the digraph `(:` and ends with `)`,
+similar to an S-expression.
+In the template language, a group is written as an S-expression starting with
+`;` (semicolon).
 
 cardinality::
 Describes the number of values that a parameter will accept when the macro is invoked.
@@ -70,16 +72,6 @@ The number of parameters declared by a macro.  Due to _optional parameters_ and 
 the _actual arity_ of a macro invocation may differ from the formal arity of the macro being
 invoked.
 
-grouped parameter::
-A macro parameter that accepts multiple arguments in an _argument group_ when the macro is
-invoked.
-See also _grouping mode_.
-
-grouping mode::
-One of three ways that a macro parameter is given arguments. A _simple parameter_ accepts one
-argument, a _grouped parameter_ accepts a list of arguments, and a _rest parameter_ accepts
-“all the rest” of the trailing arguments without a grouping list.
-
 Ion version marker::
 A keyword directive that denotes the start of a new segment encoded with a specific Ion version.
 Also known as “IVM”.
@@ -123,7 +115,6 @@ rest parameter::
 A macro parameter—always the final parameter—declared with the `*\...*` or `*\...+*` modifier,
 that accepts all remaining arguments to the macro as if they were in an implicit _argument group_.
 Similar to “varargs” parameters in Java and other languages.
-See also _grouping mode_.
 
 segment::
 A contiguous partition of a document that uses the same encoding environment. Segment boundaries
@@ -134,10 +125,6 @@ signature::
 The part of a macro definition that specifies its “calling convention”, in terms of the shape,
 type, and cardinality of arguments it accepts, and the type and cardinality of the results it
 produces.
-
-simple parameter::
-A macro parameter that matches a single argument when the macro is invoked.
-See also _grouping mode_.
 
 subform::
 A nested portion within some syntactic form of the module or macro declarations.
--- a/src/macros-by-example.adoc
+++ b/src/macros-by-example.adoc
@@ -715,7 +715,7 @@ Thank you to my Patreon supporters:
 
 The non-rest versions of multi-value parameters can be annoying to invoke, since they require the
 use of `values` or some other template to produce the stream of values.  To streamline invocation,
-there's a special delimiting syntax for an _argument group_ collecting the
+there is a special delimiting syntax for an _argument group_ collecting the
 applicable sub-expressions.
 
 [{nrm}]
--- a/src/macros-by-example.adoc
+++ b/src/macros-by-example.adoc
@@ -583,7 +583,7 @@ and the `*\...*`
 ==== Exactly-One
 
 Many parameters expect exactly one value and thus have _exactly-one cardinality_.
-This is the default for ungrouped parameters, but the `*!*` modifier can be used for clarity.
+This is the default cardinality, but the `*!*` modifier can be used for clarity.
 
 This cardinality means that the parameter requires a stream producing a single value, so one
 might refer to them as _singleton streams_ or just _singletons_ colloquially.
@@ -711,71 +711,62 @@ Thank you to my Patreon supporters:
 ----
 
 
-[#grouped_parameters]
-=== Grouped Parameters
+=== Argument Groups
 
 The non-rest versions of multi-value parameters can be annoying to invoke, since they require the
 use of `values` or some other template to produce the stream of values.  To streamline invocation,
-a macro can opt-in to special syntax that uses a list as delimiting syntax to group the
-applicable sub-expressions.  This is denoted by wrapping the parameter's type in `[]`:
+there's a special delimiting syntax for an _argument group_ collecting the
+applicable sub-expressions.
 
 [{nrm}]
 ----
 (*macro* prices
-  [(amount [*number*]),      // <1>
+  [(amount *number{asterisk}*),
    (currency *symbol*)]
   (*for* [(amt amount)]
     (price amt currency)))
 ----
 
-<1> Note the use of `[]` around `*number*`.
-
-This is referred to as a _grouped parameter_, and at invocation it requires a list delimiting its
-_argument group_:
+Here, the parameter `amount` accepts any number of `*number*`s.
+It's easy to provide exactly one:
 
 ----
-(:prices [1, 2, 3] GBP) ⇒ {amount:1, currency:GBP}
+(:prices 12.99 GBP) ⇒ {amount:12.99, currency:GBP}
+----
+
+To provide a non-singleton stream of values, use an _argument group_.
+Inside an E-expression, a group starts with the digraph `(:` similar to a macro
+invocation, but without a macro reference:
+
+----
+(:prices (:) GBP)       ⇒ _void_
+(:prices (: 1) GBP)     ⇒ {amount:1, currency:GBP}
+(:prices (: 1 2 3) GBP) ⇒ {amount:1, currency:GBP}
                           {amount:2, currency:GBP}
                           {amount:3, currency:GBP}
 ----
 
-Within the group, the invocation can have any number of arguments, including macro invocations.
+Within the group, the invocation can have any number of expressions that align
+with the parameter type.
 The macro parameter produces the results of those expressions, concatenated into a
 single stream, and the expander verifies that each value on that stream is acceptable by the
 parameter’s declared type.
 
 ----
-(:prices [1, (:values 2 3), 4] GBP) ⇒ {amount:1, currency:GBP}
+(:prices (: 1 (:values 2 3) 4) GBP) ⇒ {amount:1, currency:GBP}
                                       {amount:2, currency:GBP}
                                       {amount:3, currency:GBP}
                                       {amount:4, currency:GBP}
 ----
 
-[IMPORTANT]
-====
-To avoid ambiguity, the delimiter is required even for singleton values.  Consider this
-macro:
+Argument groups are allowed for parameters that do not have exactly-one cardinality.
 
-[{nrm}]
-----
-(*macro* ouch [(stuff [*list*])] …)
-----
+// TODO
+WARNING: The allowed combinations of cardinality and argument groups is pending
+finalization of the binary encoding.
 
-Without this rule, the E-expression `(:ouch [])` would be ambiguous whether the parameter was
-intended to be void or a singleton empty-list value.
-====
-
-Grouping says whether multiple _arguments_ can be provided, while cardinality describes the
-number of _values_ those argument(s) must produce.
-The parameter declaration `(amount [*number*])` makes grouping explicit, with a default
-cardinality of zero-or-more.
-The declaration `(amount *[number]+*)` is also valid,
-indicating that the sequence of arguments must produce at least one value.
-
-TIP: Grouped parameters cannot use the `*?*` and `*!*` modifiers; there's no point in
-requiring a grouping list when no more than one value is allowed.
-
-TIP: Rest parameters are effectively another grouping mode, so they cannot be combined with `[]`.
+TIP: Rest parameters effectively use implicit groups at the call site, and they cannot be
+invoked with explicit groups.
 
 Delimiting sequences and `values` expressions may appear similar because they both denote streams
 of values, but they are not interchangeable:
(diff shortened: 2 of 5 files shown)


src/glossary.adoc
```diff
@@ -10,13 +10,15 @@ A macro can be _fixed arity_ or _variable arity_.
...
+parameters.  This is by default a one-to-one relation, but a rest parameter accepts
+zero-or-more or one-or-more arguments.
 
 argument group::
-The concrete syntax for a _grouped parameter_.
-In a macro invocation, a list containing multiple arguments, delimited explicitly with `[...]`
-notation in Ion text. Used to express parameters that accept more than one argument.
+The concrete syntax for passing a stream of values to a macro parameter.
+In a text E-expression, a group starts with the digraph `(:` and ends with `)`,
+similar to an S-expression.
```
popematt, line 19: Suppose I have macros named "foo" and "bar". How do we prevent this ambiguity?
```
(:foo (:bar baz))
```
Is this an invocation of `foo` with an expression group of `'bar' 'baz'` or is it an invocation of `foo` where the argument is the result from the invocation of `(:bar baz)`?

src/macros-by-example.adoc
```diff
@@ -711,71 +711,62 @@ Thank you to my Patreon supporters:
...
 
-[#grouped_parameters]
-=== Grouped Parameters
+=== Argument Groups
 
 The non-rest versions of multi-value parameters can be annoying to invoke, since they require the
 use of `values` or some other template to produce the stream of values.  To streamline invocation,
-a macro can opt-in to special syntax that uses a list as delimiting syntax to group the
-applicable sub-expressions.  This is denoted by wrapping the parameter's type in `[]`:
+there's a special delimiting syntax for an _argument group_ collecting the
```
popematt: Nit
```suggestion
there is a special delimiting syntax for an _argument group_ collecting the
```

src/macros-by-example.adoc
```diff
@@ -711,71 +711,62 @@ Thank you to my Patreon supporters:
...
+(:prices (:) GBP)       ⇒ _void_
+(:prices (: 1) GBP)     ⇒ {amount:1, currency:GBP}
+(:prices (: 1 2 3) GBP) ⇒ {amount:1, currency:GBP}
                           {amount:2, currency:GBP}
                           {amount:3, currency:GBP}
 ----
 
-Within the group, the invocation can have any number of arguments, including macro invocations.
+Within the group, the invocation can have any number of expressions that align
+with the parameter type.
```
popematt, line 757: What are the implications of "that align with the parameter type"? Does this require type checking of expression groups in the macro evaluator? (Seems like this might have some of the trouble of type vs encoding.)

src/macros-by-example.adoc
```diff
@@ -711,71 +711,62 @@ Thank you to my Patreon supporters:
...
 
-TIP: Grouped parameters cannot use the `*?*` and `*!*` modifiers; there's no point in
-requiring a grouping list when no more than one value is allowed.
+// TODO
+WARNING: The allowed combinations of cardinality and argument groups is pending
+finalization of the binary encoding.
 
-TIP: Rest parameters are effectively another grouping mode, so they cannot be combined with `[]`.
+TIP: Rest parameters effectively use implicit groups at the call site, and they cannot be
+invoked with explicit groups.
```
popematt, line 780: Why not? We already have rules that define how to combine a stream in an expression group, allowing things like `(:prices (: 1 (:values 2 3) 4) GBP)`.

Is it legal to nest expression groups for non-"rest" parameters? E.g.: `(:prices (: 1 (: 2 3) 4) GBP)`

src/macros-by-example.adoc
```diff
@@ -993,19 +983,18 @@ parameter produces a stream of structs.
...
 [{nrm}]
 ----
 (*macro* scatterplot
-  [(points [point] *+*), (x_label *string*), (y_label *string*)]
+  [(points point *+*), (x_label *string*), (y_label *string*)]
   { points: [points], x_label: x_label, y_label: y_label })
 ----
 ----
-(:scatterplot [(3 17), (395 23), (15 48), (2023 5)] "hour" "widgets")
+(:scatterplot (: (3 17) (395 23) (15 48) (2023 5)) "hour" "widgets")
```
popematt: 1. What would you think about having some syntax in the text encoding that is specific for macro-type arguments... something that looks like a macro invocation but has some symbol that indicates that the name is redundant in that position? (Strawman suggestion: `(:*` ... `)`.) It might help improve readability in cases where there are also s-exps in the macro arguments.
2. Suppose `point` can accept multiple coordinates, then one could have something like this?
```
(:scatterplot (: (3 (:17 18 19)) (395 23) (15 48) (2023 5)) "hour" "widgets")
```

src/glossary.adoc
zslayton: I talked this over with Matt offline. The solution--that whitespace following `:` is the differentiator--feels like a big stumbling block. I think newcomers would often incorrectly consider:
```
(:foo (:bar baz))
```
and
```
(:foo (: bar baz))
```
to be equivalent. To be clear, I think this is just a syntactic issue. I don't object to the underlying concept of expression groups being their own construct.

I wonder if a different `[`-based enclosing syntax would work for this--say, robot smiley (`[:`), Kleene star (`[*`), or an angle bracket (`[>`).

|Syntax | E-exp arg | Expr group arg |
|-------------------|---|---|
| `(: ... )` as expr group |  `(:foo (:bar baz))` |  `(:foo (: bar baz))`  |
|`[: ... ]` as expr group |  `(:foo (:bar baz))` |  `(:foo [:bar baz])` |
|`[* ... ]` as expr group |  `(:foo (:bar baz))` |  `(:foo [*bar baz])` |
|`[> ... ]` as expr group |  `(:foo (:bar baz))` |  `(:foo [>bar baz])` |

These are more visually distinct from macro invocations IMO.

We could also include the sigil on the closing `]` if that's helpful.

|Syntax | E-exp arg | Expr group arg |
|-------------------|---|---|
|`[: ... :]` as expr group |  `(:foo (:bar baz))` |  `(:foo [:bar baz:])` |
|`[* ... *]` as expr group |  `(:foo (:bar baz))` |  `(:foo [*bar baz*])` |
|`[> ... <]` as expr group |  `(:foo (:bar baz))` |  `(:foo [>bar baz<])` |

src/macros-by-example.adoc
```diff
@@ -711,71 +711,62 @@ Thank you to my Patreon supporters:
...
-intended to be void or a singleton empty-list value.
-====
-
-Grouping says whether multiple _arguments_ can be provided, while cardinality describes the
-number of _values_ those argument(s) must produce.
-The parameter declaration `(amount [*number*])` makes grouping explicit, with a default
-cardinality of zero-or-more.
-The declaration `(amount *[number]+*)` is also valid,
-indicating that the sequence of arguments must produce at least one value.
+Argument groups are allowed for parameters that do not have exactly-one cardinality.
```
zslayton: Refreshingly simple. :)

src/macros-by-example.adoc
zslayton: I think your question is conflating expression groups and `(:values ...)`, which this PR hopes to make into distinct entities.

> Is it legal to nest expression groups for non-"rest" parameters? E.g.: `(:prices (: 1 (: 2 3) 4) GBP)`

No, this isn't legal; expression groups can only appear in argument positions. It's easier to understand why when you consider the binary encoding.  Say this is the definition of your macro (borrowed from the doc):

```
(macro prices
  [(amount number*),(currency symbol)]
  // ...BODY EXPR...
)
```

When the binary reader encounters the `amount` argument, it will know to decode an expression group and decode its contents: a sequence of tagged values that must be `number`s. There is no opcode for "expression group", so there's no way to encode `(: 2 3)` within the outer expression group.

I think we _could_ allow an explicit expression group for a rest parameter in text, but I don't think we want to.

>> Rest parameters effectively use implicit groups at the call site, and they cannot be
invoked with explicit groups.

> Why not? We already have rules that define how to combine a stream in an expression group, allowing things like (:prices (: 1 (:values 2 3) 4) GBP).

We could permit it semantically, but it would be strictly redundant syntactically. As an example, let's flip `currency` and `amount` so the latter can be a rest parameter.

```
(macro prices
  [(currency symbol) (amount ...number)]
  // ...BODY EXPR...
)

(:prices USD 1 2 3 4)
(:prices USD (: 1 2 3 4)) // if we allowed "explicit rest"
```
We'd need to add an extra branch to support this, and I don't think it buys us much. In fact, I think that someone passing an explicit expression group to a `...` parameter is a helpful indication that they've misunderstood something and an opportunity to emit a helpful error message.

src/macros-by-example.adoc
zslayton: I think this is something else that a more visually distinctive expression group would help to address.

> What would you think about having some syntax in the text encoding that is specific for macro-type arguments... something that looks like a macro invocation but has some symbol that indicates that the name is redundant in that position?

Something I really like about the expression group is that its contents are always--for any encoding, including macro-shaped--tied to the corresponding parameter's declared encoding. When an exactly-one parameter has a macro-shaped encoding, the text representation of that encoding is an s-expression containing the arguments to the declared macro. 

_(These examples use signature syntax from recent discussions instead of the as-specced syntax)_
```
(macro point2d (x y) /* body expr */)
(macro foo (point2d::x) /* body expr */)

(:foo (1 2))
```

When there's a macro-shaped expression group, it's an expression group that contains zero or more instances of the same sexp-of-args syntax.

```
(macro point2d (x y) /* body expr */)
(macro foo (point2d::x*) /* body expr */)

(:foo (: (1 2) (3 4) (5 6) ))
```

I think making macro-shaped expression groups into yet another construct would be added complexity for limited benefit.

src/glossary.adoc
toddjonker: My intention was that whitespace or `)` must follow. I'm also inclined to spec, at least early on, that a macro-ref can _only_ be followed by whitespace or `)`.  This is intended to avoid extending the tokenization weirdness we already have inside S-expressions and make the E-expression "opening" visually separate from its content.

I don't think there's ambiguity; it's in the spec that the characters immediately after `(:` must be a macro reference or `)`, else there's an error; this adds a distinct case for whitespace.  A user who mistakes the difference would likely learn from the immediate syntax error.

I'm not sure why the `[` markers would be more intuitive.  In the examples here I'm surprised by the lack of commas, which adds the need for sexp-like rules regarding tokenization, eg `[:bar@foo]`.

Adding a closing marker introduces another question regarding empty-group;  `[::]` feels off to me.

_Perhaps_ the way to slice this is to change the requirement that the leading colon must abut the paren, and instead tokenize it with the following identifier/integer.  Thus allowing `( :macro ...)`.  Then we can teach that the colon is either (and only) the start of the macro-ref or a group.  But I hesitate because (a) elsewhere in Ion "whitespace" includes commas, and (b) this means the parser may need to scan an unbounded amount of "whitespace" looking for that colon.  This is already a problem with annotations, and I'm loathe to repeat it.

src/macros-by-example.adoc
toddjonker: I think adding syntax for macro-encoded arguments defeats the purpose, which is to improve brevity.  I think

```
(:foo (: (:* 1 2) (:* 3 4) (:* 5 6)))
```
is significantly less human-friendly than:
```
(:foo (: (1 2) (3 4) (5 6)))
```

The concept also
* intersects with the ungrouped case: does one write `(:foo (1 2))` or `(:foo (:* 1 2))`?
* requires additional tokenization rules following the `(:*`

src/macros-by-example.adoc
toddjonker: @zslayton expresses my thinking well.  In particular, this is great:

> There is no opcode for "expression group"

src/macros-by-example.adoc
toddjonker: It should probably read "base encoding" or something; the terminology is muddled and inconsistent.  Also this would be clarified in the missing section tracked by #303 ... we somehow missed the section defining text E-expression syntax that would be affected by this PR.

src/glossary.adoc
zslayton: > I don't think there's ambiguity; it's in the spec that the characters immediately after (: must be a macro reference or ), else there's an error; this adds a distinct case for whitespace. A user who mistakes the difference would likely learn from the immediate syntax error.

It's not ambiguous for a parser, but it's ambiguous if you're a human. :)

> I'm not sure why the [ markers would be more intuitive. In the examples here I'm surprised by the lack of commas, which adds the need for sexp-like rules regarding tokenization, eg [:bar@foo]

I'm not prescribing a solution, I'm illustrating that having a syntax wherein expressions as similar as `(:foo)` and `(: foo)` having two different meanings is readily avoidable. Users will definitely learn of their mistake when they stub their toe on it, but I think we could spare them that trouble.

> Perhaps the way to slice this is to change the requirement that the leading colon must abut the paren,

I would also prefer not to because of the reasons you stated.

popematt: > I'm not sure why the `[` markers would be more intuitive.

I'm not looking for "more intuitive"—just something that is less "potential foot-gun" than `(: foo)` vs `(:foo)`.

It just so happens that `[:` is currently not legal Ion, so it's something that _could_ be used.

What I find incredibly *un*intuitive is that these are all equivalent:
* `(foo)` and `( foo)`
* `[foo]` and `[ foo]`
* `{foo:1}` and `{ foo:1}`
* `{{abc12=}}` and `{{ abc12=}}`
 
...but `(:foo)` and `(: foo)` are not equivalent _and_ they are both valid Ion.

> I'm surprised by the lack of commas, which adds the need for sexp-like rules regarding tokenization

If we do use `[:` or similar, then I think commas are fine.

If we're concerned about tokenization simplicity or efficiency, then why not use something like `[:`. Instead of having to check three characters to determine whether we're looking at the start of a macro vs an expression group, we only have to look at two characters—`(: ` vs `[:`. If we know because of the context that it _must_ be either a macro or expression group, then we only actually need to look at the first character—`(` or `[`—to determine which it is.

popematt: Sorry if I'm belaboring this point, but I thought of one last thing. This will be the last one, I promise.

> Users will definitely learn of their mistake when they stub their toe on it, but I think we could spare them that trouble.

Ideally, we have lots of Ion users (who could each potentially make this mistake), and even if users learn and are aware of it, it's easy to accidentally add a single whitespace. We should assume that users (collectively) will make this mistake over and over, so we should try to minimize the potential impact of and time to identify the mistake. 

When a user accidentally changes e.g. `(:foo "bar-" (:baz a b c))` to `(:foo "bar-" (: baz a b c))`:
* If argument groups start with `[:` (or similar), then they _only_ stub their toe because the result is invalid Ion. A syntax highlighter can tell you immediately that it's syntactically wrong and exactly where it is wrong.
* If argument groups start with `(: `, then the user potentially shoots their foot off by producing new data that is valid Ion but different than what they intended. It _might_ end up failing a cardinality check somewhere, but that's not guaranteed. Valid, subtly different data can only reliably be detected by a tool (or person) that is capable of understanding the application-defined semantics of the data. The problem might not be detected right away, and once it is identified that there is a problem, it won't necessarily be obvious where the mistake was made (especially if the "new" data is passed around between multiple systems or transformed in some way before a discrepancy is noticed).

If you use one of `[:`, `[>`, etc. _and_ have commas between elements in the argument group, then it's even more obvious when you have a macro vs an argument group, and it's significantly more difficult to turn an argument group into a macro invocation (or vice versa) by accident.

popematt: I think it would help me understand your point of view if you can explain why you think `(: ` is the best syntax for argument groups.

toddjonker: There's been a lot of good discussion on the syntax, but I'd like to avoid blocking this PR on it.  Changing the concrete syntax with low impact (since there's currently no usage) and happen next, but many other low-level things pivot on moving the grouping concept from parameter-declaration time to invocation-time.  I'd like to unblock that work by getting this merged while leaving the syntax open for discussion.

popematt (approved):

toddjonker:
I believe we have consensus, modulo the concrete syntax.  Holding off on merging to let people review my updates yesterday.

toddjonker:
Ooops @popematt our messages passed in spacetime.  Did you get a chance to see the latest updates?

popematt:
@toddjonker Yes, I did see the updates.

src/glossary.adoc
toddjonker: It occurs to me that another potential syntax for argument groups is `(, ...)` since comma, like colon, cannot appear unquoted in an S-expression.  It arguably carries the correct connotation of "sequence of elements", and might be better aligned with the current template-language equivalent `(; ...)`

I would not allow comma inside such a comma-group. I would want everything but the comma to follow S-expression tokenization rules, so that it's consistent with the TL syntax.

Issue #307, opened 2024-04-10
Be more thorough defining when parameter type/cardinality violations are signaled


toddjonker:
Many sections are hand-wavy about this since it's not yet determined in detail.

PR #308, opened 2024-04-10
Specify syntax of Text E-expressions, and align with template invocations.

### LINKED ISSUE #303: Spec is missing section defining text E-expressions
Somehow this was missed when we exported content to this repository.

toddjonker:
Resolves #303

### Description of changes:

I've once again tried a new narrative flow for explaining the effects of parameter type and cardinality on argument syntax.  Here I've introduced a new term _individual argument_ to mean an ungrouped subexpression, and in contrasted with _grouped argument_ (which collects individual arguments).  I think this is more clear that earlier iterations, but feedback is welcomed.

----

### DIFF
--- a/src/binary-encoding.adoc
+++ b/src/binary-encoding.adoc
@@ -359,7 +359,20 @@ The meanings of each opcode are organized loosely by their high and low nibbles.
 |===
 
 
+[[bin:eexp]]
 === Encoding Expressions
+
+The encoding of E-expressions is designed to balance density and generality.
+For example, they enable encodings with minimal tag bits, even none at all given
+a thoughtful signature. This increases density, but limits generality at the point
+of macro invocation.
+
+The <<sec:eexprs,text>> and binary forms of E-expressions enforce the same
+syntactic constraints on the type and range of data allowed as arguments.
+Any syntactically well-formed E-expression can be transcoded between text and binary,
+without expansion and without changing semantics, and independent of whether it can
+be expanded successfully.
+
 
 [[e_expression_with_the_address_in_the_opcode]]
 ==== E-expression With the Address in the Opcode
--- a/src/eexprs.adoc
+++ b/src/eexprs.adoc
@@ -63,7 +63,8 @@ The parameter type constrains the syntax of an individual argument as follows:
   primitive type.
 * For macro types, an individual argument must be an unannotated S-expression
   containing arguments acceptable to that macro’s signature.
-  Here, the S-expression is implicitly converted to the equivalent E-expression.
+  This is called a _macro-shaped argument_.
+  A macro-shaped argument is implicitly converted to the equivalent E-expression.
 
 The parameter cardinality constrains the syntax of the overall argument
 element(s), particularly whether a group is allowed.
--- a/src/eexprs.adoc
+++ b/src/eexprs.adoc
@@ -40,12 +40,12 @@ The syntax that can appear in each argument position is constrained by the
 macro’s signature.
 
 The number of argument elements (that is, the invocation’s actual arity)
-must be equal to or greater than the macro’s minimum arity, and at most its
-maximum arity, when one exists.
-In other words, an E-expression must contain one element for each required
-parameter, followed by optional elements for the remaining optional parameters.
+must be equal to or greater than the macro’s minimum <<def:arity,arity>>,
+and at most its maximum arity, when one exists.
+In other words, an E-expression must contain one element for each
+<<def:required-param,required parameter>>, followed by optional elements for the
+remaining <<def:optional-param,optional parameters>>.
 
-// TODO links for arity, optional parameters
 // TODO base type? base shape?  base form?  encoding?
 
 The syntax of each argument is defined by the associated parameter’s cardinality
--- a/src/eexprs.adoc
+++ b/src/eexprs.adoc
@@ -1,0 +1,100 @@
+[[sec:eexprs]]
+== Encoding Expressions
+
+Understanding macro signatures, we can now discuss how macros are leveraged to
+encode data.  The syntax for macro invocation is called an _encoding expression_
+or E-expression.  When the Ion parser encounters an E-expression, it automatically
+replaces it with the values produced by the corresponding macro transformation
+function.  The inputs to that transformation are determined by the arguments
+within the E-expression.
+
+IMPORTANT: This chapter details the syntax of E-expressions in Ion text format;
+the corresponding <<bin:eexp,Ion binary encoding>> enforces the
+equivalent constraints.
+
+In Ion Text, E-expressions look similar to S-expressions but are opened by `(:`
+and a reference to the macro that expands the expression.
+Text-encoded E-expressions have one of several forms that differ in how they
+reference the macro to be invoked:
+
+* `(:__macro-name__ …)` lookup by unambiguous name in system or local macro tables.
+* `(:__address__ …)` lookup by address in current macro table (not the system table).
+* `(:__module-name__:__macro-name__ …)` lookup by name in an installed module.
+* `(:__module-name__:__address__ …)` lookup by address in an installed module.
+
+// TODO link or write more precise resolution rules.
+
+NOTE: The parenthesis, colon, and macro reference are a single syntactic token,
+allowing no white space.
+The names are neither string nor symbol tokens, and thus may not use quotes or
+escapes or `$_uint_` symbol-table addresses.
+This reflects the idea that macros behave like new syntactic forms, with this
+entire character sequence determining the syntax that follows inside the
+expression.
+
+Following the opening macro reference and whitespace, E-expressions follow
+S-expression tokenization and whitespace rules.
+The remaining elements are arguments supplying inputs to the macro,
+each one either _individual_ or _grouped_.
+The syntax that can appear in each argument position is constrained by the
+macro’s signature.
+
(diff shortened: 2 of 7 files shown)


src/template-expr.adoc
```diff
@@ -199,59 +200,81 @@ https://github.com/amazon-ion/ion-docs/issues/201
...
 === Type Checking
 
 TODO
 
+// TODO #307 specify when the above constraints are enforced; particularly whether
+// an unused but not well-formed macro signals an error.
+
+// TODO #307 clarify when type and cardinality constraints are enforced.
+// Type and cardinality constraints are _also_ applied each time the macro is invoked,
+// ensuring that the type and number of values provided to a parameter _after arguments are expansnded_
```
zslayton:
```suggestion
// ensuring that the type and number of values provided to a parameter _after arguments are expanded_
```

src/eexprs.adoc
```diff
@@ -0,0 +1,100 @@
...
+The remaining elements are arguments supplying inputs to the macro,
+each one either _individual_ or _grouped_.
+The syntax that can appear in each argument position is constrained by the
+macro’s signature.
+
+The number of argument elements (that is, the invocation’s actual arity)
+must be equal to or greater than the macro’s minimum arity, and at most its
+maximum arity, when one exists.
+In other words, an E-expression must contain one element for each required
+parameter, followed by optional elements for the remaining optional parameters.
```
popematt: This wording is confusing because it's not clear if "required parameter" in this context means "a parameter with `!` cardinality" or "an argument that must be provided at the call site" (and similarly for "optional parameter"). Some readers may understand it to imply that `?` parameters must come after `!` parameters in the signature.

I think some examples would go a long way here.

src/eexprs.adoc
```diff
@@ -0,0 +1,100 @@
...
+The syntax of each argument is defined by the associated parameter’s cardinality
+and type.
+The cardinality determines whether an argument group can be used to collect a
+series of individual arguments, and the type determines the syntax of an
+individual argument, grouped or not.
+
+The parameter type constrains the syntax of an individual argument as follows:
+
+* For tagged types, an individual argument must be a single E-expression or any
+  datum (which may contain nested E-expressions).
```
popematt, line 60: Are we defining "datum" somewhere? Is it distinct from "value" in some way?

src/eexprs.adoc
```diff
@@ -0,0 +1,100 @@
...
+  or an argument group `(: …)`.
+* A rest parameter captures all remaining arguments of the E-expression,
+  each of which must match the individual argument syntax.
+  For a `...+` parameter there must be at least one such argument.
+
+// TODO clarify whether a `+` group must contain at least one element.
+
+The rules above determine whether an E-expression is _well-formed_.
+Any violation of the above constraints must signal a syntax error when the
+E-expression is parsed.
```
popematt, line 90: Does this imply anything about the eagerness of the parsing? Or whether an Ion reader may skip parsing some data in a stream (e.g. by skipping past a container)?

src/eexprs.adoc
```diff
@@ -0,0 +1,100 @@
...
+The parameter type constrains the syntax of an individual argument as follows:
+
+* For tagged types, an individual argument must be a single E-expression or any
+  datum (which may contain nested E-expressions).
+* For primitive types, an individual argument must be a non-null, non-annotated
+  datum of the corresponding concrete type, within the range accepted by the
+  primitive type.
+* For macro types, an individual argument must be an unannotated S-expression
+  containing arguments acceptable to that macro’s signature.
+  Here, the S-expression is implicitly converted to the equivalent E-expression.
```
popematt: Could we add a name for these?
```suggestion
  This is called a _macro-shaped argument_.
  A macro-shaped argument is implicitly converted to the equivalent E-expression.
```

src/template-expr.adoc
```diff
@@ -199,59 +200,81 @@ https://github.com/amazon-ion/ion-docs/issues/201
...
-than the macro’s minimum arity, and at most its maximum arity, when one exists.
-In other words, an invocation must contain one subform for each required parameter, followed by
-optional subforms for the remaining optional parameters.
+<<sec:sysmod,macros provided by the Ion implementation>>, but there are multiple sources of macros:
+
+* the defining module’s internal environment (which is being incrementally extended with each definition)
+* the macros exported from modules ``load``ed by the enclosing module
+* the macros exported from modules ``load``ed by the enclosing `$ion_encoding` directive
+
+The syntax for macro invocation in a template is similar to that of <<sec:eexprs,E-expressions>>.
```
popematt, line 209: There is a lot in this section that is repeated. Could we explain macro invocations in a way that DRYs this?

src/eexprs.adoc
toddjonker: "required parameter" should be a formally-defined term: a parameter is _required_ if its not _optional_.  Specifically, every `!` / `+` / `...+` parameter is required, as are any preceding parameters.

I thought that was in the glossary but apparently it's not, so I'll add.

toddjonker: Also, there are examples in 3.12 in the "macros by example" section that include required `?` and `*` parameters.  I'll add a link.

src/eexprs.adoc
toddjonker: We've not defined 'datum', and it might not be the best term. (Other than this line it's only used in the grammar: `(**literal** _datum_)`.  

I'm using it here to mean "syntax tree" as the semantic model defines it, but that term hasn't been incorporated into this document.

"Value" is not correct; since these are syntax trees and we use "value" to denote the results of expanding syntax trees.

Another way to write this might be:

> For tagged types, an individual argument must be a single syntax tree, which may be an E-expression and/or contain nested E-expressions.

But ATM that doesn't seem better since we don't currently talk in detail about the expansion process.  (Another gap in the document!)

src/template-expr.adoc
toddjonker: Perhaps, but there are also differences that make it hard to slot a unified description into either chapter. 
I'm inclined to leave that for later cleanup.

src/eexprs.adoc
toddjonker: I tried to avoid any such implication here, since I don't think we know exactly how that should work.   I think that gap is part of #307 noted in the next paragraph.

popematt (approved):

src/binary-encoding.adoc
```diff
@@ -359,8 +359,21 @@ The meanings of each opcode are organized loosely by their high and low nibbles.
...
+The encoding of E-expressions is designed to balance density and generality.
+For example, they enable encodings with minimal tag bits, even none at all given
+a thoughtful signature. This increases density, but limits generality at the point
+of macro invocation.
+
+The <<sec:eexprs,text>> and binary forms of E-expressions enforce the same
+syntactic constraints on the type and range of data allowed as arguments.
+Any syntactically well-formed E-expression can be transcoded between text and binary,
+without expansion and without changing semantics, and independent of whether it can
+be expanded successfully.
```
zslayton, line 374: > without expansion

I'm not sure this is true in the case of `binary -> text`, since the e-expressions can't be parsed without knowing the corresponding signature. A given e-expression can be transcoded without being expanded, but a degree of expansion has to happen to ensure that all encoding directives have been found/processed.

src/eexprs.adoc
```diff
@@ -0,0 +1,101 @@
...
+individual argument.
+
+* A `!` parameter accepts only a single individual argument.
+* A `?` parameter accepts either a single individual argument
+  or an empty argument group `(:)`.
+* A `*` or `+` parameter accepts either a single individual argument,
+  or an argument group `(: …)`.
+* A rest parameter captures all remaining arguments of the E-expression,
+  each of which must match the individual argument syntax.
+  For a `...+` parameter there must be at least one such argument.
```
zslayton, line 84: I think our presence bits encoding on the binary side makes it feasible to accept any expression in each position and enforce cardinality constraints uniformly at expansion time.

EDIT: I see #307 covers this discussion

zslayton (approved):
I added a few comments, but none are blockers.

src/binary-encoding.adoc
toddjonker: Do you have a scenario in mind to illustrate?

I think that given the signature, you can transcode.  You also need signatures of macro-shaped arguments, recursively.  But I don't think you need any expansion (meaning you don't need to compile the given macros, you just need their signatures.).   I'm writing the semantic model such that expansion is based off the syntax trees, which are intended to be more-or-less 1:1:1 with the text and binary format.

zslayton: If a top level macro produces an encoding directive, that directive will need to be processed to learn the signatures of the macros it defines. In that sense, some degree of expansion is required.

popematt: > If a top level macro produces an encoding directive, that directive will need to be processed to learn the signatures of the macros it defines. In that sense, some degree of expansion is required.

This also applies to `text -> binary`.

toddjonker: Ahh yeah... top-level expansion.  Gosh we kinda broke my whole mental model about how Ion works.  I'll cut an issue to track cleanup of this topic space.

toddjonker: I created #311 to track this topic; I think we need to come up with a better story so users understand how various encoding features affect various transcoding capabilities.

PR #315, opened 2024-05-02
WIP Rough draft illustrating use of semicolon as parameter separator.

toddjonker:
DO NOT MERGE!  For illustration only.

I updated most (if not all) of the examples in the spec to illustrate how separators work as a replacement for argument groups.  I've used semicolons for both E-expressions and TL invocations, since this process convinced me that the inconsistent use of comma was too confusing.

One thing I'm still considering is whether we should use separators after `!` and `?` single-expression parameters.  They aren't really needed there if we have a way to write "no value" for the latter case, and omitting them could avoid some verbiage.  I'm not convinced, just considering.  :-)

----

### DIFF
--- a/src/macros-by-example.adoc
+++ b/src/macros-by-example.adoc
@@ -93,7 +93,7 @@ This macro has a signature that declares two parameters, named `a` and `c`, and 
 therefore accepts two arguments when invoked.
 
 ----
-(:price 99 USD) ⇒ { amount: 99, currency: USD }
+(:price 99; USD) ⇒ { amount: 99, currency: USD }
 ----
 
 NOTE: We are careful to distinguish between the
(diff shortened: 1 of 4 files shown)


toddjonker:
So after spending most of Friday trying to figure out a workable void-placeholder to go with separators, my weekend’s musing has left me much less enthusiastic about separators on the whole.  I think I didn’t really consider the grammatical complexity, which effectively forces a brand-new third kind of sequence (beside list and sexp) with its own unique separation and quoting rules for both implementers and humans to keep track of.  I mean it looks and reads nicely, but that’s hiding a bunch of cruft that worries me.

Ultimately, if we are comparing separators versus implicit-rest for all */+ I’m back to preferring the latter.  Especially considering that it’s so close to what we’ve had all along, whereas separators are a big change both conceptually and syntactically.

Since @zslayton indicated that he's alright with sticking to the implicit-rest plan, I'm abandoning this in favor of #316

PR #326, opened 2024-06-11
Define Ion 1.1 system symbols with separate SID space

popematt:
### Issue #, if available:

None

### Description of changes:

* Proposes a list of symbols to be included in the Ion 1.1 system symbol table
* Proposes a way of encoding system symbols using a distinct SID space from user symbols
* Adds more descriptions of macros in the system module.
* Updates the System Symbol Table section of "what's new"

For System Symbols in `FlexSym`s, I noticed that we have nothing from `0x60` to `0xDF` that was already in use for any `FlexSym` escape codes, except for `$0` and the empty symbol. That range gives us 128 potential system symbol IDs, so I moved `$0` to `0x60` and I added `''` to the system symbol table so that we don't have to carve out a special case for it. System symbols are now encoded as a `FlexSym` by writing the SID as a one-byte FixedUint with a bias of `0x60`. E.g. if the byte after the `FlexSym` escape byte is `8E`, that refers to system symbol `2E` (decimal 46).

The alternatives are (a) assigning arbitrary, fragmented groups of byte values to system symbols to fit around existing values (b) moving some things, such as the "delimited container end" opcode in order to assign a contiguous block such as `0x80`-`0xFF`, or (c) requiring system symbols to be encoded with two bytes after the `FlexSym` escape byte—i.e.:`EF <ID>`. I deemed that (a) is ugly, (b) would require too much revising of other things, and (c) adds too much overhead.

I've included/kept logical types in (many of) the signatures of the system macros for now in order to convey the intention. We can go back and clean that up in a follow up PR.

I've tentatively left the logical types in the system symbol table because I can't remember if logical type checking is postponed or punted to Ion 1.2.

I also know that there are some incomplete sentences in the prose regarding system macros. I'll clean that up before merging this PR. I just wanted to get the symbol table and symbol encoding details out as soon as possible.

----

### DIFF
--- a/src/binary-encoding.adoc
+++ b/src/binary-encoding.adoc
@@ -213,7 +213,7 @@ pairs are spliced into the parent struct (TODO: Link)
               │    no more FlexInt bytes follow.
               │
 
-0 0 0 0 0 0 0 1   01100000
+0 0 0 0 0 0 0 1   01100100
 └─────┬─────┘     └───┬──┘
   2's comp.         0x64:
   zero              Indicates SSID 4
--- a/src/binary-encoding.adoc
+++ b/src/binary-encoding.adoc
@@ -162,8 +162,9 @@ represent the symbol’s text.
 * *exactly zero*, another byte follows that is an <<opcodes, opcode>>. The `FlexSym` parser is not responsible for
 evaluating this opcode, only returning it—the caller will decide whether the opcode is legal in the current context.
 Example usages of the opcode include:
-** Representing SID `$0` as `0xA0`.
-** Representing the empty string (`""`) as `0x90`.
+** Representing SID `$0` as `0x60`.
+** Representing system symbols (`0x61`-`0xDF`), where the system symbol ID is biased by `0x60`.
+*** Note that the empty symbol (i.e. the symbol `''`) is now a system symbol and can be referenced this way.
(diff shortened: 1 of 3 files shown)


src/binary-encoding.adoc
```diff
@@ -199,12 +200,25 @@ pairs are spliced into the parent struct (TODO: Link)
               │    no more FlexInt bytes follow.
               │
 
-0 0 0 0 0 0 0 1   10010000
+0 0 0 0 0 0 0 1   01100000
 └─────┬─────┘     └───┬──┘
-  2's comp.      opcode 0x90:
+  2's comp.      SSID 0x60:
```
tgregg: Unless the SSID acronym is introduced somewhere, let's just say "system SID".

src/binary-encoding.adoc
```diff
@@ -199,12 +200,25 @@ pairs are spliced into the parent struct (TODO: Link)
...
 ----
 
+.Figure {counter:figure}: `_FlexSym_` encoding of the system symbol `'name'`
+[%unbreakable]
+----
+              ┌─── The leading FlexInt ends in a `1`,
+              │    no more FlexInt bytes follow.
+              │
+
+0 0 0 0 0 0 0 1   01100000
```
```suggestion
0 0 0 0 0 0 0 1   01100100
```

popematt:
Closing in favor of #338

PR #329, opened 2024-06-27
Merge the Ion Semantic Model document in with the main spec build

toddjonker:
### Description of changes:

This pulls the denotational semantic model document (which we've reviewed in-house) into the public specification branch.

----


toddjonker:
Given the deep history here, I don't think we should squash-merge or rebase this; it should be a proper merge commit.  I'm not sure how to override the repository configuration to achieve that.

Perhaps one can just push directly and bypass the GitHub PR UI?

popematt:
The history will be preserved as part of this PR, even if it gets squashed when merging into the `ion-11-specification` branch. Is there a reason that is not enough?

If we are going to do a proper merge commit, then I think we should first push your changed to a new `ion-semantic-model` branch that is in `ion-docs` so that it's easier to follow the history within this repo.

toddjonker:
* History in a PR is not nearly as accessible as history in a DAG.  This matters a lot more off-web when browsing a local clone.
* Adding a short-term branch here won't have any effect on the DAG.  I'm not sure what that would accomplish, unless we want to work on that document separately, and then have to merge back onto the primary branch.  That seems like more trouble than its worth.
* I have a fair number of work-in-progress branches off several points of this history that will likely be harder to update and resume if the history becomes mangled.

toddjonker:
Can we merge this?  I don't think its a good idea for this spec to only be part of my personal fork.

popematt:
Yes. I'll temporarily lift the restrictions on merge commits so that we can merge it as is. We generally prefer not to have merge commits, so I'll be reinstating the restriction after this PR is merged so that we don't use merge commits on the `gh-pages` branch. (Or who knows—maybe GitHub lets you set up different rules for different branches now.)

popematt (approved):

toddjonker:
Thanks @popematt !

PR #333, opened 2024-08-14
Adds initial `ion-1-1` mdbook

zslayton:
This PR adds an `mdbook` port of our existing Asciidocs in a new folder called `books`. If we like the `mdbook` publishing experience we can eventually add others--cookbooks, a separate home for _Macros by example_, etc.

Note that this targets `gh-pages` rather than the `ion-11-specification`; this is in the hopes that we can incorporate it into our typical CI/CD build logic and have it auto-publish.

To view the rendered content locally, clone this branch and run:
```sh
cargo install mdbook
cargo install mdbook-alerts
mdbook serve
```
When you run `mdbook serve` it will run a local http server on port `3000` for you to explore the pages.

You can also view the rich diff for each page in GitHub using the button in the upper right of each file:
image

----
_**By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.**_

PR #334, opened 2024-08-20
Adds mdbook to jekyll build

popematt:
### Issue #, if available:

None. Builds on #333.

### Description of changes:

* Moves `books` to `_books` so that jekyll will watch the directory (when running `jekyll serve`) but not copy the book source directly into `_site`. (_This is the only change in the first commit of the PR._)
* Uses standard jekyll instead of GitHub jekyll so that we can use a custom plugin
  * This required changing some dependencies, so I also regenerated `Gemfile.lock`
* Adds a "deploy" GitHub actions workflow to deploy the site
* Set `.ruby-version` to make it easier to ensure that we're using the same ruby version locally and in the GH workflow
* Adds a generator plugin that runs `mdbook` and puts the Ion 1.1 draft spec into `_site` so that it gets deployed to GH Pages
* Adds an entry to `/docs` with a link to the draft spec.

Check it out at [popematt.github.io/ion-docs](https://popematt.github.io/ion-docs).

----


zslayton (approved):
🚀🚀🚀

PR #338, opened 2024-08-29
Define Ion 1.1 system symbols with separate SID space

popematt:
### Issue #, if available:

Supersedes #326

### Description of changes:

* Proposes a list of symbols to be included in the Ion 1.1 system symbol table
* Proposes a way of encoding system symbols using a distinct SID space from user symbols
* Adds more descriptions of macros in the system module.
* Updates the System Symbol Table section of "what's new"

For System Symbols in `FlexSym`s, I noticed that we have nothing from `0x60` to `0xDF` that was already in use for any `FlexSym` escape codes, except for `$0` and the empty symbol. That range gives us 128 potential system symbol IDs, so I moved `$0` to `0x60` and I added `''` to the system symbol table so that we don't have to carve out a special case for it. System symbols are now encoded as a `FlexSym` by writing the SID as a one-byte FixedUint with a bias of `0x60`. E.g. if the byte after the `FlexSym` escape byte is `8E`, that refers to system symbol `2E` (decimal 46).

The alternatives are (a) assigning arbitrary, fragmented groups of byte values to system symbols to fit around existing values (b) moving some things, such as the "delimited container end" opcode in order to assign a contiguous block such as `0x80`-`0xFF`, or (c) requiring system symbols to be encoded with two bytes after the `FlexSym` escape byte—i.e.:`EF <ID>`. I deemed that (a) is ugly, (b) would require too much revising of other things, and (c) adds too much overhead.

I've included/kept logical types in (many of) the signatures of the system macros for now in order to convey the intention. We can go back and clean that up in a follow up PR.

I've tentatively left the logical types in the system symbol table because I can't remember if logical type checking is postponed or punted to Ion 1.2.

There are still a few parts that are incomplete. I can add those now if you insist, but I was planning to follow up on that in a later PR.

----

### DIFF
--- a/_books/ion-1-1/src/binary/e_expressions.md
+++ b/_books/ion-1-1/src/binary/e_expressions.md
@@ -98,9 +98,30 @@ Address : 142918
 
 <div class="warning">
 
-> This section was obsolete and needs to be rewritten.
+> This section needs more details.
 
 </div>
+
+The opcode is `0xEE`. The macro address is given as a trailing [FlexUInt](primitives/fixed_uint.md) with no bias.
+
+
+## System Macro Invocations
+
+E-expression that invoke a system macro can be encoded using the `0xEF` opcode followed by a _positive_ 1-byte `FixedInt`.
+(Negative values are used for [system_symbols](values/symbol.md#system-symbols).)
+
+##### Encoding of the system macro `values`
+[%unbreakable]
+----
+┌──── Opcode 0xEF indicates a system symbol or macro invocation
+│  ┌─── FixedInt 0 indicates macro 0 from the system macro table
+│  │
+EF 00
+----
+
+In addition, system macros MAY be invoked using any of the `0x00`-`0x5F` or `0xEE` opcodes, provided that the macro being invoked has been given an address in user macro address space.
+
+<!-- TODO: Add link to "system-module" page. /-->
 
 ## Tagged E-expression Argument Encoding
 
--- a/_books/ion-1-1/src/binary/e_expressions.md
+++ b/_books/ion-1-1/src/binary/e_expressions.md
@@ -107,21 +107,19 @@ The opcode is `0xEE`. The macro address is given as a trailing [FlexUInt](primit
 
 ## System Macro Invocations
 
-E-expression that invoke a system macro can be encoded using the `0xEF` opcode followed by a _positive_ 1-byte `FixedInt`.
+E-expressions that invoke a [system macro](../modules/system_module.md#system-macro-addresses) can be encoded using the `0xEF` opcode followed by a _positive_ 1-byte `FixedInt`.
 (Negative values are used for [system_symbols](values/symbol.md#system-symbols).)
 
 ##### Encoding of the system macro `values`
-[%unbreakable]
-----
+```
 ┌──── Opcode 0xEF indicates a system symbol or macro invocation
 │  ┌─── FixedInt 0 indicates macro 0 from the system macro table
 │  │
 EF 00
-----
+```
 
 In addition, system macros MAY be invoked using any of the `0x00`-`0x5F` or `0xEE` opcodes, provided that the macro being invoked has been given an address in user macro address space.
-
-<!-- TODO: Add link to "system-module" page. /-->
+<!-- TODO: Add or link an example of how this can be done. /-->
 
 ## Tagged E-expression Argument Encoding
 
--- a/_books/ion-1-1/src/binary/primitives/flex_sym.md
+++ b/_books/ion-1-1/src/binary/primitives/flex_sym.md
@@ -9,17 +9,7 @@ A `FlexSym` begins with a [`FlexInt`](#flexint); once this integer has been read
 No more bytes follow.
 * **less than zero**, its absolute value represents a number of UTF-8 bytes that follow the `FlexInt`. These bytes
 represent the symbol’s text.
-* **exactly zero**, another byte follows that is an [opcode](../opcodes.md)
-or [system symbol](../../modules/system_module.md#system-symbols). The `FlexSym` parser is not responsible for
-evaluating this opcode, only returning it—the caller will decide whether the opcode is legal in the current context.
-Example usages of the opcode include:
-  * Representing SID `$0` as `0x60`.
-  * Representing system symbols (`0x61`-`0xDF`), where the system symbol ID is biased by `0x60`.
-    * Note that the empty symbol (i.e. the symbol `''`) is now a system symbol and can be referenced this way.
-  * When used to encode a struct field name, the opcode can invoke a macro that will evaluate to a struct whose key/value
-pairs are spliced into the parent [struct](../values/struct.md).
-    * Valid opcodes for invoking a macro are `0x00`-`0x5F`, `0xEE`, `0xEF`, and `0xF5`.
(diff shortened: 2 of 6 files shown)


_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -98,10 +98,31 @@ Address : 142918
...
+> This section needs more details.
 
 </div>
 
+The opcode is `0xEE`. The macro address is given as a trailing [FlexUInt](primitives/fixed_uint.md) with no bias.
+
+
+## System Macro Invocations
+
+E-expression that invoke a system macro can be encoded using the `0xEF` opcode followed by a _positive_ 1-byte `FixedInt`.
```
zslayton:
```suggestion
E-expressions that invoke a system macro can be encoded using the `0xEF` opcode followed by a _positive_ 1-byte `FixedInt`.
```

_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -98,10 +98,31 @@ Address : 142918
...
+(Negative values are used for [system_symbols](values/symbol.md#system-symbols).)
+
+##### Encoding of the system macro `values`
+[%unbreakable]
+----
```
zslayton: This is asciidoc syntax; the md spec uses
```
```ion
```
elsewhere.

_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -98,10 +98,31 @@ Address : 142918
...
+┌──── Opcode 0xEF indicates a system symbol or macro invocation
+│  ┌─── FixedInt 0 indicates macro 0 from the system macro table
+│  │
+EF 00
+----
+
+In addition, system macros MAY be invoked using any of the `0x00`-`0x5F` or `0xEE` opcodes, provided that the macro being invoked has been given an address in user macro address space.
```
zslayton, line 121: In one of the follow-on PRs we need to illustrate how this is done.

_books/ion-1-1/src/binary/primitives/flex_sym.md
```diff
@@ -45,8 +48,8 @@ pairs are spliced into the parent [struct](../values/struct.md).
               ┌─── The leading FlexInt ends in a `1`,
               │    no more FlexInt bytes follow.
               │
-0 0 0 0 0 0 0 1   10010000
+0 0 0 0 0 0 0 1   01110111
 └─────┬─────┘     └───┬──┘
-  2's comp.     opcode 0x90:
-  zero           empty symbol
+  2's comp.           FixedInt 0x77
+  zero                System SID 23, the empty symbol
```
zslayton: Nit: could you center the text on the stem of the `└───┬──┘` for consistency?

_books/ion-1-1/src/binary/primitives/flex_sym.md
```diff
@@ -9,14 +9,17 @@ A `FlexSym` begins with a [`FlexInt`](#flexint); once this integer has been read
 No more bytes follow.
 * **less than zero**, its absolute value represents a number of UTF-8 bytes that follow the `FlexInt`. These bytes
 represent the symbol’s text.
-* **exactly zero**, another byte follows that is an [opcode](opcodes.md). The `FlexSym` parser is not responsible for
+* **exactly zero**, another byte follows that is an [opcode](../opcodes.md)
+or [system symbol](../../modules/system_module.md#system-symbols). The `FlexSym` parser is not responsible for
```
zslayton: This is a bit confusing; I think we need a distinct name for this kind of byte. `FlexSymOpcode`?

_books/ion-1-1/src/binary/values/symbol.md
```diff
@@ -61,3 +61,22 @@ address that is decoded is biased by the number of addresses that can be encoded
...
+
+### System Symbols
+
+<!-- TODO: Add link to "system-module" page somewhere in this section. /-->
+
+System symbols (that is, symbols defined in the system module) can be encoded using the `0xEF` opcode followed by a _negative_ 1-byte `FixedInt`.
+(Positive values are used for [system macro invocations](../e_expressions.md#system-macro-invocations).)
+
+Unlike Ion 1.0, symbols are not required to use the lowest available SID for a given text.
+System symbols _MAY_ be encoded using other SIDs, but if an Ion data stream uses imported SIDs for system symbols in encoding directives, then the data may be inaccessible.
```
zslayton: > then the data may be inaccessible.

Here "the data" is the text of the SIDs, right? An alternative (less careful) reading would be "if you use non-system symbols for encoding directives, the reader won't be able to understand [all of] your data".

_books/ion-1-1/src/modules/system_module.md
```diff
@@ -0,0 +1,434 @@
...
+
+The symbols and macros of the system module `$ion` are available everywhere within an Ion document,
+with the version of that module being determined by the spec-version of each segment.
+The specific system symbols are largely uninteresting to users; while the binary encoding heavily
+leverages the system symbol table, the text encoding that users typically interact with does not.
+The system macros are more visible, especially to authors of macros.
+
+This chapter catalogs the system-provided symbols and macros.
+The examples below use unqualified names, which works assuming no other module exports the same
+name, but the unambiguous form `:$ion:macro-name` is always correct.
```
zslayton: I think this is ported text, but this should be `$ion::macro-name` now, right?

_books/ion-1-1/src/modules/system_module.md
```diff
@@ -0,0 +1,434 @@
...
+Like `make_string` but accepts lobs and produces a blob.
+
+#### `make_list`
+
+```ion
+(make_list (vals*)) -> list
+```
+
+Produces a non-null, unannotated list by concatenating the _content_ of any number of list or sexp inputs.
+The values `null.list` and `null.sexp` are treated as empty sequences.
```
zslayton: Can we forbid null arguments in all the `make_*` methods?

_books/ion-1-1/src/modules/system_module.md
```diff
@@ -0,0 +1,434 @@
...
+#### `make_struct`
+
+```ion
+(make_struct (structs*)) -> struct
+```
+
+Produces a non-null, unannotated struct by combining the fields of any number of structs.
+
+```ion
+(:make_struct { k1: 1, k2: 2} {k3:3} {k4: 4})  ⇒  {k1:1, k2:2, k3:3, k4:4}
```
zslayton, line 215: Nice.

zslayton (approved):
Thanks for porting this over. 👍

_books/ion-1-1/src/binary/values/symbol.md
popematt: The data is "all of your data".

Suppose you import some symbols, and `$123` is `$ion_encoding`, and it is imported from shared module Foo.
Then, you want to append some symbols, and you do this:
(There's some potential pseudo-code going on here, so don't get to concerned about the syntax.)
```
$123::(
    (import $123)
    (symbols $123 ["cat", "dog"])
)

// Lots more user data
```
If `Foo` is not available, then you can't even read the encoding directive here because you don't know that `$123` is `$ion_encoding`.

popematt: But I suppose that's no different than a macro being unavailable for a top-level e-expression.

zslayton (approved):

PR #342, opened 2024-09-27
Adds details about system macros and special forms

popematt:
### Issue #, if available:

None

### Description of changes:

Adds more details and examples for system macros and special forms.

----


_books/ion-1-1/src/macros/special_forms.md
```diff
@@ -0,0 +1,183 @@
...
+
+When a template-expression is syntactically an S-expression, its first element must be a symbol that matches either a 
+set of keywords denoting the special forms, or the name of a previously-defined macro.
+The interpretation of the S-expression’s remaining elements depends on how the symbol resolves.
+In the case of macro invocations, the elements following the operator are arbitrary template expressions, but for special forms that is not always the case.
+
+Special forms are "special" precisely because they cannot be expressed as macros and must therefore receive bespoke syntactic treatment.
+Since the elements of macro-invocation expressions are themselves expressions, when you want something to not be evaluated that way, it must be a special form.
+
+Finally, these special forms are part of the template language itself, and are not visible to encoded data;
```
zslayton:
```suggestion
Finally, these special forms are part of the template language itself, and are not addressable outside of TDL;
```

_books/ion-1-1/src/macros/special_forms.md
```diff
@@ -0,0 +1,183 @@
+## Special Forms
+
+When a template-expression is syntactically an S-expression, its first element must be a symbol that matches either a
```
zslayton: There's an empty (`TODO`) section in `defining_macros.md` that aims to introduce "TDL" as a well-defined name that we can reference later. 
image
We should figure out the best way to say "expression that can appear in TDL" (template expression? TDL expression?) and use it consistently throughout. For now, could you link to the empty TDL section?

_books/ion-1-1/src/macros/special_forms.md
```diff
@@ -0,0 +1,183 @@
...
+> to be a special form.
+> Candidates to be moved to system macros are `if_*` and `fail`.
+> Additionally, the system macro `parse_ion` may need to be classified as a special form since it only accepts literals.
+
+### `literal`
+
+```ion
+(macro USD_price (dollars) (.price dollars (.literal USD)))
+```
+In this template, we cannot write `(.price dollars USD)` because the symbol `USD` would be treated as an unbound variable reference and a syntax error, so we turn it into literal data by "escaping" it with `literal`. As an aside, there is no need for such a form in E-expressions, because in that context symbols and S-expressions are not "evaluated", and everything is literal except for E-expressions (which are not data, but encoding artifacts).
```
zslayton: The aside might make a nice `> [!TIP]`?

_books/ion-1-1/src/macros/special_forms.md
```diff
@@ -0,0 +1,183 @@
...
+
+### `if_none`
+
+```ion
+(macro if_none (stream* true_branch* false_branch*) /* Not representable in TDL */)
+```
+
+The `if_none` form is if/then/else syntax testing stream emptiness.
+It has three sub-expressions, the first being a stream to check. 
+If and only if that stream is empty (it produces no values), the second sub-expression is expanded and its results are returned by the `if_none` expression. Otherwise, the third sub-expression is expanded and returned.
```
zslayton: There are several references to "returning" values; I think it's somewhat more correct to say "expands to" or "produces" in the context of macros, though I don't feel very strongly about it.

_books/ion-1-1/src/macros/special_forms.md
```diff
@@ -0,0 +1,183 @@
...
+
+> [!Note]
+> Exactly one branch is expanded, because otherwise the empty `stream` might be used in a context that requires a value, resulting in an errant expansion error.
```
zslayton, line 46:
```suggestion
> Exactly one branch is expanded, because otherwise the empty `stream` might be used in a context that requires a value, resulting in an errant expansion.
```
or
```suggestion
> Exactly one branch is expanded, because otherwise the empty `stream` might be used in a context that requires a value, resulting in an expansion error.
```
?

_books/ion-1-1/src/macros/special_forms.md
```diff
@@ -0,0 +1,183 @@
...
+
+To refine things a bit further, trailing voidable arguments can be omitted entirely:
+```ion
+(:temperature 283) ⇒ {degrees:283, scale:K}
+```
+
+> [!TIP]
+> You can define a macro that wraps `if_none` to create a void-coalescing operator.
+> ```ion
+> (macro coalesce (maybe_none* else+) (.if_none maybe_none else maybe_none))
```
zslayton: I tripped on the name `else`; maybe `default_expr`?

_books/ion-1-1/src/macros/special_forms.md
```diff
@@ -0,0 +1,183 @@
...
+(:temperature 283 (::)) ⇒ {degrees:283, scale:K}
+```
+
+To refine things a bit further, trailing voidable arguments can be omitted entirely:
+```ion
+(:temperature 283) ⇒ {degrees:283, scale:K}
+```
+
+> [!TIP]
+> You can define a macro that wraps `if_none` to create a void-coalescing operator.
```
zslayton: s/`void`/`none`/ ?

_books/ion-1-1/src/macros/special_forms.md
```diff
@@ -0,0 +1,183 @@
...
+
+The `if_none` form is if/then/else syntax testing stream emptiness.
+It has three sub-expressions, the first being a stream to check. 
+If and only if that stream is empty (it produces no values), the second sub-expression is expanded and its results are returned by the `if_none` expression. Otherwise, the third sub-expression is expanded and returned.
+
+> [!Note]
+> Exactly one branch is expanded, because otherwise the empty `stream` might be used in a context that requires a value, resulting in an errant expansion error.
+
+```ion
+(macro temperature (degrees scale) {degrees: degrees, scale: (.if_none scale (.literal K) scale)})
```
zslayton: Subjectively, I think it helps readability to have non-trivial template expressions pretty-printed starting on the next line.

_books/ion-1-1/src/macros/special_forms.md
```diff
@@ -0,0 +1,183 @@
...
+If `expressions` evaluates to more than one value, it produces `true_branch`. Otherwise, it produces `false_branch`.
+Exactly one branch is evaluated. The `true_branch` and `false_branch` arguments are only evaluated if they are to be returned.
+
+### `for`
+
+```ion
+(for name_and_values template)
+```
+
+`name_and_values` is a list or s-expression containing one or more s-expressions of the form `(name value1 value2 ... valueN)`. The first value is a symbol to act as a variable name. The remaining values in the s-expression will be visited one at a time; for each value, expansion will produce a copy of the `template` argument expression with any appearance of the variable name replaced by the value.
```
zslayton: We'll need to revisit this if we decide to use `(%var)` syntax.

I think `(name value1 value2 ... valueN)` should be `(name expr1 expr2 ... exprN)` and an explanation that the `exprN` streams' output will be concatenated à la `.values`.

_books/ion-1-1/src/macros/special_forms.md
```diff
@@ -0,0 +1,183 @@
...
+
+For example:
+
+```ion
+(:for
```
zslayton: Here and below, I think this should be `.for` since you can't use the special form as an e-expression.

_books/ion-1-1/src/macros/system_macros.md
```diff
@@ -0,0 +1,398 @@
+## System Macros
+
+Many of the system macros _MAY_ be defined as template macros, and when possible, the specification includes a template.
+Templates are given here as normative example, but system macros are not required to be implemented as template macros.
+
+The macros that can be defined as templates are included as system macros because of their broad applicability, and
+so that Ion implementations can provide optimizations for these macros that can short-circuit the regular macro evaluation.
```
zslayton: I think an example would clarify the meaning of "short-circuit" here.

_books/ion-1-1/src/macros/system_macros.md
```diff
@@ -0,0 +1,398 @@
...
+```
+
+This is, essentially, the identity function. 
+It produces a stream from any number of arguments, concatenating the streams produced by the nested expressions.
+Used to aggregate multiple values or sub-streams to pass to a single argument, or to return multiple results.
+
+#### `flatten`
+
+```ion
+(macro flatten (sequence+) /* Not representable in TDL */)
```
zslayton: Should this take `(sequence*)` so potentially empty streams can be passed in?

_books/ion-1-1/src/macros/system_macros.md
```diff
@@ -0,0 +1,398 @@
...
+
+> [!IMPORTANT]
+> Unlike most macros, this macro specifically requires _literals_. Macros are not allowed to contain recursive calls,
+> and composing an embedded document from multiple expressions would make it possible to implement recursion in the
+> macro system.
+> 
+> Allowing context to leak from the outer scope into the document being parsed would also enable recursion.
+>
+> Furthermore, `parse_ion` cannot be invoked from a macro body (i.e. in Template Definition Language (TDL)). 
+> (Why not?)
```
zslayton: I think it _could_ be--it isn't inherently a problem--but given that it requires literals, I'm unsure it offers much value since you could always write the equivalent stream in TDL directly.

_books/ion-1-1/src/macros/system_macros.md
```diff
@@ -0,0 +1,398 @@
...
+
+
+### Value Constructors
+
+#### `annotate`
+
+```ion
+(macro annotate (ann* value) /* Not representable in TDL */)
```
zslayton, line 102: If we decide to add footnotes, it might be helpful to explain that the annotations sequence comes first because it parallels how annotations are read from the data stream.

_books/ion-1-1/src/macros/system_macros.md
```diff
@@ -0,0 +1,398 @@
...
+
+#### `make_sexp`
+
+```ion
+(macro make_sexp (sequences*) ( (.flatten sequences) ))
+```
+
+Like `make_list` but produces a sexp.
+This is the only way to produce an S-expression from a template: unlike lists, S-expressions in
+templates are not quasi-literals.
```
zslayton: I think this has been obviated by `.` syntax. Together with `flatten`, this also raises the question: do we need `make_list`/`make_sexp`?

_books/ion-1-1/src/macros/system_macros.md
```diff
@@ -0,0 +1,398 @@
...
+(:make_sexp (1 2))            => (1 2)
+(:make_sexp (1 2) [3, 4])     => (1 2 3 4)
+(:make_sexp ((1 2)) [[3, 4]]) => ((1 2) [3, 4])
+```
+
+#### `make_struct`
+
+```ion
+(macro make_struct (structs*) /* Not representable in TDL */)
+```
```
zslayton, line 171: `make_struct` (should) make it possible to provide dynamically constructed field names. As the signature stands, I don't think it can.

We need another built-in macro like this:
```
(macro field (flex_sym::name, value) ...)
```

Such that I can do:
```
(.field (.make_symbol "foo" bar 'baz') 7)

// =>

{foobarbaz: 7}
```

`make_struct` can then have the signature:
```
(macro make_struct (field::fields*) ...)
```
which enables syntax like:
```
(.make_struct
  (name Alice)
  (age 44)
  (
    (.make_symbol "favorite_" "ice_cream") 
    "vanilla"
  )
)
```

_books/ion-1-1/src/macros/system_macros.md
```diff
@@ -0,0 +1,398 @@
...
+
+```ion
+(macro make_decimal (flex_int::coefficient flex_int::exponent) /* Not representable in TDL */)
+```
+
+This is no more compact than the regular binary encoding for decimals.
+However, it can be used in conjunction with other macros, for example, to represent fixed-point numbers.
+
+```ion
+(macro usd (cents) (annotate (literal USD) (make_decimal cents -2))
```
zslayton:
```suggestion
(macro usd (cents) (.annotate (.literal USD) (.make_decimal cents -2))
```

_books/ion-1-1/src/macros/system_macros.md
```diff
@@ -0,0 +1,398 @@
...
+```ion
+(macro delta (flex_int::initial! flex_int::deltas+) /* Not representable in TDL */)
+```
+
+Example:
+```ion
+(:delta 10 1 2 3 -4) => 11 13 16 12
+```
+
+#### `sum`
```
zslayton, line 288: I'm struggling to remember whether we had a use case for `sum` besides biased integers (which `delta` enables and more)

_books/ion-1-1/src/modules/system_module.md
```diff
@@ -10,8 +10,20 @@ This chapter catalogs the system-provided symbols and macros.
...
 
-> [!WARNING]
-> This list is not complete. We expect it to grow and evolve as we gain experience writing macros.
+### Relation to Local Symbol and Macro Tables
+
+In Ion 1.0, the system symbol table is always the first import of the local symbol table.
+However, in Ion 1.1, the system symbol and macro tables have a system address space that is distinct from the local address space.
+When starting an Ion 1.1 segment (i.e. immediately after encountering an `$ion_1_1` version marker), the local symbol
+table is prepopulated with the system symbols. 
+(Why? They can also be used via opcode `EE` which requires the same number of bytes as the `E1` OpCode.)
```
zslayton: This is a good callout! There may not be a strong reason to pre-populate the symbol table. It lets you think of the initial state of the encoding module as "the system module," but that's not buying us very much.

_books/ion-1-1/src/modules/system_module.md
```diff
@@ -10,8 +10,20 @@ This chapter catalogs the system-provided symbols and macros.
...
+When starting an Ion 1.1 segment (i.e. immediately after encountering an `$ion_1_1` version marker), the local symbol
+table is prepopulated with the system symbols. 
+(Why? They can also be used via opcode `EE` which requires the same number of bytes as the `E1` OpCode.) 
+The local macro table is also prepopulated with the system macros.
+(This time it makes sense because we're actually saving a byte.)
+However, the system symbols and macros are not permanent fixtures of the local symbol and macro tables respectively.
+
+
+When a local macro has the same name as a system macro, it shadows the system macro.
+It is not possible to invoke the system macro unless that system macro has been given an alias in the local macro table.
```
zslayton: This line is a bit fuzzy to me. In TDL, if `add_symbols` gets shadowed I can still refer to it via `$ion::add_symbols`.  I think this is saying you can't invoke a system macro _via an e-expression_ if it was shadowed in the active macro table. If that's the case you could you make that context explicit?

_books/ion-1-1/src/modules/system_module.md
```diff
@@ -39,18 +51,19 @@ The Ion 1.1 System Symbol table _replaces_ rather than extends the Ion 1.0 Syste
 | 14 | `macro_table`                   |
 | 15 | `symbol_table`                  |
 | 16 | `module`                        |
-| 17 | `retain`                        |
-| 18 | `export`                        |
-| 19 | `catalog_key`                   |
+| 17 | ~~retain~~                      |
+| 18 | export                          |
```
zslayton: Did you mean to remove the backticks for this one?

_books/ion-1-1/src/modules/system_module.md
```diff
@@ -39,18 +51,19 @@ The Ion 1.1 System Symbol table _replaces_ rather than extends the Ion 1.0 Syste
...
+| 19 | ~~catalog_key~~                 |
 | 20 | `import`                        |
-| 21 | _&lt;empty string>_ (i.e. `''`) |
+| 21 | _zero-length text_ (i.e. `''`)  |
```
zslayton, line 58: Remind me when/where this gets used? It's two bytes to refer to the empty symbol vs `0xA0` to refer to a symbol with inline empty text.

_books/ion-1-1/src/modules/system_module.md
```diff
@@ -39,18 +51,19 @@ The Ion 1.1 System Symbol table _replaces_ rather than extends the Ion 1.0 Syste
 | 14 | `macro_table`                   |
 | 15 | `symbol_table`                  |
 | 16 | `module`                        |
-| 17 | `retain`                        |
-| 18 | `export`                        |
-| 19 | `catalog_key`                   |
+| 17 | ~~retain~~                      |
```
zslayton: I think leaving these in the table doesn't buy us anything unless we're also going to have a note explaining why they were removed. That feels more like a github issue to open/close though, since the spec itself doesn't need to contain its own history.

_books/ion-1-1/src/macros/system_macros.md
popematt: Yes, but `make_field`, perhaps?

_books/ion-1-1/src/macros/system_macros.md
popematt: > I'm unsure it offers much value

~~Actually, I think it should be invokable from the data stream because it enables things like:~~
* ~~Embedding text Ion into binary Ion and vice versa without having to parse and re-encode the original stream~~
* ~~Embed one binary Ion stream into another by starting a `parse_ion` invocation and then doing a plain old memcopy of the old stream into the buffer of the new stream.~~
* ~~There might be situations where having encoding contexts that are contained in a narrow scope could be useful.~~

I misinterpreted your comment.

_books/ion-1-1/src/macros/system_macros.md
```diff
@@ -0,0 +1,398 @@
...
+(:add_symbols foo bar)
+=>
+$ion_encoding::(
+  (symbol_table $ion_encoding [foo, bar])
+  (macro_table $ion_encoding)
+)
+```
+
+#### `set_macros`
+Sets the local symbol table, preserving any symbols in the symbol table.
```
tgregg:
```suggestion
Sets the local macro table, preserving any symbols in the symbol table.
```

_books/ion-1-1/src/macros/special_forms.md
popematt: TDL clause?

_books/ion-1-1/src/macros/system_macros.md
popematt: Actually, on second thought, if you want to embed another data stream into a macro without having to parse and re-encode it, then `parse_ion` _could_ be useful in TDL.

Edit—so I'm going to change this to allow `parse_ion` to be invoked from TDL. I'd rather have a questionably useful case than introduce an edge case where there's _one_ macro that is not invocable from TDL.

_books/ion-1-1/src/macros/system_macros.md
popematt: > ```
> (macro make_struct (field::fields*) ...)
> ```

If we have a macro-shaped `fields` parameter, then `make_struct` can't be called with regular structs as its arguments. That seems like something we don't want to do... but in Ion binary, the `(flex_sym::field_name value)` encoding is identical to the encoding of a struct with `FlexSym` field names, so maybe it doesn't really matter?

popematt: I'll add `make_field` regardless.

_books/ion-1-1/src/modules/system_module.md
popematt: If we had some more system macros that could make it easier/more efficient to write an encoding directive, then I could see how useful it would be to have all of those macros pre-populated into the `00` - `3F` range.

I think, though, that the main benefit is for implementers. When you have a partially complete implementation, it's a lot easier to start testing things if you have some symbols and macros that are automatically populated into their respective user-spaces.

_books/ion-1-1/src/modules/system_module.md
popematt: Nope. Thanks for catching that.

_books/ion-1-1/src/modules/system_module.md
popematt: It's needed for the FlexSym encoding of `''`. We can't use `A0`/`90` for empty & unknown symbol in FlexSyms anymore because they're in the range that we've claimed for system symbols. Now we use `0` in the system symbol range for unknown text, and this symbol for the empty string symbol text. Of course, we can revisit the exact ID of this symbol if it would be helpful to do so.

_books/ion-1-1/src/macros/special_forms.md
```diff
@@ -112,20 +124,23 @@ Exactly one branch is evaluated. The `true_branch` and `false_branch` arguments
...
 
 ### `for`
 
 ```ion
-(for name_and_values template)
+(for name_and_expressions template)
 ```
 
-`name_and_values` is a list or s-expression containing one or more s-expressions of the form `(name value1 value2 ... valueN)`. The first value is a symbol to act as a variable name. The remaining values in the s-expression will be visited one at a time; for each value, expansion will produce a copy of the `template` argument expression with any appearance of the variable name replaced by the value.
+`name_and_values` is a list or s-expression containing one or more s-expressions of the form `(name expr0 expr1 ... exprN)`.
```
zslayton:
```suggestion
`name_and_expressions` is a list or s-expression containing one or more s-expressions of the form `(name expr0 expr1 ... exprN)`.
```

_books/ion-1-1/src/macros/system_macros.md
```diff
@@ -177,6 +178,14 @@ Produces a non-null, unannotated struct by combining the fields of any number of
   {k4: 4})        =>  {k1:1, k2:2, k3:3, k4:4}
 ```
 
+#### `make_field`
+
+```ion
+(macro make_field (flex_sym::field_name value) /* Not representable in TDL */)
+```
+
+Produces a non-null, unannotated, single-field struct using the given field name and value.
```
zslayton, line 189: Maybe add a line saying that this can be used to construct field names at runtime?

_books/ion-1-1/src/macros/special_forms.md
zslayton: I think I prefer "TDL expression." To me, "clause" implies a known/expected section of something but the template is basically free-form.

_books/ion-1-1/src/modules/system_module.md
zslayton: > If we had some more **_system macros_** that could make it easier/more efficient to write an encoding directive, then I could see how useful it would be to have all of those macros pre-populated into the 00 - 3F range.

Do you mean system symbols?

_books/ion-1-1/src/macros/system_macros.md
zslayton: I'm ok with leaving it as `make_struct (structs*)` 👍. The important thing is that we have a way to make a dynamic field name in TDL, which `make_field` provides.

_books/ion-1-1/src/macros/system_macros.md
```diff
@@ -0,0 +1,417 @@
...
+```ion
+[1, 2, (:flatten [a, b]), 3, 4] => [1, 2, a, b, 3, 4]
+```
+
+#### `parse_ion`
+
+Ion documents may be embedded in other Ion documents using the `parse_ion` macro.
+
+```ion
+(macro parse_ion (uint8::data*) /* Not representable in TDL */)
```
zslayton, line 60: I suggest using a tagged encoding for `data`; I think using `uint8` would cause tagless encoding domain constraints to apply, making this weird for text inputs.

For binary it looks pretty nice though!
```
(.parse_ion 
  (..
    0xE0 0x01 0x01 0xEA    // $ion_1_1
    0xA3 0x66 0x6F 0x6F    // foo
  )
)
```

This does make me want hex blob syntax 😢

zslayton (approved):
A few things to consider before merging.

_books/ion-1-1/src/macros/system_macros.md
popematt: Thanks for catching that. That was an idea I was toying with, but I didn't intend for that to make it into the PR.

_books/ion-1-1/src/modules/system_module.md
popematt: No, I mean system macros. It wouldn't make a big difference in Ion text, but in Ion binary, having something like these could make a big difference if you have a lot of macros with relatively small bodies.
```
// Assume macro addresses starting at 20
(macro macro (flex_sym::name? parameters* body) ((.literal macro) name (parameters) body))
(macro u8_param (name) (.annotate (.. (.literal u8)) name))
(macro zero_or_one (p) (.values p (.literal ?)))
```
Let's compare some ways of declaring a macro. I'm using `null` as the macro body because it doesn't make a big deal in this particular example. (A simple `{x:x, y:y, z:z}` body would be 14 bytes, assuming that none of the field names are in the symbol table yet.) I'm using `'<char>` to represent the ascii byte for a character (I don't want to convert all of it to hex).

| Macro definition | Size | Raw Bytes |
|:---|:---|:---|
| `(macro point3d`<br/>`       (uint8::x? uint8::y? uint8::z?)`<br/>`       null)` | 36 | `F2 EE 0D A7 'p 'o 'i 'n 't '3`<br/>`'d F2 E7 01 95 A1 'x A1 '? E7`<br/>`01 95 A1 'y A1 '? E7 01 95 A1`<br/>`'z A1 '? F0 EA F0` |
| `(:macro point3d`<br/>`        (:: uint8::x? uint8::y? uint8::z?)`<br/>`        null)`  |  33  | `20 06 F9 'p 'o 'i 'n 't '3 'd`<br/>`2B E7 01 95 A1 'x A1 '? E7 01`<br/>`95 A1 'y A1 '? E7 01 95 A1 'z`<br/>`A1 '? EA`
| `(:macro point2d`<br/>`        (:: (:zero_or_one (:u8_param x))`<br/>`            (:zero_or_one (:u8_param y))`<br/>`            (:zero_or_one (:u8_param z)))`<br/>`        null)` | 24 | `20 06 F9 'p 'o 'i 'n 't '3 'd`<br/>`19 22 21 FF 'x 22 21 FF 'y 22`<br/>`21 FF 'z EA`


Similarly, macros that produce TDL invocations, such as this, could add even more savings.
```
(macro if_some (a* b* c*) (.if_some a b c))
```
Encoding `(.if_some a b c)` would require 11 bytes, whereas `(:if_some a b c)` would only require 7-8 bytes (depending on whether `if_some` has a 1 or 2 byte address). Similar savings could be gotten for every other system macro and special form. However, this would make it less user-friendly because using E-Expressions in TDL to produce more TDL adds to the cognitive burden, and it might not be feasible—I haven't considered macro hygiene at all in this example.

Since it seems like we're going to be using `.` for TDL macro invocations, even just having a `(:dot)` macro could save a byte every time we call a macro from TDL. (The symbol `.` always requires 2 bytes, whereas `(:dot)` could have a 1-byte address.)

zslayton: I see, thanks for explaining!

_books/ion-1-1/src/macros/system_macros.md
popematt: I'd be hesitant to remove them, but that's partly just because I haven't thought about it much to determine whether there is any use case for them that isn't solved in other ways.

_books/ion-1-1/src/macros/system_macros.md
popematt: I don't remember the use case either, but `sum` only produces one integer vs `delta` producing `N-1` integers, so `delta` can't entirely replace `sum`.

zslayton: I believe we started with `sum` to enable biased integers but then realized `delta` would achieve that _and_ delta encoding.

For serialization, I can't think of a distinct use case for `sum`. Writing a sequence of numbers to mean their total will always take more bytes.

Issue #343, opened 2024-09-30
TDL Syntax is excessively verbose and hard to read for symbols and s-expressions


popematt:
Any macro that uses symbol literals or s-expressions is really verbose and difficult to read. For example, this is a macro that appends macros to the local macro table.
```
(macro append_macros ($macro_definitions*)
    (annotate
        (literal $ion_encoding)
        (make_sexp [
            (literal (symbol_table $ion_encoding)),
            (make_sexp [
                (literal macro_table),
                (literal $ion_encoding),
                $macro_definitions
            ])
        ])
    )
)
```
I have also noticed that early users who are writing macros are falling into a convention of writing variable names with a `$` so that they are more visible.

It would be nice if we could avoid having to escape symbols and s-expressions so that any sort of macro that emits encoding directives or TDL won't be so excessively difficult to read.

zslayton:
If we follow the approach described in #344 (changing TDL invocation syntax from `(foo ...)` to `(.foo ...)`, we could also modify the syntax for variable expansion. Rather than treating all symbols in the template as variable names (and raising an error if they're not), we could use a syntax like `(%var_name)` to indicate expansion.

I'll copy/paste the example above for easier comparison.

**Before**
```
(macro append_macros ($macro_definitions*)
    (annotate
        (literal $ion_encoding)
        (make_sexp [
            (literal (symbol_table $ion_encoding)),
            (make_sexp [
                (literal macro_table),
                (literal $ion_encoding),
                $macro_definitions
            ])
        ])
    )
)
```
**After**
```
(macro append_macros (macro_definitions*)
    $ion_encoding::(
        (symbol_table $ion_encoding)
        (macro_table $ion_encoding (%macro_definitions))
    )
)
```

As a nice side effect, it means symbol values no longer behave differently from field names and annotations, which has been another stumbling block for newcomers.

Combined, these two changes would eliminate the need for `literal` altogether. To write an s-expression that starts with `.` or `%`, you can use the `.values` macro:
```
((.values %) some_symbol)
((.values .) some_symbol)
```

popematt:
Resolved in #354

PR #348, opened 2024-10-08
Updates the binary encoding and parameter cardinality sections

zslayton:
```
cargo install mdbook
cd ion-docs
git switch revamped-binary-enc
cd _books/ion-1-1
mdbook serve
```

----

### DIFF
--- a/_books/ion-1-1/src/binary/e_expressions.md
+++ b/_books/ion-1-1/src/binary/e_expressions.md
@@ -179,8 +179,8 @@ When the macro has a single `exactly-one` parameter, the corresponding encoded a
 ```
 ┌──── Opcode 0x00 is less than 0x40; this is an e-expression invoking
 │     the macro at address 0.
-│  ┌─── Argument 'x': opcode 0x61 indicates a 1-byte integer
-│  │   1
+│    ┌─── Argument 'x': opcode 0x61 indicates a 1-byte integer (1)
+│  ┌─┴─┐
 00 61 01
 ```
 
@@ -189,8 +189,8 @@ When the macro has a single `exactly-one` parameter, the corresponding encoded a
 ```
 ┌──── Opcode F4: An e-expression with a trailing FlexUInt address
 │  ┌──── FlexUInt 0: Macro address 0
-│  │  ┌─── Argument 'x': opcode 0x61 indicates a 1-byte integer
-│  │  │   1
+│  │    ┌─── Argument 'x': opcode 0x61 indicates a 1-byte integer (1)
+│  │  ┌─┴─┐
 F4 01 61 01
 ```
 
@@ -218,10 +218,10 @@ bytes in the stream. When no more parameters remain, parsing of the e-expression
 ```
 ┌──── Opcode 0x00 is less than 0x40; this is an e-expression
 │     invoking the macro at address 0.
-│  ┌─── Argument 'a': opcode 0x61 indicates a 1-byte integer
-│  │      ┌─── Argument 'b': opcode 0x61 indicates a 1-byte integer
-│  │      │    ┌─── Argument 'c': opcode 0x61 indicates a 1-byte integer
-│  │   1  │  2 │   3
+│    ┌─── Argument 'a': opcode 0x61 indicates a 1-byte integer (1)
+│    │     ┌─── Argument 'b': opcode 0x61 indicates a 1-byte integer (2)
+│    │     │     ┌─── Argument 'c': opcode 0x61 indicates a 1-byte integer (3)
+│  ┌─┴─┐ ┌─┴─┐ ┌─┴─┐
 00 61 01 61 02 61 03
 ```
 
@@ -229,11 +229,11 @@ bytes in the stream. When no more parameters remain, parsing of the e-expression
 ```
 ┌──── Opcode F4: An e-expression with a trailing FlexUInt address
 │  ┌──── FlexUInt 0: Macro address 0
-│  │  ┌─── Argument 'a': opcode 0x61 indicates a 1-byte integer
-│  │  │     ┌─── Argument 'b': opcode 0x61 indicates a 1-byte integer
-│  │  │     │     ┌─── Argument 'c': opcode 0x61 indicates a 1-byte integer
-│  │  │     │     │
-│  │  │   1 │   2 │   3
+│  │    ┌─── Argument 'a': opcode 0x61 indicates a 1-byte integer (1)
+│  │    │     ┌─── Argument 'b': opcode 0x61 indicates a 1-byte integer (2)
+│  │    │     │     ┌─── Argument 'c': opcode 0x61 indicates a 1-byte integer (3)
+│  │    │     │     │
+│  │  ┌─┴─┐ ┌─┴─┐ ┌─┴─┐
 F4 01 61 01 61 02 61 03
 ```
 
@@ -271,7 +271,7 @@ information in their serialized form.
 
 As first demonstrated in _[Encoding multiple exactly-one arguments](#encoding-multiple-exactly-one-arguments)_,
 the bytes of the serialized arguments begin immediately after the opcode and (if separate) the macro address.
-The reader iterates over the parameters declared in the macro signature from left to right. For each parameter,
+The reader iterates over the parameters in the macro signature in the order they are declared. For each parameter,
 the reader uses the parameter's declared encoding to interpret the next bytes in the stream. When no more parameters
 remain, parsing is complete.
 
@@ -361,13 +361,28 @@ in the body of the macro.
 
 ```ion
 (:add_macros
-    (pi () 3.14159265) // A constant `pi`
+  // Defines a constant 'hostname'
+  (hostname () "abc123.us_west.example.com")
 
-    (circle_area (pi::Pi radius) (multiply Pi radius radius))
-    //            └── ERROR: cannot use a constant as a macro shape
+  (http_ok (hostname::server page)
+  //           └── ERROR: cannot use a constant as a macro shape
+     {
+        server: (%server),
+        page: (%page),
+        message: OK,
+        status: 200,
+     }
+  )
 
-    (circle_area (radius) (multiply (.pi) radius radius))
-    //   OK: invokes `pi` as needed ──┘
+  (http_ok (page)
+    {
+      server: (.hostname),
+      //           └── OK: invokes constant as needed
+      page: (%page),
+      message: OK,
+      status: 200,
+    }
+  )
 )
 ```
 
@@ -405,61 +420,50 @@ has been read.
 |-----------------------|-----------------------------|
 | `()`                  | _&lt;No variadics, no AEB>_ |
 | `(a b c)`             | _&lt;No variadics, no AEB>_ |
-| `(a b c?)`            | `0b------cc`                |
-| `(a b* c?)`           | `0b----ccbb`                |
-| `(a+ b* c?)`          | `0b--ccbbaa`                |
-| `(a+ b c?)`           | `0b----ccaa`                |
-| `(a+ b* c? d*)`       | `0bddccbbaa`                |
-| `(a+ b* c? d* e)`     | `0bddccbbaa`                |
-| `(a+ b* c? d* e f?)`  | `0bddccbbaa 0b------ff`     |
-| `(a+ b* c? d* e+ f?)` | `0bddccbbaa 0b----ffee`     |
+| `(a b c?)`            | `------cc`                  |
+| `(a b* c?)`           | `----ccbb`                  |
+| `(a+ b* c?)`          | `--ccbbaa`                  |
+| `(a+ b c?)`           | `----ccaa`                  |
+| `(a+ b* c? d*)`       | `ddccbbaa`                  |
+| `(a+ b* c? d* e)`     | `ddccbbaa`                  |
+| `(a+ b* c? d* e f?)`  | `ddccbbaa ------ff`         |
+| `(a+ b* c? d* e+ f?)` | `ddccbbaa ----ffee`         |
 
 Each pair of bits in the AEB indicates what kind of expression to expect in the corresponding argument position.
 
-| Bit sequence | Meaning                                                                                                                           |
-|:------------:|:----------------------------------------------------------------------------------------------------------------------------------|
-|     `00`     | An **[empty expression group](#empty-groups)**. No bytes are present in the corresponding argument position.                      |
-|     `01`     | A **single expression** of the declared encoding is present in the corresponding argument position.                               |
-|     `10`     | A **[populated expression group](#populated-groups)** of the declared encoding is present in the corresponding argument position. |
-|     `11`     | _Reserved_. A bitmap entry with this bit sequence is illegal in Ion 1.1.                                                          |
+| Bit sequence | Meaning                                                                                                                  | `?` | `*` | `+` |
+|:------------:|:-------------------------------------------------------------------------------------------------------------------------|:---:|:---:|:---:|
+|     `00`     | An **empty stream**. No bytes are present in the corresponding argument position.                                        |  ✅  |  ✅  |  ❌  |
+|     `01`     | A **single expression** of the declared encoding is present in the corresponding argument position.                      |  ✅  |  ✅  |  ✅  |
+|     `10`     | A **[expression group](#expression-groups)** of the declared encoding is present in the corresponding argument position. |  ❌  |  ✅  |  ✅  |
+|     `11`     | _Reserved_. A bitmap entry with this bit sequence is illegal in Ion 1.1.                                                 |  ❌  |  ❌  |  ❌  |
+
+As noted in the table above:
+* An empty stream (`00`) cannot be used to encode an argument for a parameter with a cardinality of `one-or-more`.
+* An expression group (`10`) cannot be used to encode an argument for a parameter with a cardinality of `zero-or-one`.
 
 #### Expression groups
 
-##### Empty groups
+This section describes the encoding of an expression group. For an explanation of what an expression group is and how to use it,
+see _[Expression groups](../todo.md)_.
 
-If a parameter has a cardinality of `zero-or-one` or `zero-or-more`, callers can choose to pass an
-empty expression group as the corresponding argument. This is done by setting the bit sequence `00`
-in the appropriate position in the [Argument Encoding Bitmap](#argument-encoding-bitmap-aeb).
-
-##### Populated groups
-
-If a parameter has a cardinality of `zero-or-more` or `one-or-more`, callers can choose to pass a
-populated expression group as the corresponding argument instead of a single expression. This is
-done by setting the bit sequence `10` in the appropriate position in the
-[Argument Encoding Bitmap](#argument-encoding-bitmap-aeb) and then writing an
-[expression sequence](#expression-sequence) in the corresponding argument position to indicate
-how many bytes part of this group.
-
-##### Expression sequence
-
-An _expression sequence_ begins with a [`FlexUInt`](primitives/flex_uint.md). If the `FlexUInt`'s value
+An expression group begins with a [`FlexUInt`](primitives/flex_uint.md). If the `FlexUInt`'s value
 is:
-* **greater than zero**, then it represents the number of bytes used to encode this expression sequence. The reader
+* **greater than zero**, then it represents the number of bytes used to encode the rest of the expression group. The reader
   should continue reading expressions of the declared encoding until that number of bytes has been consumed.
-* **zero**, then it indicates that this is a delimited expression sequence and the processing varies according to
+* **zero**, then it indicates that this is a delimited expression group and the processing varies according to
   whether the declared encoding is tagged or tagless. If the encoding is:
-  * a **[tagged](#tagged-encoding)**, then each expression in the group begins with an opcode. The reader
+  * **[tagged](#tagged-encoding)**, then each expression in the group begins with an opcode. The reader
     must consume tagged expressions until it encounters a terminating `END` opcode (`0xF0`).
-  * a **[tagless](#tagless-encodings)**, then each expression in the group has no leading opcode; there is no
+  * **[tagless](#tagless-encodings)**, then each expression in the group has no leading opcode; there is no
     way to encode the terminating `END`. Instead, the sequence is broken into 'chunks' that each have a
     `FlexUInt` length prefix. The reader will continue reading chunks until it encounters a length prefix of
     `FlexUInt` `0`, indicating the end of the chunk series. Each chunk in the series must be self-contained;
     an expression of the declared encoding may not be split across multiple chunks.
 
-> [!Note]
-> Despite the name, it is possible to encode an empty expression group using this syntax.
-> However, doing so will always be significantly less efficient than using the
-> [Argument Encoding Bitmap](#argument-encoding-bitmap-aeb).
+> [!TIP]
+> While it is legal to write an empty expression group for `zero-or-more` parameters,
+> it is always more efficient to set the parameter's AEB bits to `00` instead.
 
 #### Example encoding of tagged `zero-or-one` with empty group
 
@@ -541,7 +545,7 @@ is:
 00 01 61 01
 ```
 
-#### Example encoding of tagged `zero-or-more` with populated group
+#### Example encoding of tagged `zero-or-more` with expression group
 ```ion
 (:add_macros
   (foo (a*) ...)
@@ -555,18 +559,18 @@ is:
 ```
 ┌──── Opcode 0x00 is less than 0x40; this is an e-expression
 │     invoking the macro at address 0.
-│  ┌──── AEB: 0b------aa; a=10, populated expression group
-│  │  ┌──── FlexUInt 6: 6-byte expression sequence
+│  ┌──── AEB: 0b------aa; a=10, expression group
+│  │  ┌──── FlexUInt 6: 6-byte expression group
 │  │  │    ┌──── Opcode 0x61 indicates a 1-byte int (1)
 │  │  │    │     ┌──── Opcode 0x61 indicates a 1-byte int (2)
 │  │  │    │     │     ┌─── Opcode 0x61 indicates a 1-byte int (3)
 │  │  │  ┌─┴─┐ ┌─┴─┐ ┌─┴─┐
 00 02 0D 61 01 61 02 61 03
          └───────┬───────┘
-      6-byte expression sequence
+      6-byte expression group body
 ```
 
-#### Example encoding of tagged `zero-or-more` with delimited group
+#### Example encoding of tagged `zero-or-more` with delimited expression group
 ```ion
 (:add_macros
   (foo (a*) ...)
@@ -580,8 +584,8 @@ is:
 ```
 ┌──── Opcode 0x00 is less than 0x40; this is an e-expression
 │     invoking the macro at address 0.
-│  ┌──── AEB: 0b------aa; a=10, populated expression group
-│  │  ┌──── FlexUInt 0: delimited expression sequence
+│  ┌──── AEB: 0b------aa; a=10, expression group
+│  │  ┌──── FlexUInt 0: delimited expression group
 │  │  │    ┌──── Opcode 0x61 indicates a 1-byte int (1)
 │  │  │    │     ┌──── Opcode 0x61 indicates a 1-byte int (2)
 │  │  │    │     │     ┌─── Opcode 0x61 indicates a 1-byte int (3)
@@ -589,7 +593,7 @@ is:
 │  │  │  ┌─┴─┐ ┌─┴─┐ ┌─┴─┐ │
 00 02 01 61 01 61 02 61 03 F0
          └───────┬───────┘
-        expression sequence
+        expression group body
 ```
 
 #### Example encoding of tagged `one-or-more` with single expression
@@ -612,7 +616,7 @@ is:
 00 01 61 01
 ```
 
-#### Example encoding of tagged `one-or-more` with populated group
+#### Example encoding of tagged `one-or-more` with expression group
 ```ion
 (:add_macros
   (foo (a+) ...)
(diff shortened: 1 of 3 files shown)


_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -31,19 +31,20 @@ Note that the opcode alone tells us which macro is being invoked, but it does no
...
 calling conventions_. (TODO: Link)
 
-
-#### E-expression With the Address as a Trailing `FixedUInt`
+#### E-expressions with biased `FixedUInt` addresses
 
 While E-expressions invoking macro addresses in the range `[0, 63]` can be encoded in a single byte using
 [E-expressions with the address in the opcode](#e-expression-with-the-address-in-the-opcode),
-many applications will benefit from defining more than 64 macros.
+many applications will benefit from defining more than 64 macros. The `0x4_` and `0x5_` opcodes
```
popematt, line 38: Just a FYI or something to consider. In markdown, if you have two lines of text with no blank lines between them, a markdown renderer will consider them part of the same paragraph. In light of that, it is sometimes helpful to start new sentences on a new line to produce a cleaner diff. I sometimes try to do that myself, but it's not something I would ever block a PR for.

_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -120,326 +136,529 @@ EF 01
...
-#### Encoding of e-expression `(:foo 3.14e)`
-```
-┌──── The opcode is less than 0x40, so it is an E-expression invoking the macro at
-│     address 0: `foo`. `foo` takes a tagged number as a parameter (`x`), so an opcode follows.
-│  ┌──── Opcode 0x6B indicates a 2-byte float; an IEEE-754 half-precision float follows
-│  │
-00 6B 47 42
-      └─┬─┘
-      3.14e0
+When a macro parameter does not specify an encoding (that is: the parameter name is not annotated), arguments
```
popematt:
```suggestion
When a macro parameter does not specify an encoding (i.e. the parameter name is not annotated), arguments
```
Or perhaps just "(the parameter name is not annotated)".

_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -120,326 +136,529 @@ EF 01
...
+passed to that parameter use the 'tagged' encoding. The argument begins with a leading [opcode](opcodes.md)
+that dictates how many bytes follow and how they should be interpreted.
```
popematt: nit: there are some opcodes that don't specify a length.
```suggestion
that dictates how to interpret the bytes that follow.
```

_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -120,326 +136,529 @@ EF 01
...
-#### Encoding of e-expression `(:foo $10::9)`
-```
-┌──── The opcode is less than 0x40, so it is an E-expression invoking the macro at
-│     address 0: `foo`. `foo` takes a tagged number as a parameter (`x`), so an opcode follows.
-│  ┌──── Opcode 0xE4 indicates a single annotation with symbol address. A FlexUInt follows.
-│  │  ┌──── Symbol address: FlexUInt 10 ($10); an opcode for the annotated value follows.
-│  │  │  ┌──── Opcode 0x61 indicates a 1-byte integer
-│  │  │  │   ┌──── 1-byte FixedInt 9
-00 E4 15 61 09
+When the macro has a single parameter, the corresponding encoded argument follows the opcode and (if separate) the encoded address.
```
popematt:
```suggestion
When the macro has a single _exactly-one_ parameter, the corresponding encoded argument follows the opcode and (if separate) the encoded address.
```

_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -120,326 +136,529 @@ EF 01
...
-│  │     at address 1: `bar`. `bar` takes no parameters, so no bytes follow.
-00 01
+### Encoding multiple `exactly-one` arguments
 
-// The macro expander confirms that the expansion of `(:bar)` (that is: `5`) matches
-// the expected type: `number`.
-```
+If the macro has more than one parameter, a reader would iterate over the parameters declared in the macro signature
+from left to right. For each parameter, the reader would use the parameter's declared encoding to interpret the next
+bytes in the stream. When no more parameters remain, parsing is complete.
```
popematt:
```suggestion
bytes in the stream. When no more parameters remain, parsing of the e-expression's arguments is complete.
```

_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -120,326 +136,529 @@ EF 01
...
-<div class="warning">
+##### Binary e-expression with the address in the opcode
+```
+┌──── Opcode 0x00 is less than 0x40; this is an e-expression
+│     invoking the macro at address 0.
+│  ┌─── Argument 'a': opcode 0x61 indicates a 1-byte integer
+│  │      ┌─── Argument 'b': opcode 0x61 indicates a 1-byte integer
+│  │      │    ┌─── Argument 'c': opcode 0x61 indicates a 1-byte integer
+│  │   1  │  2 │   3
+00 61 01 61 02 61 03
```
popematt: I think the numbers above the line of bytes are distracting from the main point of the diagram.
```suggestion
┌──── Opcode 0x00 is less than 0x40; this is an e-expression
│     invoking the macro at address 0.
│    ┌─── Argument 'a': opcode 0x61 indicates a 1-byte integer (1)
│    │     ┌─── Argument 'b': opcode 0x61 indicates a 1-byte integer (2)
│    │     │     ┌─── Argument 'c': opcode 0x61 indicates a 1-byte integer (3)
│  ┌─┴─┐ ┌─┴─┐ ┌─┴─┐
00 61 01 61 02 61 03
```

_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -120,326 +136,529 @@ EF 01
...
-│     address 0: `foo`. `foo` takes a tagged number as a parameter (`x`), so an opcode follows.
-│  ┌──── Opcode 0xEA represents an untyped null (aka `null.null`)
-00 EA
-
-// The macro expander confirms that `null` matches the expected type: `number`
+┌──── Opcode 0x00 is less than 0x40; this is an e-expression invoking
+│     the macro at address 0.
+│  ┌─── Argument 'x': opcode 0x61 indicates a 1-byte integer
+│  │   1
+00 61 01
```
popematt:
```suggestion
│    ┌─── Argument 'x': opcode 0x61 indicates a 1-byte integer (1)
│  ┌─┴─┐
00 61 01
```

_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -120,326 +136,529 @@ EF 01
...
 
-#### Encoding of e-expression `(:foo (:bar))`
+
+##### Binary e-expression using a trailing `FlexUInt` address
+```
+┌──── Opcode F4: An e-expression with a trailing FlexUInt address
+│  ┌──── FlexUInt 0: Macro address 0
+│  │  ┌─── Argument 'x': opcode 0x61 indicates a 1-byte integer
+│  │  │   1
+F4 01 61 01
```
popematt:
```suggestion
│  │    ┌─── Argument 'x': opcode 0x61 indicates a 1-byte integer (1)
│  │  ┌─┴─┐
F4 01 61 01
```

_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -120,326 +136,529 @@ EF 01
...
-> This section was obsolete and needs to be rewritten.
+##### Binary e-expression using a trailing `FlexUInt` address
+```
+┌──── Opcode F4: An e-expression with a trailing FlexUInt address
+│  ┌──── FlexUInt 0: Macro address 0
+│  │  ┌─── Argument 'a': opcode 0x61 indicates a 1-byte integer
+│  │  │     ┌─── Argument 'b': opcode 0x61 indicates a 1-byte integer
+│  │  │     │     ┌─── Argument 'c': opcode 0x61 indicates a 1-byte integer
+│  │  │     │     │
+│  │  │   1 │   2 │   3
```
popematt:
```suggestion
│  │    ┌─── Argument 'a': opcode 0x61 indicates a 1-byte integer (1)
│  │    │     ┌─── Argument 'b': opcode 0x61 indicates a 1-byte integer (2)
│  │    │     │     ┌─── Argument 'c': opcode 0x61 indicates a 1-byte integer (3)
│  │  ┌─┴─┐ ┌─┴─┐ ┌─┴─┐
```

_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -120,326 +136,529 @@ EF 01
...
+|          | `int16`            |       2       |                                                                                                                       |
+|          | `int32`            |       4       |                                                                                                                       |
+|          | `int64`            |       8       |                                                                                                                       |
+|          | `flex_int`         |   variable    | [`FlexInt`](primitives/flex_int.md)                                                                                   |
+| `float`  | `float16`          |       2       | [Little-endian IEEE-754 half-precision float](https://en.wikipedia.org/wiki/Half-precision_floating-point_format)     |
+|          | `float32`          |       4       | [Little-endian IEEE-754 single-precision float](https://en.wikipedia.org/wiki/Single-precision_floating-point_format) |
+|          | `float64`          |       8       | [Little-endian IEEE-754 double-precision float](https://en.wikipedia.org/wiki/Single-precision_floating-point_format) |
```
popematt, line 267: These could link to `values/float.md` instead. Your call, though.

_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -120,326 +136,529 @@ EF 01
...
+| `symbol` | `flex_sym`         |   variable    | [`FlexSym`](primitives/flex_sym.md)                                                                                   |
+
+#### Example encoding of an e-expression with primitive, `exactly-one` arguments
+
+As first demonstrated in _[Encoding multiple exactly-one arguments](#encoding-multiple-exactly-one-arguments)_,
+the bytes of the serialized arguments begin immediately after the opcode and (if separate) the macro address.
+The reader iterates over the parameters declared in the macro signature from left to right. For each parameter,
```
popematt: minor style suggestion—this is slightly more technical and precise, and if you translate the specification to a language that reads in a different direction, it would still be clear.
```suggestion
The reader iterates over the parameters in the macro signature in the order they are declared. For each parameter,
```

_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -120,326 +136,529 @@ EF 01
...
+```ion
+(:add_macros
+    (pi () 3.14159265) // A constant `pi`
 
-E-expression arguments corresponding to each parameter are encoded one after the other moving from left to right.
+    (circle_area (pi::Pi radius) (multiply Pi radius radius))
+    //            └── ERROR: cannot use a constant as a macro shape
 
-#### $1
+    (circle_area (radius) (multiply (.pi) radius radius))
```
popematt: Did we settle on whether we're going to use `(% ...)` for variable expansion (#343)?
```suggestion
    (circle_area (radius) (multiply (.pi) (%radius) (%radius)))
```
That being said, this example might be confusing to newcomers. It looks like there could be a `.` missing in front of `multiply`. I can't think of another example right now, but if I can come up with something, perhaps you would consider changing it.

_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -120,326 +136,529 @@ EF 01
...
+| `()`                  | _&lt;No variadics, no AEB>_ |
+| `(a b c)`             | _&lt;No variadics, no AEB>_ |
+| `(a b c?)`            | `0b------cc`                |
+| `(a b* c?)`           | `0b----ccbb`                |
+| `(a+ b* c?)`          | `0b--ccbbaa`                |
+| `(a+ b c?)`           | `0b----ccaa`                |
+| `(a+ b* c? d*)`       | `0bddccbbaa`                |
+| `(a+ b* c? d* e)`     | `0bddccbbaa`                |
+| `(a+ b* c? d* e f?)`  | `0bddccbbaa 0b------ff`     |
+| `(a+ b* c? d* e+ f?)` | `0bddccbbaa 0b----ffee`     |
```
popematt: I think the `0b` is distracting, especially since one of the parameters is `b`.
```suggestion
| `(a b c?)`            | `------cc`                |
| `(a b* c?)`           | `----ccbb`                |
| `(a+ b* c?)`          | `--ccbbaa`                |
| `(a+ b c?)`           | `----ccaa`                |
| `(a+ b* c? d*)`       | `ddccbbaa`                |
| `(a+ b* c? d* e)`     | `ddccbbaa`                |
| `(a+ b* c? d* e f?)`  | `ddccbbaa ------ff`     |
| `(a+ b* c? d* e+ f?)` | `ddccbbaa ----ffee`     |
```

_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -120,326 +136,529 @@ EF 01
...
+
+Each pair of bits in the AEB indicates what kind of expression to expect in the corresponding argument position.
+
+| Bit sequence | Meaning                                                                                                                           |
+|:------------:|:----------------------------------------------------------------------------------------------------------------------------------|
+|     `00`     | An **[empty expression group](#empty-groups)**. No bytes are present in the corresponding argument position.                      |
+|     `01`     | A **single expression** of the declared encoding is present in the corresponding argument position.                               |
+|     `10`     | A **[populated expression group](#populated-groups)** of the declared encoding is present in the corresponding argument position. |
```
popematt: I thought that `00` meant that the argument was an empty stream (not necessarily an empty expression group), and that `10` means that there is an expression group, but says nothing about how many expressions it contains (it could contain 0, hypothetically). Is that correct?

_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -120,326 +136,529 @@ EF 01
...
+If a parameter has a cardinality of `zero-or-more` or `one-or-more`, callers can choose to pass a
+populated expression group as the corresponding argument instead of a single expression. This is
+done by setting the bit sequence `10` in the appropriate position in the
+[Argument Encoding Bitmap](#argument-encoding-bitmap-aeb) and then writing an
+[expression sequence](#expression-sequence) in the corresponding argument position to indicate
+how many bytes part of this group.
+
+##### Expression sequence
+
+An _expression sequence_ begins with a [`FlexUInt`](primitives/flex_uint.md). If the `FlexUInt`'s value
```
popematt: Does "expression sequence" have a different meaning than "expression group"?

_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -120,326 +136,529 @@ EF 01
...
+(:foo 1)
+```
 
-<div class="warning">
+```
+┌──── Opcode 0x00 is less than 0x40; this is an e-expression
+│     invoking the macro at address 0.
+│  ┌──── AEB: 0b------aa; a=01, single expression
+│  │  ┌──── Argument 'a': opcode 0x61 indicates a 1-byte int
+│  │  │   1
```
popematt: I think having the numbers directly above the bytes is distracting from the main point here.
```suggestion
│  │    ┌── Argument 'a': opcode 0x61 indicates a 1-byte int (1)
│  │  ┌─┴─┐
```

_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -120,326 +136,529 @@ EF 01
...
-    (int::x*)    // Parameters; `x` is a grouped parameter
-    /*...*/      // Body (elided)
-)
+
+```
+┌──── Opcode 0x00 is less than 0x40; this is an e-expression
+│     invoking the macro at address 0.
+│  ┌──── AEB: 0b------aa; a=01, single expression
+│  │  ┌──── Argument 'a': opcode 0x61 indicates a 1-byte int
+│  │  │   1
```
popematt:
```suggestion
│  │    ┌── Argument 'a': opcode 0x61 indicates a 1-byte int (1)
│  │  ┌─┴─┐
```

_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -120,326 +136,529 @@ EF 01
...
-│  │  ┌──── FlexUInt: 6; the next 6 bytes are an `int` expression group.
-│  │  │
+┌──── Opcode 0x00 is less than 0x40; this is an e-expression
+│     invoking the macro at address 0.
+│  ┌──── AEB: 0b------aa; a=10, populated expression group
+│  │  ┌──── FlexUInt 6: 6-byte expression sequence
+│  │  │  ┌──── Opcode 0x61 indicates a 1-byte int
+│  │  │  │     ┌──── Opcode 0x61 indicates a 1-byte int
+│  │  │  │     │     ┌─── Opcode 0x61 indicates a 1-byte int
+│  │  │  │   1 │  2  │   3
```
popematt:
```suggestion
│  │  │    ┌── Opcode 0x61 indicates a 1-byte int (1)
│  │  │    │     ┌── Opcode 0x61 indicates a 1-byte int (2)
│  │  │    │     │     ┌─ Opcode 0x61 indicates a 1-byte int (3)
│  │  │  ┌─┴─┐ ┌─┴─┐ ┌─┴─┐
```

_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -120,326 +136,529 @@ EF 01
...
+┌──── Opcode 0x00 is less than 0x40; this is an e-expression
+│     invoking the macro at address 0.
+│  ┌──── AEB: 0b------aa; a=10, populated expression group
+│  │  ┌──── FlexUInt 0: delimited expression sequence
+│  │  │  ┌──── Opcode 0x61 indicates a 1-byte int
+│  │  │  │     ┌──── Opcode 0x61 indicates a 1-byte int
+│  │  │  │     │     ┌─── Opcode 0x61 indicates a 1-byte int
+│  │  │  │     │     │      ┌─── Opcode 0xF0 is delimited end
+│  │  │  │   1 │  2  │   3  │
+00 02 01 61 01 61 02 61 03 F0
```
popematt:
```suggestion
│  │  │    ┌── Opcode 0x61 indicates a 1-byte int (1)
│  │  │    │     ┌── Opcode 0x61 indicates a 1-byte int (2)
│  │  │    │     │     ┌─ Opcode 0x61 indicates a 1-byte int (3)
│  │  │    │     │     │    ┌─── Opcode 0xF0 is delimited end
│  │  │  ┌─┴─┐ ┌─┴─┐ ┌─┴─┐  │
00 02 01 61 01 61 02 61 03 F0
```

_books/ion-1-1/src/macros/defining_macros.md
```diff
@@ -56,17 +56,69 @@ To specify an encoding, the [parameter name](#parameter-names) is annotated with
...
+Optional parameters that are _not_ in tail position cannot be elided, as this would
+cause them to appear in a position corresponding to a different argument.
+
+```ion
+(:set_macros
+    (foo (x? y) // `x` is optional, `y` is required
+        [x, y]
+    )
+)
+(:annotate foo)
```
popematt: Is this supposed to be here?

_books/ion-1-1/src/binary/e_expressions.md
zslayton: When writing markdown, I find there's a tension between keeping the file tidy and keeping the diff tidy. :/ I once read a suggestion that every sentence in the file should be on its own line for optimal diffing; I really liked that, but found the result to occasionally be visually jarring.

It'd be cool if I could make an optimal-diff commit and then run an autoformatter as a final step to re-wrap the paragraphs to fit some number of columns.

_books/ion-1-1/src/binary/e_expressions.md
zslayton: I find `i.e.` is less approachable than "that is", but YMMV. (I say that having already written "i.e." in a bunch of places. 😓)

> Or perhaps just "(the parameter name is not annotated)".

I like this version best, thanks.

_books/ion-1-1/src/binary/e_expressions.md
zslayton: The encodings in `values/float.md` all start with the appropriate opcode; we'd have to add a section for the primitive encodings.

_books/ion-1-1/src/binary/e_expressions.md
zslayton: Sure, I'll use that.

> and if you translate the specification to a language that reads in a different direction, it would still be clear

I'm not sure I follow. If the specification were in, say, Arabic, that wouldn't affect the direction that one would read the parameter names in the macro signature. They would still be left-to-right.

_books/ion-1-1/src/binary/e_expressions.md
zslayton: Yeah, it's hard not to read it as math. The `.` was omitted intentionally--this example emits a somewhat arbitrary s-expression. I'll think of something else.

zslayton: I'm going to hold off on using `(%radius)` syntax because my next PR (in progress) updates variable expansion syntax throughout the spec.

_books/ion-1-1/src/binary/e_expressions.md
zslayton: > I thought that 00 meant that the argument was an empty stream (not necessarily an empty expression group), 

You've pinpointed a distinction I was wrestling with while writing this! "Empty stream" is a much better term for this; as written, I had to coin a couple of other phrases to distinguish between populated/empty, implicit/explicit argument groups.

Let me take another pass at this using updated terminology.

_books/ion-1-1/src/macros/defining_macros.md
```diff
@@ -56,17 +56,68 @@ To specify an encoding, the [parameter name](#parameter-names) is annotated with
...
-|   `!`    | exactly-one value   |
-|   `?`    | zero-or-one value   |
-|   `+`    | one-or-more values  |
-|   `*`    | zero-or-more values |
+| Modifier |         Cardinality | Accepts single expression | Accepts empty group | Accepts populated group |
+|:--------:|--------------------:|:-------------------------:|:-------------------:|:-----------------------:|
+|   `?`    |   zero-or-one value |             ✅             |          ✅          |            ❌            |
+|   `*`    | zero-or-more values |             ✅             |          ✅          |            ✅            |
+|   `!`    |   exactly-one value |             ✅             |          ❌          |            ❌            |
+|   `+`    |  one-or-more values |             ✅             |          ❌          |            ✅            |
```
popematt: I would not include this extension to the table. I understand what you're trying to get at, but I think it's best to keep the abstract model of macro arguments distinct from the encoding concerns for macro arguments.

My mental model of "Cardinality" is that it applies to the variable expansion. I.e. this is the essence of cardinality:
>If an argument expression expands to a number of values that the cardinality forbids, the reader must raise an error.


Then, restrictions on the encoding can be derived from the cardinality, and any such restrictions should be specified in the encoding sections (which you have done for example on L434 of `e_expressions.md`).

What do you think? Am I understanding things correctly?

zslayton: > I think it's best to keep the abstract model of macro arguments distinct from the encoding concerns for macro arguments. My mental model of "Cardinality" is that it applies to the variable expansion.

Unfortunately, cardinality is a confluence of encoding and expansion. I wish that it weren't--my preference was that it only affect encoding (`!` or `*`) and that we'd denote expansion constraints separately (` -> string+`).  As it stands, the text encoding prohibits passing a populated expression group to `?` expressions. It also prohibits passing an empty expression group to ` +`. (Groups also cannot be passed to `!`, but that's not controversial.)

We need the text and binary encodings to be consistent enough in this regard that we can easily transcribe between the two. Either binary needs to prohibit the same combinations, or we allow variadics to use any encoding and only validate the expansion.

I'm open to removing this table extension for readability. However I think we should be _able_ to discuss the limitations at this spec level because they're consistent across the encodings. Ideally, reproducing the binary section's table in the text section would be redundant.

Let's talk about this more later today.

_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -120,326 +136,583 @@ EF 01
...
 
-// The macro expander confirms that `$10::9` (an annotated `int`) matches the expected type: `number`.
-```
+#### Example encoding of an e-expression with a tagged, `exactly-one` argument
 
-#### Encoding of e-expression `(:foo null.int)`
+##### Macro definition
+```ion
+(:set_macros
+  (foo (x) ...)
```
tgregg: I think we should either make it clear that `...` is not actual syntax, or we should use real syntax for the template body.

_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -120,326 +136,583 @@ EF 01
...
+|          | `flex_int`         |   variable    | [`FlexInt`](primitives/flex_int.md)                                                                                   |
+| `float`  | `float16`          |       2       | [Little-endian IEEE-754 half-precision float](https://en.wikipedia.org/wiki/Half-precision_floating-point_format)     |
+|          | `float32`          |       4       | [Little-endian IEEE-754 single-precision float](https://en.wikipedia.org/wiki/Single-precision_floating-point_format) |
+|          | `float64`          |       8       | [Little-endian IEEE-754 double-precision float](https://en.wikipedia.org/wiki/Single-precision_floating-point_format) |
+| `symbol` | `flex_sym`         |   variable    | [`FlexSym`](primitives/flex_sym.md)                                                                                   |
+
+#### Example encoding of an e-expression with primitive, `exactly-one` arguments
+
+As first demonstrated in _[Encoding multiple exactly-one arguments](#encoding-multiple-exactly-one-arguments)_,
+the bytes of the serialized arguments begin immediately after the opcode and (if separate) the macro address.
```
tgregg:
```suggestion
the bytes of the serialized arguments begin immediately after the e-expression opcode and (if separate) the macro address.
```
Or something like that, to be clear about which opcode we're referring to (since the arguments themselves don't have opcodes in this case).

_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -120,326 +136,583 @@ EF 01
...
+The reader iterates over the parameters in the macro signature in the order they are declared. For each parameter,
+the reader uses the parameter's declared encoding to interpret the next bytes in the stream. When no more parameters
+remain, parsing is complete.
+
+##### Macro definition
+```ion
+(:set_macros
+  (foo (flex_uint::a int8::b uint16) ...)
```
tgregg:
```suggestion
  (foo (flex_uint::a int8::b uint16::c) ...)
```

_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -120,326 +136,583 @@ EF 01
...
+   │  │  │  └────   $end/$y: FlexInt 4
+   │  │  └───────   $end/$x: FlexInt 3
+   │  └────────── $start/$y: FlexInt 2
+   └───────────── $start/$x: FlexInt 1
+```
+
+Any macro can be used as a macro shape except for _constants_--macros which take zero parameters.
+Constants cannot be used as a macro shape because their serialized representation would be empty,
+making it impossible to encode them in [expression groups](#expression-groups). However, this
+limitation not sacrifice any expressiveness; the desired constant can always be invoked directly
```
tgregg:
```suggestion
limitation does not sacrifice any expressiveness; the desired constant can always be invoked directly
```

_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -120,326 +136,583 @@ EF 01
...
+but these parameters have always had the same [cardinality](../macros/defining_macros.md#parameter-cardinalities):
+`exactly-one`.
+
+This section explains how to encode e-expressions invoking a macro whose signature contains
+_variadic parameters_--parameters with a cardinality of `zero-or-one`, `zero-or-more`, or `one-or-more`.
+
+#### Argument Encoding Bitmap (AEB)
+
+If a macro signature has one or more variadic parameters, then e-expressions invoking that macro will include an additional
+construct: the _Argument Encoding Bitmap (AEB)_. This is a little-endian byte sequence precedes the first serialized argument
```
tgregg:
```suggestion
construct: the _Argument Encoding Bitmap (AEB)_. This little-endian byte sequence precedes the first serialized argument
```

_books/ion-1-1/src/binary/e_expressions.md
```diff
@@ -120,326 +136,583 @@ EF 01
...
+
+An expression group begins with a [`FlexUInt`](primitives/flex_uint.md). If the `FlexUInt`'s value
+is:
+* **greater than zero**, then it represents the number of bytes used to encode the rest of the expression group. The reader
+  should continue reading expressions of the declared encoding until that number of bytes has been consumed.
+* **zero**, then it indicates that this is a delimited expression group and the processing varies according to
+  whether the declared encoding is tagged or tagless. If the encoding is:
+  * **[tagged](#tagged-encoding)**, then each expression in the group begins with an opcode. The reader
+    must consume tagged expressions until it encounters a terminating `END` opcode (`0xF0`).
+  * **[tagless](#tagless-encodings)**, then each expression in the group has no leading opcode; there is no
```
tgregg: I suggest rewording this slightly to emphasize that the first byte(s) that follows the expression group's starting FlexUInt 0 byte is a `FlexUInt` length prefix. I think talking about the expressions first risks confusion.

_books/ion-1-1/src/macros/defining_macros.md
```diff
@@ -56,17 +56,68 @@ To specify an encoding, the [parameter name](#parameter-names) is annotated with
...
+cause them to appear in a position corresponding to a different argument.
+
+```ion
+(:set_macros
+    (foo (x? y) // `x` is optional, `y` is required
+        [x, y]
+    )
+)
+
+(:foo 1)   => ERROR: missing required argument `y`
```
tgregg: We could add a line to demonstrate how to use an empty expression group to elide x.

_books/ion-1-1/src/binary/e_expressions.md
zslayton: Yeah, that's fair. I'll replace each `...` with `/*...*/`--does that work?

_books/ion-1-1/src/binary/e_expressions.md
zslayton: Good catch

_books/ion-1-1/src/macros/defining_macros.md
zslayton: I've [removed the extension](https://github.com/amazon-ion/ion-docs/pull/348/commits/45084a9685848198c5205a0878dc35fa2078deda) in the interest of getting the rest of this merged. We can revisit the table later.

popematt (approved):

_books/ion-1-1/src/binary/e_expressions.md
tgregg: Yep!

tgregg (approved):

PR #352, opened 2024-10-15
Adds `default`, renames `comment` to `meta`, removes `fail`

zslayton:
Updates macros per offline discussion.

----


_books/ion-1-1/src/macros/special_forms.md
```diff
@@ -184,10 +184,3 @@ a b c
 
 The `for` special form can only be invoked in the body of template macro. It is not valid to use as an E-Expression.
 
-### `fail`
-
-```ion
-(macro fail (message?) /* Not representable in TDL */)
-```
-
-Causes macro evaluation to immediately halt and causes the Ion reader to raise an error to the user.
```
zslayton, line 193: 🗺️ Ion 1.1 only offers branching on expansion cardinality, which means that the only time you could selectively call `fail` would be based on expansion cardinality.

Expansion cardinality can already be enforced by `?`/`!`/`*`/`+`, so we're removing this.

_books/ion-1-1/src/macros/system_macros.md
```diff
@@ -29,6 +29,24 @@ This is, essentially, the identity function.
...
+        // If it wasn't empty, then expand `expr`.
+        (%expr)
+    )
+)
+```
+
+`default` tests `expr` to determine whether it expands to the empty stream.
+If it does not, `default` will produce the expansion of `expr`.
+If it does, `default` will produce the expansion of `default_expr` instead.
+
```
zslayton, line 49: 🗺️ User testing found that most uses of `if_none` were implementing `default`. Adding `default` is a nice convenience.

_books/ion-1-1/src/macros/system_macros.md
```diff
@@ -297,19 +315,22 @@ Examples:
...
+It allows writers to encode data that will be not be surfaced to most readers.
+Readers can be configured to intercept calls to `meta`, allowing them to read the otherwise invisible data.
+
+When transcribing from one format to another, writers should preserve invocations of `meta` when possible.
 
 Example:
 ```ion
 (:values
-    (:comment equivalent to {foo:1,foo:2})
+    (:meta {author: "Mike Smith", email: "mikesmith@example.com"})
```
zslayton, line 333: 🗺️ `comment` is an unnecessarily narrow name. This macro can accept arbitrary data.

popematt (changes requested):
Can you make sure to update `system_module.md` to match these changes? Otherwise, this looks great.

_books/ion-1-1/theme/old_highlight.js
popematt, line 1: Was this included in the diff by accident?

zslayton: Yes, I'm currently working on excising it. I did a `git add .` and forgot I was toying with the highlighting. 🤦

popematt (approved):

tgregg (approved):

PR #359, opened 2024-10-18
Adds glossary and consolidates grammar

popematt:
### Description of changes:

* Consolidates the macro and module grammar
* Adds a glossary
* Fixes some typos in `defining_macros.md` and adds a missing link
* Drive by fix of the Ion 1.0 spec to clarify that keywords (`true`, `false`, `null`, `nan`) are not identifier symbols.

----


_books/ion-1-1/src/glossary.md
```diff
@@ -0,0 +1,159 @@
...
+
+**declaration**<br/>
+The association of a name with an entity (for example, a module or macro). See also _definition_. 
+Not all declarations are definitions: some introduce new names for existing entities.
+
+**definition**<br/>
+The specification of a new entity.
+
+**directive**<br/>
+A keyword or unit of data in an Ion document that affects the encoding environment, and thus the way the document's data is decoded.
```
tgregg: encoded and decoded?

_books/ion-1-1/src/glossary.md
```diff
@@ -0,0 +1,159 @@
...
+A macro reference that consists of a module name and either a macro name exported by that module,
+or a numeric address within the range of the module's exported macro table. In TDL, these look
+like _module-name_::_name-or-address_.
+
+**required parameter**<br/>
+A macro parameter that is not _optional_ and therefore requires an argument at each invocation.
+
+**rest parameter**<br/>
+A macro parameter—always the final parameter—declared with `*` or `+` cardinality,
+that accepts all remaining individual arguments to the macro as if they were in an implicit _argument group_.
```
tgregg, line 110: It may be worth clarifying that implicit argument groups for rest parameters are only applicable in the text format.

_books/ion-1-1/src/glossary.md
```diff
@@ -0,0 +1,159 @@
...
+Similar to "varargs" parameters in Java and other languages.
+
+**segment**<br/>
+A contiguous partition of a _document_ that uses the same _active encoding module_. Segment boundaries
+are caused by directives: an IVM starts a new segment, while `$ion_symbol_table` and `$ion_encoding`
```
tgregg: An IVM ends the previous segment (if any) and starts a new one, correct?

_books/ion-1-1/src/glossary.md
```diff
@@ -0,0 +1,159 @@
...
+directives end segments (with a new one starting immediately afterward).
+
+**shared module**<br/>
+A module that exists independent of the data stream of an Ion document. It is identified by a
+name and version so that it can be imported by other modules.
+
+**signature**<br/>
+The part of a macro definition that specifies its "calling convention", in terms of the shape,
+type, and cardinality of arguments it accepts, and the type and cardinality of the results it
```
tgregg: Does the signature say anything about the results the macro produces? Or is this language left over from a previous revision?

_books/ion-1-1/src/glossary.md
```diff
@@ -0,0 +1,159 @@
...
+produces.
+
+**symbol table directive**<br/>
+A top-level struct annotated with `$ion_symbol_table`.  Defines a new encoding environment
+without any macros.  Valid in Ion 1.0 and 1.1.
+
+**system E-Expression**<br/>
```
tgregg: Are we always capitalizing "E-Expression" as a stylistic choice? I think I prefer "e-expression" here, for what it's worth.

_books/ion-1-1/src/glossary.md
```diff
@@ -0,0 +1,159 @@
...
+varies by segment according to its Ion version.
+
+**TDL**<br/>
+See _template definition language_.
+
+**template**<br/>
+The part of a macro definition that expresses its transformation of inputs to results.
+
+**template definition language**<br/>
+An Ion-based, domain-specific language that declaratively specifies the output produced by a _macro_.
```
tgregg, line 155: Is it worth pointing out that TDL is expressed using valid Ion 1.0 syntax?

_books/ion-1-1/src/glossary.md
popematt: Leftover.

_books/ion-1-1/src/glossary.md
popematt: Text and TDL, I think.

_books/ion-1-1/src/glossary.md
```diff
@@ -0,0 +1,160 @@
+# Glossary
+
+**active encoding module**<br/>
+The _encoding module_ whose symbol table and macro table are available in the current _segment_ of an Ion _document_.
+The active encoding module is set by a _directive_.
```
zslayton, line 5: Thinking out loud, it would be really nice to incorporate something like [this pre-processor](https://github.com/caukub/mdbook-hints) to make phrases in the spec show their definition on hover.

Having a glossary is great, but not having to leave the page to define a term would be even better.

EDIT: I just saw that preprocessor's [first and only issue](https://github.com/caukub/mdbook-hints/issues/1). 😆

_books/ion-1-1/src/glossary.md
```diff
@@ -0,0 +1,160 @@
...
+
+**argument**<br/>
+The sub-expression(s) within a macro invocation, corresponding to exactly one of the macro's parameters.
+
+**cardinality**<br/>
+Describes the number of values that a parameter will accept when the macro is invoked.
+One of zero-or-one, exactly-one, zero-or-more, or one-or-more.
+Specified in a signature by one of the modifiers `?`, `!`, `*`, or `+`.
```
zslayton:
```suggestion
Describes both:
1. the number of argument expressions that a parameter will accept when the macro is invoked.
2. the number of values that the parameter will expand to during evaluation.
A parameter's cardinality can be `zero-or-one`, `exactly-one`, `zero-or-more`, or `one-or-more`,
specified in a signature by one of the modifiers `?`, `!`, `*`, or `+` respectively.
If no modifier is specified, cardinality defaults to `exactly-one`.
```

_books/ion-1-1/src/glossary.md
```diff
@@ -0,0 +1,160 @@
...
+
+**macro**<br/>
+A transformation function that accepts some number of streams of values, and produces a stream of values.
+
+**macro definition**<br/>
+Specifies a macro in terms of a _signature_ and a _template_.
+
+**macro reference**<br/>
+Identifies a macro for invocation or exporting. Must always be unambiguous. Lexically
+scoped, and never a "forward reference" to a macro that is declared later in the document.
```
zslayton: Minor tweak to avoid suggesting that there might be another way to achieve forward referencing.
```suggestion
scoped. Cannot be a "forward reference" to a macro that is declared later in the document;
these are not legal.
```

_books/ion-1-1/src/glossary.md
```diff
@@ -0,0 +1,160 @@
...
+**module**<br/>
+The data entity that defines and exports both symbols and macros.
+
+**opcode**<br/>
+A 1-byte, unsigned integer that tells the reader what the next expression represents
+and how the bytes that follow should be interpreted.
+
+**optional parameter**<br/>
+A parameter that can have its corresponding subform(s) omitted when the macro is invoked.
+A parameter is optional if it is _voidable_ and all following arguments are also voidable.
```
zslayton:
```suggestion
A parameter is optional if both it and the parameters that follow it in the macro signature can accept an empty stream.
```

_books/ion-1-1/src/glossary.md
```diff
@@ -0,0 +1,160 @@
...
+
+**TDL**<br/>
+See _template definition language_.
+
+**template**<br/>
+The part of a macro definition that expresses its transformation of inputs to results.
+
+**template definition language**<br/>
+An Ion-based, domain-specific language that declaratively specifies the output produced by a _macro_.
+Template definition language uses only the Ion data model (i.e. valid Ion 1.0 syntax).
```
zslayton: Is this true, though? I can use an e-expression to write a macro definition.

> Template definition language uses only the Ion data model

This part is true; no matter how I encode my template definition in an Ion 1.1 stream, there's an equivalent way of expressing it in Ion 1.0.

_books/ion-1-1/src/glossary.md
```diff
@@ -0,0 +1,160 @@
...
+
+**unqualified macro reference**<br/>
+A macro reference that consists of either a macro name or numeric address, without a qualifying module name. 
+These are resolved using lexical scope and must always be unambiguous.
+
+**variable expansion**<br/>
+In _TDL_, a special form that causes the expanded _arguments_ for the given _parameter_ to be substituted into the _template_.
```
zslayton: There are some mixed plural/singular usages of "argument" that make its meaning a bit hazy. Based on usage, a reader could understand it to mean either:
1. a single expression (possibly one of a group) passed to a parameter
  OR
2. the only expression passed to a parameter (this expression may be an expression group containing subexpressions)

We should consolidate our usage to reflect the intended meaning.

zslayton (approved):
Looks good, a few small comments that can be TODOs

_books/ion-1-1/src/glossary.md
popematt:
```suggestion
Describes both the number of argument expressions that a parameter will accept when the macro is invoked,
and the number of values that the parameter may expand to during evaluation.
A parameter's cardinality can be `zero-or-one`, `exactly-one`, `zero-or-more`, or `one-or-more`,
specified in a signature by one of the modifiers `?`, `!`, `*`, or `+` respectively.
If no modifier is specified, cardinality defaults to `exactly-one`.
```

_books/ion-1-1/src/glossary.md
popematt:
```suggestion
In _TDL_, a special form that causes all argument expression(s) for the given _parameter_ to be expanded and the result of the expansion to be substituted into the _template_.
```

_books/ion-1-1/src/glossary.md
popematt: I wasn't quite sure how to phrase this because the Ion 1.1 spec doesn't include a specific chapter on the Ion data model, but I can drop the bit about Ion 1.0.

```suggestion
Template definition language uses only the Ion data model.
```

PR #360, opened 2024-10-21
fix `0d` in decimal.md

liigo:
### Description of changes:

minor fix to the docs

----


popematt:
While we appreciate your interests in the project, the use of the letter `d` is actually not a problem here. If you refer to the Ion 1.0 specification, you can see that `decimal` uses `d` to separate the coefficient from the exponent in Ion text. Removing the `d` leaving only an integer actually represents a different data type (an `integer`), and swapping the `d` for an `e` likewise changes the data type from `decimal` to `float`. Please see https://amazon-ion.github.io/ion-docs/docs/spec.html#real-numbers for details.

liigo:
I see. Thanks.

PR #361, opened 2024-10-21
avoid double `U`s

liigo:
the letter code `U` stands for Utc not Unkown

### Description of changes:

----

### DIFF
--- a/_books/ion-1-1/src/binary/values/timestamp.md
+++ b/_books/ion-1-1/src/binary/values/timestamp.md
@@ -84,10 +84,10 @@ The `Month` and `Day` subfields are one-based; `0` is not a valid month or day.
 |     `H`     |               5               | Hour                              |
 |     `m`     |               6               | Minute                            |
 |     `o`     |               7               | Offset                            |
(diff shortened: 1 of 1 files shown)


popematt:
We appreciate your interest in Ion 1.1 and your willingness to contribute. The choice of the word "unknown" is not by accident. An offset of "none" could imply that the offset is actually zero, which is not what "unknown offset" refers to. The Ion 1.0 specification says of [timestamps](
https://amazon-ion.github.io/ion-docs/docs/spec.html#timestamp):

> Ion follows the “Unknown Local Offset Convention” of [RFC3339](http://www.ietf.org/rfc/rfc3339.txt):
>
>>If the time in UTC is known, but the offset to local time is unknown, this can be represented with an offset of “-00:00”. This differs semantically from an offset of “Z” or “+00:00”, which imply that UTC is the preferred reference point for the specified time. [RFC2822](http://www.ietf.org/rfc/rfc2822.txt) describes a similar convention for email.

Given that Ion follows the convention of RCF 3339, which uses "Unknown" to describe the offset, Ion will continue using "Unknown".

PR #366, opened 2024-11-14
Adds chapter on security

popematt:
### Issue #, if available:

None

### Description of changes:

Adds a chapter describing various security considerations for Ion 1.1

----

### DIFF
--- a/_books/ion-1-1/src/security-considerations.md
+++ b/_books/ion-1-1/src/security-considerations.md
@@ -65,10 +65,13 @@ $ion_1_1
 Implementations of Ion 1.1 _MUST_ have some mechanism by which to mitigate data expansion attacks.
 
 The macro evaluator of Ion 1.1 implementations _SHOULD_ have a (possibly configurable) limit on the number of values 
-produced by the expansion of a macro or e-expression. If the macro evaluator reaches that limit, evaluation should halt 
+produced by the expansion of any macro or e-expression. If the macro evaluator reaches that limit, evaluation should halt 
 and the reader should signal an error. This is similar to the [Token Bucket Algorithm](https://en.wikipedia.org/wiki/Token_bucket),
 but instead of refilling the bucket, the bucket starts at the maximum capacity whenever the reader begins evaluating an
-e-expression that is not nested in any other e-expression at any other depth.
+e-expression that is not nested in any other e-expression at any other depth. In order to prevent a malicious input that
+produces no values (for example, `(macro sneaky_lolz () (.meta (.lolz)))`), tokens _SHOULD_ be consumed at every level of
+expansion, including special forms and TDL macro invocations. Expansions that are skipped are not required to consume 
+tokens (since they are not expanded), but an empty expansion _MUST_ consume at least one token.
 
 ```ion
 $ion_1_1
@@ -86,23 +89,27 @@ $ion_1_1
 {
   // Fill bucket here
   foo: (:make_string "foo" "bar")
-  // Fill bucket here
-  bar: (:make_string "a" "b")
+  // Fill bucket here.
+  // Consume one token for each value produced by repeat and for each value produced by make_string
+  bar: (:make_string (:repeat 16 "na") " batman!")
 }
 ```
 
 
 ## Remote code execution
 
-Remote code execution (RCE) attacks allow an attacker to remotely execute malicious code on a computer.
+The template definition language (TDL) is a domain specific programming language used to declare template macros 
+in Ion 1.1. It is intentionally limited in its capabilities—it cannot recurse and does not support forward references.
+In general, it supports combining Ion values to produce other Ion values, but it does not support arbitrary computation
+on those values.
 
-The template definition language (TDL) is a type of code, and by invoking e-expressions in the body of an Ion document, 
-an attacker can cause the recipient to execute arbitrary TDL when reading the document.
+Remote code execution (RCE) attacks allow an attacker to remotely execute malicious code on a computer. By invoking 
+e-expressions in the body of an Ion document, an attacker can cause the recipient to execute arbitrary TDL (code)
+when reading the document.
 
-This is unlikely to be a concern in practice because TDL is not arbitrary code.
+This is unlikely to be a concern in practice because TDL is not _arbitrary_ code.
 TDL is intentionally not Turing complete, to make it impossible to perform arbitrary computation.
 It also has a very limited domain—it can only transform/produce Ion data model values.
-
 While it could be possible to attempt a denial-of-service attack using TDL, TDL expansion is guaranteed to terminate in
 a finite number of steps, and implementations can additionally limit the expansion size (as described above).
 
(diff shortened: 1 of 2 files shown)


_books/ion-1-1/src/security-considerations.md
```diff
@@ -0,0 +1,156 @@
...
+                    (:repeat 2147483647 "abc")))))))))))
+```
+The expansion of these e-expressions results in a stream of ~450 [googol](https://en.wikipedia.org/wiki/Googol) string
+values. Any attempt to hold all of this in memory or write it to disk will exhaust all available resources and 
+eventually fail. Even an attempt to count the length of the stream, while it may theoretically succeed if using an 
+appropriate `BigInteger` type, will require a considerable amount of CPU operations (over a googol), and even the 
+fastest processors will require many millennia to completely count the number of values in the stream.
+
+Even without using `repeat` or `for`, a [Billion laughs attack](https://en.wikipedia.org/wiki/Billion_laughs_attack)
+could exist for any data format with macro expansion, and it is certainly possible with Ion 1.1.
```
zslayton, line 48: Neat, we've discussed this attack a bunch of times but I didn't know it had an established name.

_books/ion-1-1/src/security-considerations.md
```diff
@@ -0,0 +1,156 @@
...
+             (macro lol2 () (.values (.lol1) (.lol1) (.lol1) (.lol1) (.lol1) (.lol1) (.lol1) (.lol1) (.lol1) (.lol1)))
+             (macro lol3 () (.values (.lol2) (.lol2) (.lol2) (.lol2) (.lol2) (.lol2) (.lol2) (.lol2) (.lol2) (.lol2)))
+             (macro lol4 () (.values (.lol3) (.lol3) (.lol3) (.lol3) (.lol3) (.lol3) (.lol3) (.lol3) (.lol3) (.lol3)))
+             (macro lol5 () (.values (.lol4) (.lol4) (.lol4) (.lol4) (.lol4) (.lol4) (.lol4) (.lol4) (.lol4) (.lol4)))
+             (macro lol6 () (.values (.lol5) (.lol5) (.lol5) (.lol5) (.lol5) (.lol5) (.lol5) (.lol5) (.lol5) (.lol5)))
+             (macro lol7 () (.values (.lol6) (.lol6) (.lol6) (.lol6) (.lol6) (.lol6) (.lol6) (.lol6) (.lol6) (.lol6)))
+             (macro lol8 () (.values (.lol7) (.lol7) (.lol7) (.lol7) (.lol7) (.lol7) (.lol7) (.lol7) (.lol7) (.lol7)))
+             (macro lol9 () (.values (.lol8) (.lol8) (.lol8) (.lol8) (.lol8) (.lol8) (.lol8) (.lol8) (.lol8) (.lol8)))
+             (macro lolz () (.lol9)) )
+(:lolz)
```
zslayton, line 62: We should make a `TODO` to track adding these examples to `ion-tests`.

_books/ion-1-1/src/security-considerations.md
```diff
@@ -0,0 +1,156 @@
...
+## Remote code execution
+
+Remote code execution (RCE) attacks allow an attacker to remotely execute malicious code on a computer.
+
+The template definition language (TDL) is a type of code, and by invoking e-expressions in the body of an Ion document, 
+an attacker can cause the recipient to execute arbitrary TDL when reading the document.
+
+This is unlikely to be a concern in practice because TDL is not arbitrary code.
+TDL is intentionally not Turing complete, to make it impossible to perform arbitrary computation.
+It also has a very limited domain—it can only transform/produce Ion data model values.
```
zslayton: I think it would be good to have a brief section on the intentional limitations built into TDL at the top of this section. Stating early on that it cannot recurse, has no forward references, etc would help clarify the reader's mental model going into the discussion of actual vulnerabilities.

_books/ion-1-1/src/security-considerations.md
```diff
@@ -0,0 +1,156 @@
...
+client trusts the server.)
+
+If shared modules must come from an untrusted source, then applications should take steps to ensure that the shared
+modules originate from the same source as the data that uses them, and they can be treated as if they are one composite
+piece of data from that source.
+
+## Arbitrary-sized values
+
+The Ion specification places no limits on the size of Ion values, so an attacker could send a sufficiently large value,
+it could consume enough system resources to disrupt the application reading the value.
```
zslayton, line 144: The token bucket solution covers a lot of outsized-expansion issues, but it doesn't cover macros where (by using built-in operations like `make_string`) an enormous single value could be produced. We may wish to either restrict the number of input arguments or put a size limit on `make_string`, `make_blob`, etc.

popematt: I think that most programming languages already have a limit on the length of strings, byte arrays, lists/vecs, etc. Do you think we need more than that?

zslayton: I think we're talking about two different things. I'm describing a flavor of the small-input-big-output problem that doesn't generate lots of values, it generates one gigantic value.
```
(.make_string (.repeat 1_000_000_000 "lol"))
```
The token bucket/fuel approach doesn't handle this case because:
1. `repeat` is nested inside of an e-expression, so it isn't subject to the token limit
2. `make_string` "only" produces one value, so it's within the token limit.

It makes sense for us to offer limits on (for example) the number of arguments that can be passed to `make_string`.

zslayton: > repeat is nested inside of an e-expression, so it isn't subject to the token limit

Re-reading that section of the doc, I may have misunderstood this part. If `repeat` is _also_ pulling from the same token bucket as the parent e-expression, this might be ok as-is.

_books/ion-1-1/src/security-considerations.md
```diff
@@ -0,0 +1,156 @@
...
+unknown and/or are dependent on the underlying runtime environment.
+
+## Symbol table and macro table inflation
+ 
+An attacker could try to create an input that results in excessively large symbol and macro tables in the Ion reader 
+that could exhaust the memory of the receiving system and lead to a denial of service.
+
+Although Ion 1.1 does not specify a maximum size for symbol tables or macro tables, Ion implementations _MAY_ impose 
+upper bounds on the size of symbol tables, macro tables, module bindings, and any other direct or indirect component of
+the encoding context.
```
tgregg, line 161: I'd suggest that an implementation may also allow the user to configure size limits for any kind of value, including user values.

tgregg (approved):
This is a good read.

_books/ion-1-1/src/security-considerations.md
popematt: Yes, that is my intention—that every value that is produced at _any_ level in a macro evaluator would count, and perhaps that any macro that evaluates to nothing also uses one token to produce nothing.

I don't know how prescriptive we need to be. The point is to set a limit on the expansion to prevent expansion bombs, so I don't think Ion 1.1 libraries need to implement this in exactly the same way because we don't necessarily need to have a precise limit. (After all, if your data looks like an expansion bomb, then you might need to rethink your data.)

_books/ion-1-1/src/security-considerations.md
popematt: I've added it to https://github.com/amazon-ion/ion-tests/issues/88

zslayton (approved):

PR #367, opened 2024-11-18
Update float.md links

popematt:
### Issue #, if available:

None

### Description of changes:

The link for Clinger's algorithm was broken, and the other links had permanent redirects. I have updated two of the references to point to the ACM page for that paper (available for free as PDF and EPUB). For the "Correctly Rounded Binary-Decimal and Decimal-Binary Conversions" paper, I was unable to find a better source since only the abstract seems to be available on the Bell Labs website (see [here](https://www.bell-labs.com/institute/publications/bl9100445/)), and it's not available from ACM, arXiv, or IEEE Xplore. I'll try to determine whether we can host a copy of it on our own website so that we're not in danger of the link breaking for that one.

----

### DIFF
--- a/docs/float.md
+++ b/docs/float.md
@@ -108,6 +108,6 @@ the following text forms all map to the same `binary64` value:
 
 <!-- References -->
 [docs]: {{ site.baseurl }}/docs.html
-[rounding]: http://ampl.com/REFS/rounding.pdf
-[clinger]: http://www.cesura17.net/~will/professional/research/papers/howtoread.pdf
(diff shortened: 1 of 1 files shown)


jobarr-amzn (approved):

popematt:
Follow up on the paper by David M Gay—it looks like the author was [involved in the creation of AMPL](https://dl.acm.org/doi/10.5555/2804729.2804730), so the AMPL website is probably a safe place to continue linking for that paper.

PR #381, opened 2025-02-15
Adds chapters on text encoding

popematt:
### Issue #, if available:

N/A

### Description of changes:

Adds chapters about the text encoding to the Ion 1.1 specification. They are mostly lifted (with some editing and rearranging) from parts of the Ion 1.0 spec, but has an added section on E-expressions.

See it here: https://popematt.github.io/ion-docs/books/ion-1-1/text/encoding.html

----


_books/ion-1-1/src/text/values.md
```diff
@@ -0,0 +1,489 @@
+# Values
+
+## Annotations
+In the text format, type annotations are denoted by a [symbol token](symbol-tokens.md) and double-colons preceding any value.
```
tgregg, line 4: We should probably note that e-expressions and expression groups are not values, and cannot be annotated.

_books/ion-1-1/src/text/values.md
```diff
@@ -0,0 +1,489 @@
...
+int32::12                                // Suggests 32 bits as end-user type
+degrees::'celsius'::100                  // You can have multiple annotaions on a value
+'my.custom.type'::{ x : 12 , y : -1 }    // Gives a struct a user-defined type
+
+{ field: some_annotation::value }        // Field's name must precede annotations of its value
+
+jpeg :: {{ ... }}                        // Indicates the blob contains jpeg data
+bool :: null.int                         // A very misleading annotation on the integer null
+'' :: 1                                  // An empty annotation
+null.symbol :: 1                         // ERROR: type annotation cannot be null
```
tgregg, line 17: Should add ERROR examples for annotations on an e-expression and expression group.

_books/ion-1-1/src/text/values.md
```diff
@@ -0,0 +1,489 @@
...
+'true'::1     // an integer annotated with the text "true"
+true::1       // ERROR: cannot use an unquoted keyword as an annotation
+
+{ 'true': 1 } // a struct containing a field name with the text "true"
+{ true: 1 }   // ERROR: cannot use an unquoted keyword as a field name
+```
+
+## Integers
+
+Integer values may be encoded in binary, decimal, and hexadecimal notation.
```
tgregg, line 82: Should we add some note in this section that integers provided as tagless arguments have limited range?

_books/ion-1-1/src/text/values.md
```diff
@@ -0,0 +1,489 @@
...
+{ first : "Tom" , last: "Riddle" }  // Structure with two fields
+{"first":"Tom","last":"Riddle"}     // The same value with confusing style
+{center:{x:1.0, y:12.5}, radius:3}  // Nested struct
+{ x:1, }                            // Trailing comma is legal in Ion (unlike JSON)
+{ "":42 }                           // A struct value containing a field with an empty name
+{ x:1, x:null.int }                 // WARNING: repeated name 'x' leads to undefined behavior
+{ x:1, , }                          // ERROR: missing field between commas
+```
+
+Note that field names are symbol *tokens*, not symbol *values*, and thus may not be annotated.
```
tgregg, line 485: There should probably be some mention of using an e-expression in field name position.

_books/ion-1-1/src/text/encoding.md
```diff
@@ -0,0 +1,10 @@
+# Ion 1.1 Text Encoding
+
+The Ion text encoding is a stream of UTF-8 encoded text.
+It is intended to be easy to read and write by humans in any text-based application.
+
+Whitespace is insignificant and is only required where necessary to separate tokens, and C-style comments 
+(either block or in-line) are treated as whitespace (and are not part of the binary encoding).
+
+A text Ion 1.1 stream begins with the Ion 1.1 [version marker](../todo.md) (`$ion_1_1`) followed by a series of
```
tgregg, line 10: In the binary section I think I linked to the glossary for "Ion version marker". Do you think it needs its own page or section somewhere?

_books/ion-1-1/src/text/e_expressions.md
```diff
@@ -0,0 +1,91 @@
...
+
+Values in the E-expression body follow the same syntax as values in an [S-expression](values.md#s-expressions) body.
+E-expressions are not values, so they may not be annotated.
+
+```ion
+(:pi)              // Invokes the macro 'pi'
+(:1)               // Invokes the macro with address 1 in the macro table
+(:constants::pi)   // Invokes the macro 'pi' from the module 'constants'
+
+(: pi)             // ERROR: whitespace is not permitted between '(:' and the macro reference
```
tgregg, line 24: Add an ERROR example for an annotated e-expression.

_books/ion-1-1/src/text/e_expressions.md
```diff
@@ -0,0 +1,91 @@
...
+followed by a macro reference, which must be one of:
+ * a base-10 integer macro address
+ * a macro name
+ * a qualified macro name consisting of a module name, double-colon (`::`), and the macro name. (See [Encoding modules](../modules/encoding_modules.md) for details about qualified macro references.)
+
+Macro and module names follow the syntax rules for _identifier_ [symbol tokens](symbol-tokens.md), excluding _symbol identifiers_.
+There may not be any whitespace from the start of the E-expression through to the end of the macro reference.
+
+Values in the E-expression body follow the same syntax as values in an [S-expression](values.md#s-expressions) body.
+E-expressions are not values, so they may not be annotated.
```
tgregg: Should we mention *how* to annotate the result of an e-expression?

_books/ion-1-1/src/text/e_expressions.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+### Expression Groups
+
+To pass multiple arguments to a single macro parameter, the arguments to the parameter must be delimited by an _Expression Group_.
+
+Inside an E-expression, an expression group starts with `(::`.
+The remainder of the expression group uses the same syntax as an [S-expression](values.md#s-expressions).
+Expression groups are not values, so they may not be annotated.
+Expression groups may not have another expression group as the direct
```
tgregg: Sentence is cut off

tgregg (commented):
Looks good; a few simple suggestions.

_books/ion-1-1/src/text/encoding.md
```diff
@@ -0,0 +1,10 @@
+# Ion 1.1 Text Encoding
+
+The Ion text encoding is a stream of UTF-8 encoded text.
+It is intended to be easy to read and write by humans in any text-based application.
```
zslayton: I may be missing the distinction that the latter clause is making?

```suggestion
It is intended to be easy to read and write by humans.
```

_books/ion-1-1/src/text/encoding.md
```diff
@@ -0,0 +1,10 @@
...
+
+Whitespace is insignificant and is only required where necessary to separate tokens, and C-style comments 
+(either block or in-line) are treated as whitespace (and are not part of the binary encoding).
```
zslayton: Even though it's correct, I think mentioning the binary encoding here is potentially confusing.
```suggestion
Whitespace is insignificant and is only required where necessary to separate tokens.
C-style comments (either block or in-line) are treated as whitespace;
they are not part of the data model and implementations are not required to preserve them.
```

_books/ion-1-1/src/text/values.md
```diff
@@ -0,0 +1,489 @@
...
+## Integers
+
+Integer values may be encoded in binary, decimal, and hexadecimal notation.
+
+A decimal-encoded `int` consists of the digit `0` OR a non-zero digit followed by zero-or more base 10 digits (`0123456789`)—leading zeros are not allowed.
+A binary-encoded `int` consists of `0b` followed by one or more base 2 digits (`01`).
+A hexadecimal-encoded `int` consists of `0x` followed by one or more case-insensitive base 16 digits (`0123456789abcdefABCDEF`).
+
+All integer values may be preceded by an optional minus sign (`-`), indicating that the value is negative.
+(The token `-0` is legal and equivalent to `0`.)
```
zslayton: Maybe suggest that users needing to preserve `-0` can use `decimal` or `float` instead?

_books/ion-1-1/src/text/values.md
```diff
@@ -0,0 +1,489 @@
...
+This mode/tie-break is the common default used in most programming environments and is discussed in detail in
+["Correctly Rounded Binary-Decimal and Decimal-Binary Conversions"](https://ampl.com/_archive/first-website/REFS/rounding.pdf).
+This conversion algorithm is illustrated in a straightforward way in [Clinger's Algorithm](https://dl.acm.org/doi/10.1145/93548.93557).
+
+When encoding a `float` value to Ion text, an implementation MAY want to consider the approach described in
+["Printing Floating-Point Numbers Quickly and Accurately"](https://dl.acm.org/doi/10.1145/231379.231397).
+
+
+#### Examples
+Although the textual representative of `1.2e0` itself is irrational, its
```
zslayton:
```suggestion
Although the textual representation of `1.2e0` itself is irrational, its
```
?

_books/ion-1-1/src/text/values.md
```diff
@@ -0,0 +1,489 @@
...
+In the text notation, decimal values must be followed by one of the
+fifteen numeric stop-characters: `{}[](),\"\'\ \t\n\r\v\f`.
+
+
+## Timestamps
+
+In the text format, timestamps follow the [W3C note on date and time formats](https://www.w3.org/TR/NOTE-datetime),
+but they must end with the literal `T` if not at least whole-day precision.
+Fractional seconds are allowed, with at least one digit of precision and an unlimited maximum.
+Local-time offsets may be represented as either hour:minute offsets from UTC, or as the literal `Z` to denote a local time of UTC.
```
zslayton, line 267:
```suggestion
Local-time offsets may be represented as either hour:minute offsets from UTC, or as the literal `Z` to denote a local time of UTC.
If the offset is `-00:00`, it indicates that the local offset in which the timestamp was recorded is unknown, and that the time is therefore encoded as UTC.
```

_books/ion-1-1/src/text/values.md
```diff
@@ -0,0 +1,489 @@
...
+In the text format, lists are bounded by square brackets and elements
+are separated by commas.
+
+```ion
+[]                // An empty list value
+[1, 2, 3]         // List of three ints
+[ 1 , two ]       // List of an int and a symbol
+[a , [b]]         // Nested list
+[ 1.2, ]          // Trailing comma is legal in Ion (unlike JSON)
+[ 1, , 2 ]        // ERROR: missing element between commas
```
zslayton, line 441: Maybe throw in an example that has a nested e-expression?

_books/ion-1-1/src/text/e_expressions.md
```diff
@@ -0,0 +1,91 @@
+# E-expressions
+
+In Ion text, encoding expressions ([E-expressions](../macros/macros_by_example.md)) start with `(:`, immediately 
+followed by a macro reference, which must be one of:
+ * a base-10 integer macro address
+ * a macro name
+ * a qualified macro name consisting of a module name, double-colon (`::`), and the macro name. (See [Encoding modules](../modules/encoding_modules.md) for details about qualified macro references.)
```
zslayton: This can also be a qualified address (`foo::17`).

_books/ion-1-1/src/text/values.md
popematt: I can mention it, but it's not specific to the text encoding. I tried to keep this mostly focused to the encoding itself.

_books/ion-1-1/src/text/values.md
popematt: That's covered in the chapter on Macros.

_books/ion-1-1/src/text/e_expressions.md
popematt: I thought that was only in TDL, not for E-expressions. For E-Expressions, there is a single macro table that is defined by the encoding context, right?

_books/ion-1-1/src/text/encoding.md
popematt: I think we need a section that defines the semantics of an Ion stream. By that I mean a chapter that covers things like IVMs, system values vs user values, and `$ion_literal`.

_books/ion-1-1/src/text/values.md
popematt: I will add it to the e-expression page.

_books/ion-1-1/src/text/e_expressions.md
zslayton: The macro table is a concatenation of the encoding modules' respective macro tables. You can use a qualified address to reference a macro by its address within a module provided that the module is in the encoding module sequence. (It looks like we don't show an example of this in the [_Encoding modules_](https://amazon-ion.github.io/ion-docs/books/ion-1-1/modules/encoding_modules.html) chapter, however. We only have examples for qualified names, not qualified addresses.)

tgregg (approved):

Issue #298, opened 2024-02-21
Finalize the Ion 1.1 specification


tgregg:

toddjonker:
Here are some thing that I think are (some) requirements for "finalizing" the spec.

* Merge the denotational semantics document from its home at https://github.com/toddjonker/ion-docs/commits/main/ into the `ion-11-specification` branch with the narrative spec doc.
* Finish the denotational semantic model, so we have a formal spec and reference implementation for the dynamic behavior.
  * We should reach consensus on what it would take to make the semantic model the normative standard. That is, treat it as the primary specification such that the old prose documents (particularly the symbols page) are supplemental explanations.
* Merge the latter branch into the `gh-pages` branch
  * Automate building the PDFs and making them available for viewing online.
* Personally, I think the spec shouldn't be considered finished without a fairly complete suite of conformance tests.

This list is not strongly-ordered and is not a complete list of requirements.  For example, the narrative doc still needs a fair amount of editing for consistency/correctness and has some notable gaps.

PR #390, opened 2025-04-16
Update Ion 1.1 spec with feedback from final comment period

popematt:
### Issue #, if available:

#298 

### Description of changes:

* Changes the order of the symbol and macro tables in a module definition
* Changes `symbol_table` and `macro_table` to `symbols` and `macros` respectively
* Clarifies that `use`, `set_symbols`, etc. can only be invoked in locations where system values can appear
* Remove `symbol_table` from the system symbols and renumber the symbols.

I also split `lob.md` into `blob.md` and `clob.md` because mdbook was complaining about a duplicate link in the `SUMMARY.md` file.

----


_books/ion-1-1/src/modules.md
```diff
@@ -22,10 +22,10 @@ The interface to a module consists of:
 
 The spec version is external to the module body and the precise way it is determined depends on the type of module being defined. This is explained in further detail in [Module Versioning](#module-versioning).
 
-The exported symbol array is denoted by the `symbol_table` clause of a module definition, and
+The exported symbol array is denoted by the `symbols` clause of a module definition, and
 by the `symbols` field of a shared symbol table.
 
-The exported macro array is denoted by the module’s `macro_table` clause, with addresses
+The exported macro array is denoted by the module’s `macros` clause, with addresses
 allocated to macros or macro bindings in the order they are declared.
```
zslayton: I suggest swapping the order of these two statements to align with the order of the clauses. It would be good to explicitly note that new symbols defined in the `symbols` clause are not yet available in the preceding `macros` clause.

zslayton (approved):
👌

PR #392, opened 2025-07-17
Prefer Finch over Docker, if it's available.

toddjonker:
### Description of changes:

In general I prefer Finch over Docker, and since it's Amazon-owned, it feels reasonable to prefer it in the build logic for this project.

----

### DIFF
--- a/build-docker.sh
+++ b/build-docker.sh
@@ -11,6 +11,14 @@ IMAGE_TAG="ion-spec-asciidoc"
 #DOCKER_ARGS="--platform linux/amd64"
 DOCKER_ARGS=""
 
+
+# http://stackoverflow.com/a/677212
+if hash finch 2>/dev/null
+then
+    function docker() { finch "$@"; }
+fi
+
+
 function build_image()
 {
   run_check docker build ${DOCKER_ARGS} \
--- a/README.md
+++ b/README.md
@@ -4,8 +4,8 @@ This repository holds the draft proposal for Ion 1.1.
 
 ## Building
 
-The easiest way to build the spec is to use the included `Dockerfile`.  Assuming you have `docker` you can run the build
-script in the project directory as follows:
+The easiest way to build the spec is to use the included `Dockerfile`.  Assuming you have `finch` or `docker` on your
+`PATH`, you can run the build script in the project directory as follows:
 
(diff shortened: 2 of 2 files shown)


build-docker.sh
```diff
@@ -11,6 +11,14 @@ IMAGE_TAG="ion-spec-asciidoc"
 #DOCKER_ARGS="--platform linux/amd64"
 DOCKER_ARGS=""
 
+
+# http://stackoverflow.com/a/677212
+if hash finch 2>/dev/null
+then
+    function docker() { finch "$@"; }
+fi
```
popematt, line 19: I'm happy to support Finch, but this does give me pause:

> Finch is a new project and not meant to be a direct drop-in replacement for other tools. Therefore, we don't recommend aliasing or linking other command names to finch. ([source](https://github.com/runfinch/finch?tab=readme-ov-file#finch-and-other-tools))

Can we pick one of the two instead of trying to support both?

toddjonker: My reading is that Finch doesn't promise to replace everything Docker does.  But the needs here are simple, and it seems to work fine.  We're not making a general alias or globally linking, we're just invoking it from a specific, controlled code path.

I'm not sure what "supporting" means in this case of a team-internal project; we can't break anyone but ourselves, and even then the workaround is trivial.

popematt: I guess what I'd like to know is why not just change it to use Finch unconditionally? (It's quite common to require specific build tools—e.g. `cargo`, `pip`—so I don't have any issue with doing the same for Finch here.)

tgregg (approved):

build-docker.sh
nirosys: I can understand not wanting to force a container runtime in general. Both docker and finch require VMs on Mac and Windows (without WSL), which is a bit of an ask if you're already using one and not the other. It's not quite the same as pip and cargo imo. Given the scope of this repo though I wouldn't consider it as much of a concern.

toddjonker: @nirosys Yeah, I figured that individuals reasonably prefer various OCI stacks based on OS and other concerns.

Personally, I find Docker Desktop on macOS to be flaky and heavyweight.

PR #393, opened 2025-07-24
Remove wording around macro name omission

nirosys:
### Description of changes:
As discussed earlier this week, the current implementations do not allow a macro name to be omitted, and instead rely on `null` to be used as the macro name for anonymous macros. Since macros are intended to be used by machinery, auto-generating and consuming macros, the inclusion of `null` is not seen as a readability concern, and the savings of removing it does not outweigh the parsing complexity.

This PR removes the wording around macro names being omitted, and replaces it with using the value `null` to indicate an anonymous macro.

----

### DIFF
--- a/_books/ion-1-1/src/macros/defining_macros.md
+++ b/_books/ion-1-1/src/macros/defining_macros.md
@@ -32,7 +32,7 @@ A macro is defined using a `macro` clause within a [module](../modules.md)'s [`m
 Syntactically, macro names are [identifiers](../text/symbol-tokens.md). Each macro name in a macro table must be unique.
 
 In some circumstances, it may not make sense to name a macro. (For example, when the macro is generated automatically.)
-In such cases, authors may use the symbol `null` to indicate that the macro does not have a name.
+In such cases, authors must use `null` to indicate that the macro does not have a name.
 Anonymous macros can only be referenced by their address in the macro table.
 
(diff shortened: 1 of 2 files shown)


_books/ion-1-1/src/macros/defining_macros.md
```diff
@@ -32,7 +32,7 @@ A macro is defined using a `macro` clause within a [module](../modules.md)'s [`m
 Syntactically, macro names are [identifiers](../text/symbol-tokens.md). Each macro name in a macro table must be unique.
 
 In some circumstances, it may not make sense to name a macro. (For example, when the macro is generated automatically.)
-In such cases, authors may omit the macro name to indicate that the macro does not have a name.
+In such cases, authors may use the symbol `null` to indicate that the macro does not have a name.
```
tgregg:
```suggestion
In such cases, authors may use `null` to indicate that the macro does not have a name.
```

tgregg (approved):

_books/ion-1-1/src/macros/defining_macros.md
popematt: Since we're trying to remove ambiguity here, should we say "In such cases, authors must use `null` ..."
(Specifically, the use of **must** rather than **may** because it is the require way for doing "such cases")

Issue #397, opened 2025-09-19
Can we simplify Ion 1.1?


popematt:
_**TLDR;** Ion 1.1, when considering the costs and benefits from a business perspective, does not seem to be a feasible product, and the specification itself seems to be at the root of key performance issues. Can we simplify Ion 1.1, or skip Ion 1.1 and create a simpler Ion 1.2?_

----

Attempts to implement an Ion 1.1 reader in `ion-java` have run into issues with performance of macro evaluation. I (and others) have tried multiple different strategies for implementing a macro evaluator that fits in the current `IonReader` APIs. We have also tried implementation strategies that are unconstrained by the existing `IonReader` API. In all of our attempts, we have not been able to get Ion 1.1 with macros to reach the read throughput of Ion 1.0 despite multiple people working on this for over a year.

The read throughput is a significant problem—we don't think anyone will actually use Ion 1.1 macros if it results in a 40% or more drop in read throughput. Furthermore, the complexity of the specification makes it difficult and time-consuming to iterate on different implementation ideas. This is anecdotal, but I have made a few different Ion 1.0 binary readers in the last few years, and the time for me to do that is usually on the order of 1-2 days to get something reasonably performant. Attempting to implement the current Ion 1.1 spec takes significantly longer (several weeks) just to get to simple module management and basic template macro evaluation, and there's a long tail of several months to complete all of the system macros and special forms.

For a long time, we thought that the problem was in the way we were implementing it in `ion-java` because the `ion-rust` experimental 1.1 implementation benchmarks seemed to prove that there as a high performance ceiling. However, we recently realized that the measurements we were using were not up-to-date and did not accurately reflect the current Ion 1.1 specification; `ion-rust` suffers from similar performance issues as `ion-java`.

Rather than throw away Ion 1.1 entirely, I attempted an experiment with a simplified Ion templates implementation that's closer to [RFC: Ion Templates](https://github.com/amazon-ion/ion-docs/blob/cd844cb595261189c3980981dcaf2fe7058a0eb4/rfcs/ion_templates.md).

Not only did it take only about 1 week to implement (without even vibe-coding any of it), prior to any performance optimization, the read throughput was comparable to Ion 1.0 read throughput. Cleaning up the obvious bits of lazy/bad, slow code resulted in cases where the simplified templates were ~15% better read throughput that Ion 1.0 and 100-200% better read throughput than our best attempts at implementing the current Ion 1.1 specification.

[...]

PR #399, opened 2025-10-02
Updates the Ion 1.1 macros pages

tgregg:
### Issue #, if available:
#397 

### Description of changes:
Although some other files were touched, this PR focuses on updating the following files:
* macros.md
* macros/defining_macros.md
* macros/macros_by_example.md
* replacing macros/system_macros.md with macros/directives.md

Other files likely require more changes that will be covered in a future PR. We will also need a follow-up PR once we have decided on the final set of allowed tagless types.

----


_books/ion-1-1/src/security-considerations.md
```diff
@@ -2,127 +2,14 @@
...
-                    (:repeat 2147483647 "abc")))))))))))
-```
-The expansion of these e-expressions results in a stream of ~450 [googol](https://en.wikipedia.org/wiki/Googol) string
-values. Any attempt to hold all of this in memory or write it to disk will exhaust all available resources and 
-eventually fail. Even an attempt to count the length of the stream, while it may theoretically succeed if using an 
-appropriate `BigInteger` type, will require a considerable amount of CPU operations (over a googol), and even the 
-fastest processors will require many millennia to completely count the number of values in the stream.
-
-Even without using `repeat` or `for`, a [Billion laughs attack](https://en.wikipedia.org/wiki/Billion_laughs_attack)
-could exist for any data format with macro expansion, and it is certainly possible with Ion 1.1.
```
popematt, line 48: It is still possible.
```
(:$add_macros (lol0 "lol"))
(:$add_macros (lol1 [(:lol0), (:lol0), (:lol0), (:lol0), (:lol0), (:lol0), (:lol0), (:lol0), (:lol0), (:lol0),]))
(:$add_macros (lol2 [(:lol1), (:lol1), (:lol1), (:lol1), (:lol1), (:lol1), (:lol1), (:lol1), (:lol1), (:lol1),]))
(:$add_macros (lol3 [(:lol2), (:lol2), (:lol2), (:lol2), (:lol2), (:lol2), (:lol2), (:lol2), (:lol2), (:lol2),]))
// ...

(:lol9)
```

It can still produce a billion `"lol"`. This time it has some list structure around it, but it doesn't change the basic ability to perform this sort of attack. A simpler mitigation is possible though—we can have a cap on the size of any one macro or perhaps the macro table as a whole.

_books/ion-1-1/src/macros/macros_by_example.md
```diff
@@ -155,720 +132,160 @@ The outer e-expression receives the results from the expansion of the inner e-ex
...
-    // │       └─── stream to map over
-    // └─────────── variable name
-
-    // Template
-    (.price (%amt) (%currency))
-  )
+(:$set_macros
+  (prefix_suffix [(:?), (:?)])
+  (website_url (:prefix_suffix "https://www.amazon.com/" (:?)))
 )
```
popematt: The `prefix_suffix` macro is not available in the macro table until after the directive closes, so it would need to be:
```suggestion
(:$set_macros (prefix_suffix [(:?), (:?)]))
(:$add_macros (website_url (:prefix_suffix "https://www.amazon.com/" (:?))))
```

_books/ion-1-1/src/macros/macros_by_example.md
```diff
@@ -155,720 +132,160 @@ The outer e-expression receives the results from the expansion of the inner e-ex
...
-                                       {amount:4, currency:GBP}
-```
-
-Expression groups may only appear inside macro invocations where the corresponding
-parameter has `?`, `*`, or `+` cardinality.
-There is no binary opcode for these constructs; the encoding uses a tagless format to keep
-things as dense as possible.
-As usual, the text format mirrors this constraint.
+The `website_url` macro's template contains an E-expression `(:prefix_suffix ...)` which is
+expanded when the macro is defined.
```
popematt, line 166: Maybe add an example to show the expansion in the macro definition.
```
(:$add_macros (website_url (:prefix_suffix "https://www.amazon.com/" (:?))))  
  ⇒ (:$add_macros (website_url ["https://www.amazon.com/", (:?)]))  
```

_books/ion-1-1/src/macros/macros_by_example.md
```diff
@@ -155,720 +132,160 @@ The outer e-expression receives the results from the expansion of the inner e-ex
...
-the type tags are unnecessary and may add up to a non-trivial amount of wasted space.
-Furthermore, the overhead for each value also includes length information: encoding an
-octet of data takes two octets on the stream.
-
-Ion 1.1 tries to mitigate this overhead in the binary format by allowing macro parameters to use
-more-constrained _tagless types_. These are subtypes of the concrete types,
-constrained such that type tags are not necessary in the binary form. In general this can shave
-4-6 bits off each value, which can add up in aggregate. In the extreme, that octet of data can
-be encoded with no overhead at all.
+## Available Tagless Types
```
popematt: This sort of stuff is probably going to be repeated in the binary encoding and text encoding chapters, and it's not specific to macros now that we have tagless-element sequences. Let's just link to a `tagless_encodings.md` page. (It can be left as `todo.md` until that page is created.)

_books/ion-1-1/src/macros/macros_by_example.md
```diff
@@ -155,720 +132,160 @@ The outer e-expression receives the results from the expansion of the inner e-ex
...
 ```ion
-(:point null.int 17)   ⇒ // Error: primitive flex_int does not accept nulls
-(:point a::3 17)       ⇒ // Error: primitive flex_int does not accept annotations
-(:point (:values 1) 2) ⇒ // Error: cannot use macro for a primitive argument
+(:annotated_value 42) ⇒ [foo::42]
 ```
 
-While Ion text syntax doesn’t use tags—the types are built into the syntax—these errors ensure
-that a text E-expression may only express things that can also be expressed using an equivalent
-binary E-expression.
```
popematt, line 756: This would be useful to keep.

_books/ion-1-1/src/macros/macros_by_example.md
```diff
@@ -155,720 +132,160 @@ The outer e-expression receives the results from the expansion of the inner e-ex
...
-```ion
-(:point 3 17) ⇒ {x:3, y:17}
+(annotated_value [foo::(:?)])
 ```
 
-The tagless encoding has no real benefit here in text, as primitive types aim to improve the binary
-encoding.
-
-This density comes at the cost of flexibility. Primitive types cannot be annotated or null, and
-arguments cannot be expressed using macros, like we’ve done before:
```
popematt, line 746: These sentences are still good to keep. (Although not in the middle of the annotated placeholders section.)

_books/ion-1-1/src/macros/macros_by_example.md
```diff
@@ -155,720 +132,160 @@ The outer e-expression receives the results from the expansion of the inner e-ex
...
 
-> [!TIP]
-> The primary goal of macro-shaped arguments, and tagless types in general, is to increase
-> density by tightly constraining the inputs.
+- **Templates are Ion values** with placeholders marking where arguments go
+- **Tagged placeholders** `(:?)` accept any Ion value and can have defaults
+- **Tagless placeholders** `(:?\type\)` require specific types (from the enumerated set) but enable compact encoding
+- **E-expressions in templates** are expanded at definition time to help construct the template
+- **Binary encoding** makes macro invocations extremely compact
+- **Templates cannot consist solely of a placeholder** (annotated or unannotated)
```
popematt: Not a blocker; just a feeling I have—a spec doesn't seem like the right place for this sort of summary.
 Instead of a summary on individual pages, it might be more helpful to have a spec-wide "cheat sheet" sort of reference page.

_books/ion-1-1/src/modules/directives.md
```diff
@@ -78,4 +80,4 @@ $ion::
 
 The new encoding module sequence takes effect immediately after the directive and remains the same until the next `encoding` directive or Ion version marker.
 
-Note that the [default module](encoding_modules.md#the-default-module) is always implicitly at the head of the encoding module sequence.
\ No newline at end of file
+Note that the [default module](encoding_modules.md#the-default-module) is always implicitly at the head of the encoding module sequence.
```
popematt: Is there a difference here?

_books/ion-1-1/src/modules/directives.md
```diff
@@ -16,9 +16,11 @@ $ion::
...
 
 In Ion 1.1, there are three supported directive operations:
-1. [`module`](#module-directives)
-2. [`import`](#import-directives)
-3. [`encoding`](#encoding-directives)
+- [Directives](#directives)
+  - [Top-level bindings](#top-level-bindings)
+    - [`module` directives](#module-directives)
+    - [`import` directives](#import-directives)
+  - [`encoding` directives](#encoding-directives)
```
popematt: This is not accurate.

Edit: sorry. actually, it might be correct.

_books/ion-1-1/src/macros/macros_by_example.md
popematt, line 1: It would be good to mention somewhere in this chapter that macros can be annotated, and that the annotations are prepended to the macro result. E.g.:
```
// Given
(:foobar) ⇒ foo::"bar"
// Then
baz::(:foobar) ⇒ baz::foo::"bar"
```

(And similarly for annotated placeholders)

_books/ion-1-1/src/macros/directives.md
popematt, line 1: I think this content should probably be left for the `modules/directives.md` chapter so that we're not duplicating things.

_books/ion-1-1/src/modules/directives.md
tgregg: Reverting changes to this file since they're superseded by https://github.com/amazon-ion/ion-docs/pull/400

popematt (approved):

Issue #176, opened 2022-09-09
Clarify the associativity of `+` and `-` in s-expressions


popematt:
These behaviors are present in `ion-java`, `ion-js`, and `ion-rust`. (I did not check any other implementations.) 

```
( +inf    )  // parses as ( +inf )
( +inf'+' )  // parses as ( +inf '+' )
( +inf+   )  // parses as ( '+' 'inf' '+' ) <= WAT?
```
This is also holds if you substitute `-` for any `+` or if you substitute any other operator symbol in place of the `+` after the `inf`.

Also:
```
( +inf+inf )  // parses as ( '+' 'inf' +inf ) <= WAT?
```

So, why does the addition of an unquoted operator symbol or another infinity _after_ `+inf` cause `+inf` to split into `'+' 'inf'`?

rmarrowstone:
Another example:
`(2000T -2001T)`

Is not valid per the ANTLR grammar (Ion doesn't allow negative timestamps), but it _could_ be parsed as:
`(2000T '-' 2001T)`

popematt:
Another potentially confusing case:
```
(x+1)  // ( 'x' '+' 1 )
(x-1)  // ( 'x' -1 )
(x*-1) // ( 'x' '*-' 1 )
```
The `-` associates with the number, unless it is preceded by another operator symbol character.

Issue #175, opened 2022-09-08
Clarify whether numeric stop characters are required after special float values


popematt:
The Ion Spec says, about `float` and `decimal`:
> In the text notation, real values must be followed by one of the fifteen numeric stop-characters: `{}[](),\"\'\ \t\n\r\v\f`.

However, `nan`, `+inf`, and `-inf` are not real numbers (i.e. not in ℝ). Furthermore, the existing implementations are not in agreement about whether a numeric stop character is required after these values.

| Input | `ion-java` | `ion-js` | `ion-rust` | 
| --- | --- | --- | --- |
|`(nan-nan)`|`(nan '-' nan)`|`(nan '-' nan)`|invalid|
|`nan-inf`|`nan -inf`|`nan -inf`|invalid|
|`nan+inf`|`nan +inf`|`nan +inf`|invalid|
|`+inf+inf`|invalid|invalid|invalid|

There are inconsistencies between `ion-rust` and the other two implementations, and implementations seem to be inconsistent within themselves because if `nan+inf` is valid, then it seems like `+inf+inf` should also be valid.

rmarrowstone:
To make a question explicit: the Spec says "real values" as to whether the intention was "real numbers" (in the mathematical sense) or "real values" (as in not-null?) is not clear.

PR #401, opened 2025-10-09
Specify requirements for numeric stop chars and other token boundaries

austnwil:
### Description of changes:

There are numerous open issues involving ambiguity/incorrect behavior around how parsers should split certain tokens in Ion text (see #176, #175, [amazon-ion/ion-rust#414](https://github.com/amazon-ion/ion-rust/issues/414), [amazon-ion/ion-js#724](https://github.com/amazon-ion/ion-js/issues/724)). These have caused confusing behavior and inconsistencies between and even within implementations on how certain tokens that don't have any whitespace between them are split. A common theme across some of these issues is that the spec is not very clear on requirements around numeric stop characters and behavior of `+`/`-` in `sexp`s.

This PR adds some explicit documentation on when tokens can appear next to each other and what the behavior of `+`/`-` in sexp is (subject to change). Once this is cleared up it should be easier to fix some of the linked issues because it will be more clear whether the current behavior is correct or not (particularly with #175 and #176).

----

### DIFF
--- a/docs/boundaries.md
+++ b/docs/boundaries.md
@@ -1,0 +1,119 @@
+---
+redirect_from: "/boundaries.html"
+title: Token Boundaries in Ion Text
+description: "Details regarding how Ion text is tokenized and where values are split in cases where no whitespace is used between them."
+---
+
+# [Docs][docs]/ {{ page.title }}
+
+Ion text values are best separated by explicit whitespace (space, horizontal tab, vertical tab, line feed, carriage return, form feed, and comments). However, the specification defines circumstances in which it is acceptable for values to appear immediately adjacent to each other.
+
+In general, Ion text values can appear adjacent to each other when the boundary between the values is unambiguous. The boundaries between quoted symbols, double-quoted strings, triple-quoted strings, lobs, structs, lists, and S-expressions are unambiguous, so these elements can appear adjacent to one another in Ion text:
+
+```ion
+{% raw %}
+// A valid Ion data stream
+(1)[2]{a:struct}'hello'"world"{{ SSBsb3ZlIElvbiE= }}{{ "Same with clob" }}'annotations too!'::123
+{% endraw %}
+```
+
+Additionally, any value can appear immediately *after* a value of these types, including numeric values, keywords, and identifiers.
+
+Container separators delimit values unambiguously, so any value can also appear adjacent to a container boundary: 
+
+```ion
+// Some valid, compact containers
+[1,2,"three"]
+{name:"Austin",message:"Hello world!"}
+```
+
+Indentifiers (unquoted symbols) and reserved keywords (`null` and its typed variants, `true`, `false`, `nan`, `-inf`, and `+inf`) are terminated by the first non-identifier character encountered. The identifier characters are ASCII letters, digits, or the characters `$` (dollar sign) or `_` (underscore). This allows symbols to contain a reserved keyword as a prefix:
+
+```ion
+// All single symbols, even when prefixed by a keyword
+trueSymbol
+null123
+nanfalse
+```
+
+This means that any value that begins with a non-identifier character can appear immediately after an identifier or reserved keyword:
+
+```ion
+// All valid pairs of two top-level values
+abc-5
+symbol["And a list"]
+anotherOne'andAnother'
+true(story)
+null.float+inf
+```
+
+Typed nulls and `+/-inf` do not form valid Ion if not followed by a non-identifier character:
+
+```ion
+null.struct5E10      // ERROR: not a valid Ion value - struct5E10 is a bad null decorator
++inftrue             // ERROR: this also is invalid
+```
+
+## Numeric stop-characters
+
+Ion text enforces stricter rules on certain "numeric" types. Ion text enforces that integers, real values (decimals and floats, excluding the special float values `nan`, `-inf`, and `+inf`), and timestamps must be followed by one of the following 15 numeric stop-characters: `{}[](),"' \t\n\r\v\f`.
+
+This means that strings, quoted symbols, lobs, structs, lists, and S-expressions can appear immediately after a numeric or timestamp value in a top-level datagram or S-expression:
+
+```ion
+// All of this is valid Ion
+123{a: "struct"}456
+-0.5[a, list]
+5.34e9"then a string"
+0xdeadbeef'hello'::world
+[1,2,3]
+(1["list"]2000T{a: "struct"}0b010101("sexp")4)
+```
+
+Anything that is not a numeric stop-character appearing immediately after a numeric or timestamp value is a syntax error. This notably includes comments and S-expression operators:
+
+```ion
+123// a comment              // ERROR: single-line comment is not a valid numeric stop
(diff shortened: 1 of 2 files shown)


toddjonker:
Are these issues not clarified by the grammar? That's the proper spec for text parsing questions.

austnwil:
> Are these issues not clarified by the grammar? That's the proper spec for text parsing questions.

The grammar is not clear on what the behavior of `+`/`-` is in sexp's. [Here](https://amazon-ion.github.io/ion-docs/grammar/IonText.g4.txt) the grammar states:

> //   * There are some ambiguities involving symbol operators within s-expressions
> //      - Do '-' and '+' (possibly others) right or left associate?
> //      - Do symbol operators terminate special float symbols (nan, -inf, +inf)?

This has led to implementations handling ambiguous cases in ways that appear inconsistent.

I'm not suggesting we change the spec, just that we more clearly document what the correct behavior is, or determine it if it is not known. This should make it easier to fix the inconsistencies (or close the above issues if the behavior is actually correct).

Even if the grammar is unambiguous it would still be helpful to explicitly state the correct behavior in the docs as well.

austnwil:
Most implementations currently seem to require numeric stops after `+inf` and `-inf` but not `nan`. Assuming this is correct, this makes the behavior in #176 correct, although unintuitive, so that can be closed. I updated this PR to document this behavior.

When #175 was reported, rust parsed certain inputs differently from java and javascript; this no longer seems to be the case. Rust and java implementations still reject `+inf+inf` and accept `nan+inf`, but if the above holds, then that is also correct behavior, so this can be closed as well.

docs/boundaries.md
```diff
@@ -0,0 +1,131 @@
...
+0xdeadbeef'hello'::world
++inf{{ "<-- also works with -inf" }}
+(1["list"]2000T{a: "struct"}0b010101("sexp")4)
+{% endraw %}
+```
+
+Anything that is not a numeric stop-character appearing immediately after a numeric or timestamp value is a syntax error. This notably includes comments and S-expression operators:
+
+```ion
+123// a comment              // ERROR: single-line comment is not a valid numeric stop
```
popematt, line 77: This is surprising. Comments are explicitly called out in the spec as being equivalent to whitespace, and I believe that all of our implementations treat comments as a valid whitespace for the purposes of terminating a numeric value.

austnwil: I also found this surprising, but some implementations do reject Ion that uses a comment immediately after a numeric.

The spec does say specifically:

> In the text notation, real values [and int, ts] must be followed by one of the fifteen numeric stop-characters: {}[](),\"\'\ \t\n\r\v\f.

Which calls out specific whitespace characters ` \t\n\r\v\f` and excludes comments.

Here are comparisons of our Ion JS, Python, Java, and Rust impls. I did not test Go, C yet.

Handling of `123//comment`:

| Language   | Success         | Result                                                                                               |
|------------|-----------------|------------------------------------------------------------------------------------------------------|
| JavaScript | ❌: Error        | Error: invalid character after number                                                                |
| Python     | ❌: IonException | IERR_INVALID_SYNTAX                                                                                  |
| Java       | ✅               | <br>$ion_1_0<br>123                                                                                  |
| Rust       | ❌: Decoding     | invalid Ion syntax encountered<br>        offset=3<br>        buffer head=<//comment><br>        buffer tail=<//comment><br>        buffer len=9<br>         |

Handling of `123/*comment*/`:

| Language   | Success         | Result                                                                                               |
|------------|-----------------|------------------------------------------------------------------------------------------------------|
| JavaScript | ❌: Error        | Error: invalid character after number                                                                |
| Python     | ❌: IonException | IERR_INVALID_SYNTAX                                                                                  |
| Java       | ✅               | <br>$ion_1_0<br>123                                                                                  |
| Rust       | ❌: Decoding     | invalid Ion syntax encountered<br>        offset=3<br>        buffer head=</*comment*/><br>        buffer tail=</*comment*/><br>        buffer len=11<br>         |


Package versions used:

  | JS | Python | Java | Rust
-- | -- | -- | -- | --
Ion version | 5.2.1 | 0.13.0 | 1.11.11 | 1.0.0-rc.7

Here's the code I used to test:

Java impl:

```java
import com.amazon.ion.*;
import com.amazon.ion.system.IonSystemBuilder;
import com.amazon.ion.system.IonTextWriterBuilder;
import com.fasterxml.jackson.databind.ObjectMapper;
import java.util.HashMap;
import java.util.Map;

public class Main {
    public static void main(String[] args) {
        ObjectMapper mapper = new ObjectMapper();
        Map<String, Object> response = new HashMap<>();
        
        try {
            String input = args[0];
            IonSystem ION = IonSystemBuilder.standard().build();
            IonValue datagram = ION.getLoader().load(input);
            StringBuilder sb = new StringBuilder();
            IonWriter writer = IonTextWriterBuilder.pretty().build(sb);
            datagram.writeTo(writer);
            writer.close();
            
            response.put("success", true);
            response.put("result", sb.toString());
            System.out.println(mapper.writeValueAsString(response));
        } catch (Exception e) {
            response.put("success", false);
            response.put("errorType", e.getClass().getSimpleName());
            response.put("error", e.toString());
            try {
                System.out.println(mapper.writeValueAsString(response));
            } catch (Exception jsonE) {
                System.out.println("{\"success\":false,\"errorType\":\"JsonProcessingException\",\"error\":\"" + jsonE.toString() + "\"}");
            }
        }
    }
}
```

JS impl:

```js
import * as ion from 'ion-js'

try {
    const input = process.argv[2]
    const reader = ion.makeReader(input)
    const writer = ion.makePrettyWriter()
    writer.writeValues(reader)
    writer.close()

    const decoder = new TextDecoder("utf-8")
    console.log(JSON.stringify({
        success: true,
        result: decoder.decode(writer.getBytes())
    }))
} catch (e) {
    console.log(JSON.stringify({
        success: false,
        errorType: e.constructor.name,
        error: e.toString()
    }))
}

```

Python impl:

```py
#!/usr/bin/env python3
import sys
import json
import amazon.ion.simpleion as ion

try:
    input_str = sys.argv[1]
    data = ion.loads(input_str, single_value=False)
    result = ion.dumps(data, sequence_as_stream=True, binary=False, indent='  ')
    
    print(json.dumps({
        "success": True,
        "result": result
    }))
except Exception as e:
    print(json.dumps({
        "success": False,
        "errorType": type(e).__name__,
        "error": str(e)
    }))
```

Rust impl:

```rs
use ion_rs::*;
use serde_json::json;
use std::env;

fn main() {
    let args: Vec<String> = env::args().collect();
    
    match run(&args[1]) {
        Ok(result) => {
            println!("{}", json!({
                "success": true,
                "result": result
            }));
        }
        Err(e) => {
            println!("{}", json!({
                "success": false,
                "errorType": format!("{:?}", e).split('(').next().unwrap_or("Error"),
                "error": format!("{}", e)
            }));
        }
    }
}

fn run(input: &str) -> IonResult<String> {
    let elements = Element::read_all(input.as_bytes())?;
    let out: String = elements.encode_as(v1_0::Text.with_format(TextFormat::Pretty))?;
    Ok(out)
}
```

PR #405, opened 2025-10-14
Updates the Ion 1.1 text grammar.

tgregg:
### Issue #, if available:
#397 

### Description of changes:
This grammar was never exhaustive, but I tried to maintain about the same level of exhaustiveness as before. The biggest assumption here is that there is a variable called `value` that defines the grammar for any Ion value literal.

This should more or less align the grammar with the simplified Ion 1.1 spec, but I would appreciate any comments on correctness, factoring, or things I forgot.

----

### DIFF
--- a/_books/ion-1-1/src/grammar.md
+++ b/_books/ion-1-1/src/grammar.md
@@ -4,42 +4,50 @@ This chapter presents Ion 1.1's _domain grammar_, by which we mean the grammar o
 of values that drive Ion's encoding features.
 
 We use a BNF-like notation for describing various syntactic parts of a document, including Ion data structures. 
-In such cases, the BNF should be interpreted loosely to accommodate Ion-isms like commas and unconstrained ordering of struct fields.
+In such cases, the BNF should be interpreted loosely to accommodate Ion-isms like commas, whitespace, and unconstrained ordering of struct fields.
+
+For brevity, the following components are not explicitly defined in this grammar:
+* `value` - any Ion value literal
+* `unannotated-identifier-symbol` - an unannotated "identifier symbol" (as defined in the specification)
+* `unannotated-uint` - an unannotated integer greater than or equal to zero
 
 ### Documents
 ```bnf
-document           ::= ivm? segment*
+document           ::= ivm segment*
 
-ivm                ::= '$ion_1_0' | '$ion_1_1'
+ivm                ::= '$ion_1_1'
 
-segment            ::= value* directive?
+any-value-encoding ::= value | macro-invocation | tagless-element-list | tagless-element-sexp
 
-directive          ::= ivm 
-                     | encoding-directive 
-                     | symtab-directive 
+segment            ::= any-value-encoding* directive?
 
-symtab-directive   ::=  local-symbol-table     ; As per the Ion 1.0 specification¹
-
-encoding-directive ::= '$ion::(encoding ' module-name* ')'
+directive          ::= ivm | encoding-directive
 ```
 
-&nbsp;&nbsp;&nbsp;&nbsp;¹[Symbols – Local Symbol Tables](https://amazon-ion.github.io/ion-docs/docs/symbols.html#local-symbol-tables).
+### Directives
+```bnf
+encoding-directive    ::= '(:$ion ' any-directive-content ')'
+
+any-directive-content ::= set-symbols-content | add-symbols-content | set-macros-content | add-macros-content
+                        | use-content | module-content | import-content | encoding-content
+
+set-symbols-content   ::= 'set_symbols' symbol-text*
+add-symbols-content   ::= 'add_symbols' symbol-text*
+set-macros-content    ::= 'set_macros' macro-definition*
+add-macros-content    ::= 'add_macros' macro-definition*
+use-content           ::= 'use' catalog-key
+module-content        ::= 'module' module-name macro-table? symbol-table?
+import-content        ::= 'import' module-name catalog-key
+encoding-content      ::= 'encoding' module-name*
+```
 
 ### Modules
 ```bnf
-module-body             ::= import* inner-module* macro-table? symbol-table? 
-
-shared-module           ::= '$ion_shared_module::' ivm '::(' catalog-key module-body ')'
-
-import                  ::= '(import ' module-name catalog-key ')'
-
 catalog-key             ::= catalog-name catalog-version?
 
 catalog-name            ::= string
 
-catalog-version         ::= unannotated-uint                   ; must be positive
-
-inner-module            ::= '(module' module-name module-body ')'
+catalog-version         ::= unannotated-uint
 
 module-name             ::= unannotated-identifier-symbol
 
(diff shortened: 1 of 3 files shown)


_books/ion-1-1/src/grammar.md
```diff
@@ -53,64 +61,70 @@ symbol-text             ::= symbol | string
...
+macro-table-entry       ::= macro-definition-in-module | module-name
 ```
+
 ### Macro references
 ```bnf
 qualified-macro-ref     ::= module-name '::' macro-ref
 
 macro-ref               ::= macro-name | macro-addr
 
-qualified-macro-name    ::= module-name '::' macro-name
```
popematt, line 68: We need to keep this one because this is used in E-Expressions for things like `(:foo::bar ...)`.

_books/ion-1-1/src/grammar.md
```diff
@@ -53,64 +61,70 @@ symbol-text             ::= symbol | string
...
+                             | 'timestamp_ms' | 'timestamp_us' | 'timestamp_ns'
+                             | 'symbol'
+
+primitive-encoding-opcode  ::= '0x60' | '0x61' | '0x62' | '0x64' | '0x68'
+                             | '0xE0' | '0xE1' | '0xE2' | '0xE4' | '0xE8'
+                             | '0x6B' | '0x6C' | '0x6D'
+                             | '0x70'
+                             | '0x82' | '0x83' | '0x84'
+                             | '0x85' | '0x86' | '0x87'
+                             | '0xEE'
```
popematt: Let's just drop this. When we were planning to allow eg 0x63 which didn't have a name/alias, opcodes were an escape hatch, but now I think it's just extra complexity.

_books/ion-1-1/src/grammar.md
```diff
@@ -53,64 +61,70 @@ symbol-text             ::= symbol | string
...
+tagless-element-sexp       ::= '(' any-type-marker-encoding value* ')'
+```
 
-macro-arg               ::= tdl-expression | expression-group
+### Macro invocations
 
-expression-group        ::= '(..' tdl-expression* ')'
+```bnf
+no-argument             ::= '(:)'
+macro-argument          ::= any-value-encoding | no-argument
```
popematt: I don't know if it's reflected in this grammar, but...
* it should be possible to use an E-Exp as a macro argument
* within a template definition, you can use a placeholder as an argument
* you can use an e-expression to provide the default value for a placeholder
* you cannot use a placeholder in the default value of another placeholder

Some of those _might_ be more exhaustive than what we want to describe in the grammar, but should perhaps be noted.

_books/ion-1-1/src/grammar.md
```diff
@@ -53,64 +61,70 @@ symbol-text             ::= symbol | string
 
 macro-table             ::= '(macros' macro-table-entry* ')'
 
-macro-table-entry       ::= macro-definition
-                          | macro-export
-                          | module-name
-
-macro-export            ::= '(export' qualified-macro-ref macro-name-declaration? ')'
```
popematt, line 60: The chapter on modules still says it supports the `export` clause.

_books/ion-1-1/src/grammar.md
```diff
@@ -53,64 +61,70 @@ symbol-text             ::= symbol | string
...
+```
 
-macro-arg               ::= tdl-expression | expression-group
+### Macro invocations
 
-expression-group        ::= '(..' tdl-expression* ')'
+```bnf
+no-argument             ::= '(:)'
+macro-argument          ::= any-value-encoding | no-argument
+macro-invocation        ::= '(:' any-macro-ref-encoding macro-argument* ')'
```
popematt: We should probably name this one `e-expression`.

popematt (commented):
There's a few changes I think we need to make now. Long term, I think we should try to have a complete grammar because having such a grammar will make it easier to reliably implement things like syntax highlighting in IDEs.

_books/ion-1-1/src/grammar.md
tgregg: Yes, but in those cases can't it always be `qualified-macro-ref`?

_books/ion-1-1/src/grammar.md
tgregg: Is that what we intend? Or is everything exported?

_books/ion-1-1/src/grammar.md
tgregg: Sounds good to me.

_books/ion-1-1/src/grammar.md
tgregg: I can do that

_books/ion-1-1/src/grammar.md
popematt: Oh yes, you're right.

_books/ion-1-1/src/grammar.md
tgregg: > it should be possible to use an E-Exp as a macro argument

```
macro-argument          ::= any-value-encoding | no-argument
any-value-encoding     ::= value | macro-invocation | tagless-element-list | tagless-element-sexp
```

> within a template definition, you can use a placeholder as an argument

Is that what we want?

> you can use an e-expression to provide the default value for a placeholder
> you cannot use a placeholder in the default value of another placeholder

I think this change will achieve both:
```
- tagged-placeholder-default ::= '(:? ' value')'
+ tagged-placeholder-default ::= '(:? ' any-value-encoding ')'
```

This will allow value literals, e-expressions, and tagless-element sequences (which aren't very useful there, but are not necessary to prohibit); but not placeholders.

_books/ion-1-1/src/grammar.md
popematt: I don't know what you specifically mean by "everything" in this context, but I'll do my best to answer.

 The entire macro table of a module is "exported" from that module, but the export clause was intended to allow a macro table to include or use in that macro table a macro that was defined elsewhere, possibly with a different name or with no name at all. Maybe the word `export` is misleading. 

I think we originally introduced `export` in order to deal with name conflicts when composing a module from other modules. That seems like it's probably less of a concern now that we have the encoding module sequence.

_books/ion-1-1/src/grammar.md
popematt: > > it should be possible to use an E-Exp as a macro argument
>
> ``` 
> macro-argument          ::= any-value-encoding | no-argument
> any-value-encoding     ::= value | macro-invocation | tagless-element-list | tagless-element-sexp
> ```

Oh, I see that now. That's what I get for trying to review it on my tiny phone screen.


>> within a template definition, you can use a placeholder as an argument
>
> Is that what we want?

Yes, I think so. This is the only way we have to support composition of macros.

tgregg: >> within a template definition, you can use a placeholder as an argument
> Is that what we want?

If so, I think the following will achieve it:

```
macro-definition                  ::= '(' macro-definition-contents ')'
macro-definition-contents  ::= macro-name-declaration value-with-placeholders ; This prohibits "identity" macros
argument-or-placeholder    ::= value-with-placeholders | no-argument | placeholder
e-exp-with-placeholders     ::= '(:' any-macro-ref-encoding argument-or-placeholder* ')'
value-with-placeholders      ::= value | list-with-placeholders | sexp-with-placeholders
                             | struct-with-placeholders | e-exp-with-placeholders
```

_books/ion-1-1/src/grammar.md
tgregg: Should we leave it out for now, mark it for further review, and potentially re-add with more examples later?

popematt: Sure, we can leave it out for now and mark it for further review.

popematt (approved):

PR #408, opened 2025-10-16
Clarifies that Ion 1.0 symbol table syntax is a no-op in Ion 1.1.

tgregg:
### Description of changes:
Only directives may be used to modify the encoding context in Ion 1.1.

----

### DIFF
--- a/_books/ion-1-1/src/modules/directives.md
+++ b/_books/ion-1-1/src/modules/directives.md
@@ -166,6 +166,6 @@ When the version is not provided, a default value of 1 is used.
 
 ## Ion 1.0 symbol table syntax not supported
 
-In Ion 1.1 streams, Ion 1.0 symbol table syntax (`$ion_symbol_table::{...}`) is not supported.
+In Ion 1.1 streams, Ion 1.0 symbol table syntax (for example, `$ion_symbol_table::{...}`) is not supported.
 Ion 1.1 writers should not write such an annotated struct at the top level; Ion 1.1 readers
 must treat such an annotated struct as a no-op at the top level.
--- a/_books/ion-1-1/src/modules/directives.md
+++ b/_books/ion-1-1/src/modules/directives.md
@@ -163,3 +163,9 @@ When the version is not provided, a default value of 1 is used.
 // Defaults to version 1
 (:$ion use "com.example.geometry")
 ```
+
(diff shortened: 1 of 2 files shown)


_books/ion-1-1/src/modules/directives.md
```diff
@@ -163,3 +163,9 @@ When the version is not provided, a default value of 1 is used.
 // Defaults to version 1
 (:$ion use "com.example.geometry")
 ```
+
+## Ion 1.0 symbol table syntax not supported
+
+In Ion 1.1 streams, Ion 1.0 symbol table syntax (`$ion_symbol_table::{...}`) is not supported.
```
popematt: And `$ion_symbol_table::null.struct`?

From Ion 1.0 spec:

> If the struct is null (null.struct) then it is treated as if it were an empty struct.

I think that means it's treated as an empty symbol table struct, rather than an empty user value struct.

austnwil (commented):
Do we have a convention for serializing shared modules in Ion 1.1? Should we clarify if `$ion_shared_symbol_table::` in a 1.1 data stream still carries shared symbol table semantics?

This is a bit of a separate issue but I am curious.

popematt:
> Do we have a convention for serializing shared modules in Ion 1.1? Should we clarify if `$ion_shared_symbol_table::` in a 1.1 data stream still carries shared symbol table semantics?
> 
> This is a bit of a separate issue but I am curious.

We are taking the position that the convention for shared modules is specifically a _convention_ and not a _specification_.

The `$ion_shared_symbol_table::` annotation doesn't have any special semantics in an Ion parser. In only gets special semantics at the application layer, if the application layer is looking for that annotation in order to find shared symbol table definitions.

_books/ion-1-1/src/modules/directives.md
tgregg: I don't think it's necessary to re-specify Ion 1.0 symbol table syntax here. I will add "for example".

tgregg:
```suggestion
In Ion 1.1 streams, Ion 1.0 symbol table syntax (for example, `$ion_symbol_table::{...}`) is not supported.
```

popematt (approved):

PR #415, opened 2026-09-17
Add clarification for import handling with max_id=0

popematt:
### Issue #, if available:

None

### Description of changes:

I was working on implementing something and realized that an import with a `max_id` of `0` is a no-op. No symbols will be imported, and no SIDs will be reserved. The only thing that could happen is that it might try to resolve the shared symbol table name in the Catalog. Resolving the name does nothing helpful for the application, but it _might_ be useful for an external actor who is trying to probe an application to see what symbol tables are available. Choosing to ignore the import means less work for the reader.

I'm open to using `MAY` instead of `SHOULD` if you think I'm being too prescriptive.

----


almann (approved):

toddjonker (commented):
Good thinking, Matt.

